PR #1, opened 2016-10-06
Mal character scraping

toyhammered:
_hides in corner_


NuckChorris:
Seems reasonable, LGTM

PR #2, opened 2016-10-11
Ap importer

toyhammered:
There are some parts that def need refactoring. Still need to do `#media` tests (even though I don't quite know how Mapping works :| )

@NuckChorris


app/models/list_import/anime_planet/row.rb
```diff
@@ -0,0 +1,154 @@
...
+        chapters = fragment.at_css('.entryBar .iconVol')&.text
+
+        chapters&.split('Ch:')&.last&.tr('+', '')&.to_i
+      end
+
+      # Status
+      def extract_status(fragment)
+        show_type = fragment.at_css('.myListBar')&.content
+
+        show_type&.split(':')&.last&.split('-')&.first&.strip
```
NuckChorris: Why so many amperdots?  Are these really sometimes not there?

Perhaps it would be better to just add `return unless show_type` as an early return?

app/models/list_import/anime_planet/row.rb
```diff
@@ -0,0 +1,154 @@
...
+      private
+
+      def tooltip
+        @tooltip ||= Nokogiri::HTML.fragment(
+          node.at_css('.tooltip').attr('title')
+        )
+      end
+
+      # Media Info
+      def extract_title(fragment)
```
NuckChorris: What's the thinking behind the `extract_` methods?

NuckChorris (changes requested):

app/models/list_import/anime_planet/row.rb
toyhammered: mainly started it because of `media_info` and it looking nice with just 1 line. Kind of just adopted to everything afterwards, can prob remove most of them.

app/models/list_import/anime_planet/row.rb
toyhammered: Is it better to return early? Wasn't actually sure of the best way.

app/models/list_import/anime_planet.rb
```diff
@@ -0,0 +1,82 @@
...
+class ListImport
+  class AnimePlanet < ListImport
+    ANIME_PLANET_HOST = 'http://www.anime-planet.com/users/'.freeze
+
+    # accepts a username as input
+    validates :input_text, length: {
+      minimum: 3,
+      maximum: 20
+    }, presence: true
+    # does not accept file uploads
```
NuckChorris, line 34: You should validate the absence of the `input_file` even though you can't easily test it.

app/models/list_import/anime_planet/row.rb
```diff
@@ -0,0 +1,154 @@
...
+        when 'Stalled' then :on_hold
+        when 'Dropped' then :dropped
+        when 'Watched', /Read([0-9.]+)?/ then :completed
+        end
+      end
+
+      def progress
+        if status == (:completed || :planned || nil)
+          if type == 'anime'
+            amount = media_info[:episode_count]
```
NuckChorris: This logic definitely doesn't make sense for `:planned`

app/models/list_import/anime_planet/row.rb
NuckChorris: I see, that makes sense for `media_info`, not sure about the others.

I also would recommend instead of naming methods after an action like `extract_title`, name them after their _returned data_, such as `media_title` or `show_type`

app/models/list_import/anime_planet/row.rb
toyhammered: it was in there because it might cause some issues (it is never going to have an episode count ever)

app/models/list_import/anime_planet/row.rb
NuckChorris: Chained `&.` is not as clear as an early return.  Plus it might (theoretically) be slower, though I don't have any empirical evidence

app/models/list_import/anime_planet/row.rb
NuckChorris: Yes, but if you have something on your `planned` list, it doesn't make sense to set your progress to 100%.  More likely, your progress is 0%

toyhammered: aaahhh I see what you are saying I actually thought it was doing something else but now that I look at it, you are right.

spec/models/list_import/anime_planet_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+
+    # Page 1 Sierra
+    stub_request(:get, "#{host}sierrawithas/anime#{extensions}1")
+      .to_return(body: fixture('list_import/anime_planet/sierra-anime.html'))
+    stub_request(:get, "#{host}sierrawithas/manga#{extensions}1")
+      .to_return(body: fixture('list_import/anime_planet/sierra-manga.html'))
+  end
+
+  it { should validate_presence_of(:input_text) }
+  it { should validate_length_of(:input_text)
```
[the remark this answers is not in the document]
NuckChorris: > Prefer do...end for multi-line blocks without chaining. (https://github.com/bbatsov/ruby-style-guide#single-line-blocks)

This can be ignored in specs (both of these)

app/models/list_import/anime_planet.rb
```diff
@@ -0,0 +1,83 @@
...
+    def each
+      %w[anime manga].each do |type|
+        amount = get("#{input_text}/#{type}").css('.pagination li')
+          &.map(&:content)
+          &.map(&:to_i)
+          &.max
+        amount ||= 1
+
+        1.upto(amount) do |page|
+          get("#{input_text}/#{type}", page).css('.cardDeck .card').each do |card|
```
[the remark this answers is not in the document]
NuckChorris: Ignore this, the 2 characters aren't worth wrapping this

app/models/list_import/anime_planet.rb
```diff
@@ -0,0 +1,83 @@
...
+      end
+    end
+
+    # private
+
+    def get(url, page = 1, opts = {})
+      Nokogiri::HTML(
+        Typhoeus::Request.get(
+          build_url(url, page),
+            {
```
NuckChorris: My recommendation is to split this across multiple lines, like this:

``` ruby
url = build_url(url, page, ...)
request = Typheous::Request.get(...)
html = Nokogiri::HTML(request.body)
html
```

app/models/list_import/anime_planet.rb
```diff
@@ -0,0 +1,83 @@
...
+          get("#{input_text}/#{type}", page).css('.cardDeck .card').each do |card|
+            row = Row.new(card, type)
+            # yield row.media, row.data
+            yield row.media
+          end
+        end
+      end
+    end
+
+    # private
```
NuckChorris: Why is this commented out?

app/models/list_import/anime_planet.rb
```diff
@@ -0,0 +1,83 @@
...
+          &.map(&:content)
+          &.map(&:to_i)
+          &.max
+        amount ||= 1
+
+        1.upto(amount) do |page|
+          get("#{input_text}/#{type}", page).css('.cardDeck .card').each do |card|
+            row = Row.new(card, type)
+            # yield row.media, row.data
+            yield row.media
```
NuckChorris: You need to yield the data for the library entry

NuckChorris (changes requested):

app/models/list_import/anime_planet/row.rb
```diff
@@ -0,0 +1,159 @@
...
+      private
+
+      def tooltip
+        @tooltip ||= Nokogiri::HTML.fragment(
+          node.at_css('.tooltip').attr('title')
+        )
+      end
+
+      # Media Info
+      def media_title(fragment)
```
NuckChorris: Why not just access `tooltip` directly within the method and not take an argument?

toyhammered: that does sound more logical.

toyhammered:
I expect a good picture!

NuckChorris (approved):

PR #3, opened 2016-10-16
Anilist importer

toyhammered:
Still need to implement notes. Having issues with my `#each` test not actually sure _why_ it is returning an error (at the `yield row.media`)
@NuckChorris


spec/support/factory_girl.rb
```diff
@@ -1,14 +0,0 @@
-require 'factory_girl'
```
toyhammered, line 1: hmmm this is most likely a screw up on my end, accidentally added my version and removed it using git rm before commiting.

app/models/list_import/anilist.rb
```diff
@@ -0,0 +1,97 @@
...
+
+    private
+
+    def get_auth_token
+      url = "#{ANILIST_API}auth/access_token"
+      request = Typhoeus::Request.new(url,
+        method: :post,
+        body: {
+          grant_type: 'client_credentials',
+          client_id: 'toyhammered-c6imc',
```
toyhammered, line 70: should these be added to `.env`? (I don't really care it is only `GET` access anyways)

NuckChorris: Would be good to read from `ENV['ANILIST_CLIENT_ID']` and `ENV['ANILIST_CLIENT_SECRET']`, but this works for now

NuckChorris (approved):

PR #4, opened 2016-10-19
adds LinkedProfile and LinkedSite boilerplate

toyhammered:
I am not entirely sure _what_ to be testing in the models aside for their relationships.
🙏 
@NuckChorris


app/resources/linked_site_resource.rb
```diff
@@ -0,0 +1,4 @@
+class LinkedSiteResource < BaseResource
+  immutable
```
toyhammered, line 2: Was not sure if this was right.

app/resources/linked_profile_resource.rb
```diff
@@ -0,0 +1,20 @@
+class LinkedProfileResource < BaseResource
+  attribute :token
+  attributes :external_user_id, :url, :share_to, :share_from
+
+  has_one :user
+  has_one :linked_site
+
+  filter :self, apply: -> (_r, _v, options) {
```
toyhammered: don't forget to look into if it needs to be `_r` or `records`

app/models/linked_profile.rb
```diff
@@ -0,0 +1,30 @@
...
+#  index_linked_profiles_on_linked_site_id  (linked_site_id)
+#  index_linked_profiles_on_user_id         (user_id)
+#
+# Foreign Keys
+#
+#  fk_rails_166e103170  (user_id => users.id)
+#  fk_rails_25de88e967  (linked_site_id => linked_sites.id)
+#
+
+class LinkedProfile < ActiveRecord::Base
```
NuckChorris: Could you change these to inherit from `ApplicationRecord`?

app/models/linked_profile.rb
```diff
@@ -0,0 +1,30 @@
...
+  belongs_to :user
+  belongs_to :linked_site
```
NuckChorris: These should have `required: true` since that's becoming the default in Rails 5

app/models/linked_site.rb
```diff
@@ -0,0 +1,18 @@
...
+#  link_type  :integer
+#  name       :string
+#  share_from :boolean
+#  share_to   :boolean
+#  created_at :datetime         not null
+#  updated_at :datetime         not null
+#
+
+class LinkedSite < ActiveRecord::Base
+  LINK_TYPE = %w[username OAuth2].freeze
```
NuckChorris: `oauth2` — symbols like this shouldn't be uppercase

Also should be a `%i` literal for symbols, not `%w`

And it's not necessary to declare it as a constant, you can also just do `enum link_type: %i[username oauth2]`

app/models/linked_profile.rb
```diff
@@ -0,0 +1,30 @@
+# == Schema Information
+#
+# Table name: linked_profiles
```
NuckChorris, line 4: There should probably be a `public` boolean column too

app/policies/linked_profile_policy.rb
```diff
@@ -0,0 +1,11 @@
+class LinkedProfilePolicy < ApplicationPolicy
+  def show?
+    record.user == user || user.has_role?(:admin)
```
NuckChorris: Should probably be `record.public? || record.user == user || is_admin?` because:
1. linked profiles need to be public so we can show 'em on your profile
2. `is_admin?` is a shorthand for `user.has_role?(:admin)`

app/resources/linked_profile_resource.rb
```diff
@@ -0,0 +1,20 @@
...
+    current_user = options[:context][:user]
+    records.where(id: current_user&.id) || User.none
+  }
+
+  def fetchable_fields
+    if current_user == _model
+      super
+    else
+      super - :token
```
NuckChorris: You missed an important detail here: `:token` needs to be `[:token]` — you can't subtract a symbol from an array, but you can subtract an array from an array

app/resources/linked_site_resource.rb
NuckChorris: It is

db/migrate/20161019004726_create_linked_profiles.rb
```diff
@@ -0,0 +1,16 @@
+class CreateLinkedProfiles < ActiveRecord::Migration
+  def change
+    create_table :linked_profiles do |t|
+      t.references :user, index: true, foreign_key: true
+      t.references :linked_site, index: true, foreign_key: true
```
NuckChorris: These two should probably have `null: false`

db/migrate/20161019004726_create_linked_profiles.rb
```diff
@@ -0,0 +1,16 @@
...
+
+      t.integer :external_user_id
```
NuckChorris: This should be a string... I know, confusing, but some other sites are going to be dipshits and not provide an integer ID.  For example, some sites might have UUIDs like `59208787-86ec-4aef-97ae-cede8e72ecc4` or weird base64 crap like `Mmvofg3gGq0c4w`

We need to be ready to accept this

app/models/linked_profile.rb
```diff
@@ -0,0 +1,30 @@
...
+# Foreign Keys
+#
+#  fk_rails_166e103170  (user_id => users.id)
+#  fk_rails_25de88e967  (linked_site_id => linked_sites.id)
+#
+
+class LinkedProfile < ActiveRecord::Base
+  belongs_to :user
+  belongs_to :linked_site
+end
```
NuckChorris, line 36: I would recommend tossing in a validation of `validates :url, presence: true, if: :public?` (once you add a `public` column)

db/migrate/20161019004726_create_linked_profiles.rb
```diff
@@ -0,0 +1,16 @@
...
+      t.string :url
+      t.boolean :share_to
+      t.boolean :share_from
```
NuckChorris: These should be `null: false` w/ a default probably, like `default: false`

db/migrate/20161019003515_create_linked_sites.rb
```diff
@@ -0,0 +1,12 @@
+class CreateLinkedSites < ActiveRecord::Migration
+  def change
+    create_table :linked_sites do |t|
+      t.string :name
+      t.boolean :share_to
+      t.boolean :share_from
+      t.integer :link_type
```
NuckChorris: All of these should have `null: false` on them, since they're all switches.

Remember: a switch is either on or off.  You don't want a switch that can randomly cease existence

NuckChorris (changes requested):
A good start :+1:

app/resources/linked_profile_resource.rb
NuckChorris: Turns out, it's supposed to be `records`, nice catch!

db/migrate/20161019004726_create_linked_profiles.rb
toyhammered: I am assuming this only references `lines 9 & 10`?

db/migrate/20161019003515_create_linked_sites.rb
toyhammered: I am assuming this only references `lines 5 - 7`

NuckChorris: No, I mean all of them.  What good is a website without a name?

app/policies/linked_profile_policy.rb
toyhammered: isn't `is_admin?` only accessible in `create, update, destroy` because of the `alias_method`?

```
  def is_admin? # rubocop:disable Style/PredicateName
    user && user.has_role?(:admin, model_class)
  end
  alias_method :create?, :is_admin?
  alias_method :update?, :is_admin?
  alias_method :destroy?, :is_admin?
```

NuckChorris: No, `alias_method from, to` means "create a method with `from` name that just defers to `to`"

So that basically creates the following method:

``` ruby
def create?(*args)
  return is_admin?(*args)
end
```

It does not affect scoping, it's not magic.  It literally creates an _alias_ for a _method_ ;)

toyhammered: aye aye captain! 

(now a bit off-topic) why do you do this then?

```
class LibraryEntryPolicy < ApplicationPolicy
  def show?
    # Yes, if you own it or are admin
    return true if record.user == user || user.has_role?(:admin)
    # No, if it's private
    return false if record.private?
    # No, if it's nsfw and you're not a perv
    return false if user.sfw_filter? && record.media.nsfw?
    # Otherwise, yeah
    true
  end
...
```

NuckChorris: Sometimes I make mistakes

db/migrate/20161019003515_create_linked_sites.rb
toyhammered: :|

app/policies/linked_profile_policy.rb
toyhammered: impossibru

app/models/linked_profile.rb
toyhammered: is `public` reserved cause it is causing me issues when I am trying to test it.

spec/controllers/users_controller_spec.rb
```diff
@@ -1,3 +1,79 @@
...
+#  last_recommendations_update :datetime
+#  last_sign_in_at             :datetime
+#  last_sign_in_ip             :string(255)
+#  life_spent_on_anime         :integer          default(0), not null
+#  location                    :string(255)
+#  mal_username                :string(255)
+#  name                        :string(255)
+#  ninja_banned                :boolean          default(FALSE)
+#  onboarded                   :boolean          default(FALSE), not null
+#  past_names                  :string           default([]), not null, is an Array
```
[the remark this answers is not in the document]
toyhammered: go fuck yourself hound...

toyhammered:
I have a feeling my mistake is not doing `validates :name, presence: true` and for all the other `null: false` fields added to the database (is this why FactoryGirl is failing?)

app/models/user.rb
```diff
@@ -89,6 +87,7 @@ class User < ApplicationRecord
     dependent: :destroy
   has_many :following, class_name: 'Follow', foreign_key: 'follower_id',
     dependent: :destroy
+  has_many :linked_profiles
```
NuckChorris: Pretty much all `has_many` associations ought to have `dependent: :destroy`.  I forgot about this until recently, but it's important, because otherwise things will break if we try to delete a user

app/resources/linked_profile_resource.rb
```diff
@@ -0,0 +1,20 @@
...
+  attribute :token
+  attributes :external_user_id, :url, :share_to, :share_from
+
+  has_one :user
+  has_one :linked_site
+
+  filter :self, apply: -> (records, _v, options) {
+    current_user = options[:context][:user]
+    records.where(id: current_user&.id) || User.none
+  }
```
NuckChorris: You don't need a `self` filter for this, the client has the user_id and can use it for querying (or can just request `include=linkedProfile` on a user)

spec/models/linked_profile_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+
+RSpec.describe LinkedProfile, type: :model do
+  it { should belong_to(:user) }
+  it { should belong_to(:linked_site) }
+
+  subject { described_class.new }
+
+  describe 'validates url' do
+    context 'if public' do
+      before { allow(subject).to receive(:public?).and_return(true) }
```
NuckChorris: You can better express this by doing this:

`subject { described_class.new(public: true) }`

Nested blocks can have different subjects from their parents

spec/factories/linked_sites.rb
```diff
@@ -0,0 +1,17 @@
...
+#  link_type  :integer          not null
+#  name       :string           not null
+#  share_from :boolean          default(FALSE), not null
+#  share_to   :boolean          default(FALSE), not null
+#  created_at :datetime         not null
+#  updated_at :datetime         not null
+#
+
+FactoryGirl.define do
+  factory :linked_site do
```
NuckChorris, line 17: You should generate valid fake data for columns without defaults here, like:

``` ruby
name { Faker::Company.name }
link_type { LinkedSite.link_types.sample }
```

spec/factories/linked_profiles.rb
```diff
@@ -0,0 +1,33 @@
...
+# Foreign Keys
+#
+#  fk_rails_166e103170  (user_id => users.id)
+#  fk_rails_25de88e967  (linked_site_id => linked_sites.id)
+#
+
+FactoryGirl.define do
+  factory :linked_profile do
+    user
+    linked_site
```
NuckChorris, line 33: You should generate fake data here, like this:

``` ruby
url { Faker::Internet.url }
external_user_id { Faker::Internet.user_name }
```

app/models/linked_site.rb
```diff
@@ -0,0 +1,16 @@
...
+#  link_type  :integer          not null
+#  name       :string           not null
+#  share_from :boolean          default(FALSE), not null
+#  share_to   :boolean          default(FALSE), not null
+#  created_at :datetime         not null
+#  updated_at :datetime         not null
+#
+
+class LinkedSite < ApplicationRecord
+  enum link_type: %i[username oauth2]
```
NuckChorris, line 17: Even if it's not going to be touched by most users, this should definitely have at least basic presence validations for not-null columns

app/models/linked_profile.rb
```diff
@@ -0,0 +1,33 @@
...
+#
+#  fk_rails_166e103170  (user_id => users.id)
+#  fk_rails_25de88e967  (linked_site_id => linked_sites.id)
+#
+
+class LinkedProfile < ApplicationRecord
+  belongs_to :user, required: true
+  belongs_to :linked_site, required: true
+
+  validates :url, presence: true, if: :public?
```
NuckChorris: You've also marked `external_user_id` not-null, so there should be a presence validation for that too.

Also, if `url` isn't always required then it shouldn't be marked `not null` — that'll cause :boom:

spec/models/linked_site_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+#  share_from :boolean          default(FALSE), not null
+#  share_to   :boolean          default(FALSE), not null
+#  created_at :datetime         not null
+#  updated_at :datetime         not null
+#
+
+require 'rails_helper'
+
+RSpec.describe LinkedSite, type: :model do
+  pending "add some examples to (or delete) #{__FILE__}"
```
NuckChorris: Please add some examples or delete this :P

app/models/linked_profile.rb
toyhammered: is url not going to be required if public is false?

toyhammered: Isn't it required though if public is true (this shouldn't blow up because of the failsafe if statement?)

app/policies/linked_profile_policy.rb
```diff
@@ -0,0 +1,11 @@
+class LinkedProfilePolicy < ApplicationPolicy
+  def show?
+    record.user == user || record.public? || is_admin?
+  end
```
NuckChorris: With Pundit-Resources we no longer use `#show?`

Instead, add a `Scope` class like you see in the `LibraryEntryPolicy`.  The logic should be pretty similar.

app/models/installment.rb
```diff
@@ -14,7 +13,6 @@
 #  index_installments_on_franchise_id             (franchise_id)
 #  index_installments_on_media_type_and_media_id  (media_type,media_id)
 #
-# rubocop:enable Metrics/LineLength
```
NuckChorris, line 18: Could you undo all the places you removed `# rubocop:enable Metrics/LineLength`?

spec/controllers/anime_controller_spec.rb
```diff
@@ -1,3 +1,47 @@
...
+#
+# Indexes
+#
+#  index_anime_on_age_rating  (age_rating)
+#  index_anime_on_slug        (slug) UNIQUE
+#  index_anime_on_user_count  (user_count)
+#  index_anime_on_wilson_ci   (average_rating)
+#
+# rubocop:enable Metrics/LineLength
+
```
NuckChorris: Could you remove the annotations from `spec/controllers/*`

P.S. `git add -p` is your friend :wink:

NuckChorris (changes requested):
Pretty much all the logic is right, just got a little bit left to go on the cleanup :+1:

app/policies/linked_profile_policy.rb
```diff
@@ -0,0 +1,25 @@
+class LinkedProfilePolicy < ApplicationPolicy
+  def update?
+    record.user == user || is_admin?
+  end
+  alias_method :create?, :update?
+  alias_method :destroy?, :update?
+
+  class Scope < Scope
+    def resolve
+      return scope.where(private: true) unless user
```
NuckChorris: This seems backwards.  Limit anonymous users (`unless user`) to a scope where `private: true`?  So they can _only_ see private stuff?

spec/factories/installments.rb
```diff
@@ -15,7 +14,6 @@
 #  index_installments_on_franchise_id             (franchise_id)
 #  index_installments_on_media_type_and_media_id  (media_type,media_id)
 #
-# rubocop:enable Metrics/LineLength
```
NuckChorris, line 18: Missed a spot 😛

app/policies/linked_profile_policy.rb
toyhammered: oh.. I am an idiot was thinking backwards

spec/factories/installments.rb
toyhammered: 🖕

NuckChorris (approved):

NuckChorris:
LGTM :tada:

PR #7, opened 2016-10-28
Create twitter authentication and import of followers

dissident:


lib/authorization/assertion/twitter_auth.rb
```diff
@@ -0,0 +1,49 @@
+module Authorization
+  module Assertion
+    class TwitterAuth
+      def initialize(access_token, access_token_secret)
+        @access_token = access_token
+        @access_token_secret = access_token_secret
+        @twitter_client = twitter_client
```
NuckChorris: Instead of saving this to an instance variable in the `initialize` method and then accessing that, it might be better to memoize it within the `twitter_client` method and refer to the method, like so:

`@twitter_client ||= Twitter::REST::Client.new do ...`

lib/authorization/assertion/twitter_auth.rb
```diff
@@ -0,0 +1,49 @@
...
+        @user_data = user_data
```
NuckChorris: `@user_data` _probably_ doesn't need to be saved to an instance variable, I suspect Twitter will handle the caching in the client

lib/authorization/assertion/twitter_auth.rb
```diff
@@ -0,0 +1,49 @@
...
+        Net::HTTP.get_response(img)
+      end
+
+      def user!
+        User.where(twitter_id: @user_data.id).first if @user_data.present?
+      end
+
+      def import_friends
+        return unless @user_data.present?
+        user = user!
```
NuckChorris: You don't need to save `user!`, you can use it directly on line 30 as `follower: user!,`

spec/controllers/anime_controller_spec.rb
```diff
@@ -1,3 +1,47 @@
...
+#
+# Indexes
+#
+#  index_anime_on_age_rating  (age_rating)
+#  index_anime_on_slug        (slug) UNIQUE
+#  index_anime_on_user_count  (user_count)
+#  index_anime_on_wilson_ci   (average_rating)
+#
+# rubocop:enable Metrics/LineLength
+
```
NuckChorris: Would you mind removing these comments in `spec/controllers` from your PR?  It's not your fault, I just can't get the `annotate` gem to stop putting them in these files until ctran/annotate_models#370 lands, so we're just stuck manually removing them whenever it tries to add them :/

NuckChorris (changes requested):
Looks fantastic, great job 🎉 Just a couple minor tweaks and this is ready

NuckChorris (approved):

NuckChorris:
I re-ran the specs now that the `SSLError` has disappeared, any idea what's broken?

dissident:
No, I got this error locally on the day after the review. I thought it was my local environment. I tinkered with the problem.

NuckChorris:
Looks like the Twitter gem is trying to hit up `https://api.twitter.com/1.1/users/show.json?user_id=2244994945` when it first logs in, to get the user info, and WebMock is complaining because that doesn't have a `stub_request`

dissident:
Fix it

NuckChorris (approved):
LGTM

PR #9, opened 2016-11-02
Support uppercase characters and correct URL

trmcnvn:
Wanted to PR to make sure we're good.

### DIFF
--- a/lib/html/pipeline/kitsu_mention_filter.rb
+++ b/lib/html/pipeline/kitsu_mention_filter.rb
@@ -4,13 +4,17 @@ module HTML
       # Disable "@mention" mentions
       MentionLogins = []
 
+      def base_url
+        '/users/'
+      end
+
       # Check if user exists before we linkify it
       def link_to_mentioned_user(login)
         super if User.by_name(login).exists?
       end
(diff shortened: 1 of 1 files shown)


lib/html/pipeline/kitsu_mention_filter.rb
```diff
@@ -4,13 +4,17 @@ class KitsuMentionFilter < MentionFilter
...
+      end
+
       # Check if user exists before we linkify it
       def link_to_mentioned_user(login)
         super if User.by_name(login).exists?
       end
 
       def username_pattern
-        /[a-z0-9][a-z0-9_]*/
+        /[a-zA-Z0-9][a-zA-Z0-9_]*/
```
NuckChorris, line 17: username_pattern is actually run w/ `i` flag so case shouldn't matter

trmcnvn: Can confirm locally that it wasn't working before this change.

NuckChorris: Bizarre.  Well then, LGTM

PR #12, opened 2016-11-08
Badges

dissident:


config/routes.rb
```diff
@@ -409,7 +409,9 @@
     jsonapi_resources :anime
     jsonapi_resources :manga
     jsonapi_resources :drama
-    jsonapi_resources :users
+    jsonapi_resources :users do
+      jsonapi_resources :bestowments, only: [:index]
```
NuckChorris: It's preferable to keep everything at the top level I think, JR will automatically add routes for the association

config/initializers/zhe_badges.rb
```diff
@@ -0,0 +1,3 @@
+Dir[Rails.root + './app/badges/*.rb'].each do |file|
+  require file
+end
```
NuckChorris, line 3: The filename on this is interesting, where did `zhe` come from, is that to keep this loading last?

app/chewy/library_index.rb
```diff
@@ -0,0 +1,19 @@
...
+        value: -> (e) { e.media.abbreviated_titles }
+      field :status, type: 'string'
+      field :media_type, type: 'string'
+      field :progress, type: 'integer'
+      field :notes, type: 'string'
+      field :updated_at, type: 'date'
+      field :user_id, type: 'integer'
+    end
+  end
+end
```
NuckChorris: Would you mind removing this file?  I didn't mean to include it in the commit of badges :x My bad

lib/badge.rb
```diff
@@ -0,0 +1,75 @@
...
+
+  attr_reader :title, :description, :goal, :rank, :user
+
+  def self.slug
+    name.underscore.dasherize.sub('-badge', '')
+  end
+
+  def slug
+    self.class.slug
+  end
```
NuckChorris, line 17: These `slug` methods don't seem to be used for anything (instead you're using the class name, which makes more sense), so maybe just remove 'em

lib/badge.rb
```diff
@@ -0,0 +1,75 @@
...
+
+  def earned?
+    return false unless @goal
+    if progress.nil?
+      instance_eval(&@goal)
+    else
+      progress > @goal
+    end
+  end
```
NuckChorris, line 45: This logic is duplicated between `#earned?` and `#get_goal_result`, I think

lib/badge.rb
```diff
@@ -0,0 +1,75 @@
...
+        end
+        next unless progress >= current_goal
+        @rank = state.rank
+        @title = state.title
+        @description = state.description
+        next_goal = true
+      end
+    end
+    @rank
+  end
```
NuckChorris, line 58: I think this may have been a miscommunication, sorry.  Each rank is a separate `Badge` — the `Badge` info doesn't vary by the rank of the Bestowment.  Instead, as you gain ranks, you get _additional_ badges.

The DSL is just there to make it easier to declare those badge classes (otherwise we'd have like 50+ files for all the different ranks) while sharing common logic (`progress`, etc.) and allowing for the flow of only creating the lowest-ranked unfinished bestowment.

NuckChorris (changes requested):
You're on the right track, though it seems we had a misunderstanding on the logic.  If you have some questions on that, Josh and I can probably sit down some time and discuss that

app/resources/bestowment_resource.rb
```diff
@@ -0,0 +1,5 @@
+class BestowmentResource < BaseResource
+  attributes :badge_id, :rank, :progress, :bestowed_at
+
+  has_one :user
```
NuckChorris, line 5: This is probably better to structure as another `Badge` resource (with no route of its own) which is `has_one`'d off of `Bestowment`, so we can access information on it

config/initializers/zhe_badges.rb
dissident: Exactly, this file must be loaded after devise initializer, this is important for `on` method casting on `User` model. In the future, there can be problems with other initializers.

lib/badge.rb
dissident: You can view the latest implementation

lib/badge.rb
dissident: `earned?` return boolean, but `#get_goal_result` not. I think they different. [flay gem](http://ruby.sadi.st/Flay.html) don't find similarities.

app/models/bestowment.rb
```diff
@@ -0,0 +1,92 @@
...
+        )
+      end
+    end
+  end
+
+  def users_have
+    all_users_count = User.count
+    with_this_badge = Bestowment.where(badge_id: badge_id, rank: rank).count
+    (with_this_badge.to_f / all_users_count.to_f) * 100
+  end
```
NuckChorris, line 48: This might get really slow really fast.  Would you mind adding a table to cache the number of Bestowments of a badge?  Something like this would probably suffice:

```
badges {
  String id
  Integer bestowments_count
}
```

config/initializers/zhe_badges.rb
NuckChorris: Makes sense

db/schema.rb
```diff
@@ -262,6 +274,16 @@
...
+  create_table "friendships", force: :cascade do |t|
+    t.integer  "user_id"
+    t.integer  "friend_id"
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+  end
+
+  add_index "friendships", ["friend_id"], name: "index_friendships_on_friend_id", using: :btree
+  add_index "friendships", ["user_id"], name: "index_friendships_on_user_id", using: :btree
+
```
NuckChorris: I don't think you meant to include this table

db/schema.rb
```diff
@@ -823,13 +845,15 @@
...
 
+  add_foreign_key "bestowments", "users"
   add_foreign_key "blocks", "users"
   add_foreign_key "blocks", "users", column: "blocked_id"
   add_foreign_key "comment_likes", "comments"
   add_foreign_key "comment_likes", "users"
   add_foreign_key "comments", "comments", column: "parent_id"
   add_foreign_key "comments", "posts"
   add_foreign_key "comments", "users"
+  add_foreign_key "friendships", "users"
```
NuckChorris: or this foreign key

app/models/bestowment.rb
```diff
@@ -0,0 +1,92 @@
...
+          bestowed_at: DateTime.now,
+          title: badge.title,
+          description: badge.description
+        )
+      else
+        unless badge.earned?
+          bestowment.progress = badge.progress
+          bestowment.save
+        end
+      end
```
NuckChorris: It seems like much of this logic could be consolidated and simplified with a `first_or_create` combined with an `update`?

lib/badge.rb
NuckChorris: If I may offer a recommendation, this might be simpler if handled with a `Badge#previous_rank`, then you can just do something like this:

``` ruby
class Bestowment
  def show_progress?
    previous_bestowment.earned?
  end

  def previous_bestowment
    Bestowment.where(badge_id: badge.previous_rank, user: user)
  end
end
```

lib/badge.rb
```diff
@@ -0,0 +1,100 @@
...
+    # bestowed previous rank if it not earned
+    prev = @rank - 1
+    if prev.positive? && has_progress? && check_bestowment(prev).nil?
+      bestowed(@rank - 1)
+    end
+  end
+
+  def has_progress?
+    self.class.ranks.present?
+  end
```
NuckChorris, line 49: `progress` is not dependent on whether a badge has `ranks`.  `progress` is a completely different thing.

lib/badge.rb
```diff
@@ -0,0 +1,100 @@
...
+    describe_context
+    @bestowment = check_bestowment(@rank)
+  end
+
+  def check_bestowment(rank = nil)
+    bestowment = Bestowment.where(
+      badge_id: self.class,
+      user: @user
+    )
+    bestowment = bestowment.where(rank: rank) if has_progress?
```
NuckChorris: I know it's not clear, but each rank is a separate class, as expressed in the `DSL` class.

So if there's 5 ranks set to run `on LibraryEntry`, all 5 should execute their `#run` methods.

lib/badge.rb
```diff
@@ -0,0 +1,100 @@
...
+        current_goal = get_goal_result(state.goal)
+        next unless progress < current_goal
+        @rank = state.rank
+        @title = state.title
+        @description = state.description
+        @goal = current_goal
+        stop = true
+      end
+    end
+  end
```
NuckChorris, line 58: This logic seems a bit heavy, and generally having both a `break` and a `next` in a loop is a sign you've gone too far.  Perhaps easier would be to provide a `Badge#previous_rank` method returning the badge class of the previous rank, so all you need to do is:
1. Check `Bestowment.find_by(badge_id: self.previous_rank.class.name, user: user).bestowed?`
2. If true: `Bestowment.where(badge_id: self.class.name, user: user).first_or_initialize`, generate the `progress` and `save!`
3. If false: return

NuckChorris (changes requested):

lib/badge.rb
NuckChorris: (this has serious implications for the logic, but without it users could wind up in an indeterminate state)

NuckChorris (commented):
From reading the code, it seems that all the logic is where it should be.  I'll check it out locally and take it for a spin in morning, but it looks good :ok_hand:

NuckChorris:
Ah, could you rebase it on top of `the-future` to resolve the conflicts?

dissident:
yes, of course

NuckChorris:
Thanks, merged into the badges branch!

PR #14, opened 2016-11-21
PE-376 - video & image preview while typing

emilt27:


NuckChorris (changes requested):
`LongPIpeline` isn't the fastest thing.

It would be better to extract the URLs on the client side as they're pasted/typed, and then request `GET /api/edge/etc/oneboxes?url=https://youtube.com/watch?v=Q3sigAJXG1E` to get the onebox output, from calling `Onebox.preview('https://youtube.com/watch?v=Q3sigAJXG1E').to_s`.

PR #15, opened 2016-11-28
Quotes Feature

toyhammered:
- Base setup of creating files + relationships
  - WIP:
    - Finding character based on name and replacing with id (integer)
    - Character.guess method does not exist (nuck)
    - form supports multi-line
    - make sure db changes are correct (and run migration)


app/models/quote.rb
```diff
@@ -0,0 +1,8 @@
+class Quote < ActiveRecord::Base
+  # defaults to required: true in Rails 5
+  belongs_to :user, required: true
+  belongs_to :anime, required: true
+  belongs_to :character required: true
```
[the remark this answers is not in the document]
toyhammered: oops.. fixed (I blame my hands being freezing cold)

toyhammered:
@NuckChorris, tests that are failing are with the streaming on your end.

db/migrate/20161128200400_update_quotes.rb
```diff
@@ -0,0 +1,11 @@
+class UpdateQuotes < ActiveRecord::Migration
+  def change
+    change_column_null :quotes, :anime_id, false
+    change_column_null :quotes, :user_id, false
+    change_column_null :quotes, :content, false
+
+    add_foreign_key :quotes, :users
+    add_foreign_key :quotes, :anime
+    add_foreign_key :quotes, :characters
```
NuckChorris, line 9: In my experience, you always have some row which fucks shit up on adding a nullability constraint or foreign key.  My advice is: don't add foreign keys to existing data, always handle null values (remove or replace) before adding constraints

db/migrate/20161128201646_create_quote_likes.rb
```diff
@@ -0,0 +1,10 @@
+class CreateQuoteLikes < ActiveRecord::Migration
+  def change
+    create_table :quote_likes do |t|
+      t.references :user, index: true, foreign_key: true, null: false
+      t.references :quote, index: true, foreign_key: true, null: false
+
+      t.timestamps null: false
+    end
```
NuckChorris, line 8: Probably also wanna copy existing data from the `votes` table, like I did when migrating to `post_likes` or `review_likes` (see their migrations for example code to copy+paste)

db/migrate/20161129222053_polymorphize_quotes.rb
```diff
@@ -0,0 +1,8 @@
+class PolymorphizeQuotes < ActiveRecord::Migration
+  def change
+    rename_column :quotes, :anime_id, :media_id
+    add_column :quotes, :media_type, :string, null: false
```
NuckChorris, line 4: When adding a polymorphism, you kinda gotta do it in three steps:

1. add `media_type`
2. fill `media_type` with `Anime`
3. set `null: false`

I think `change_column_null` can handle both 2 and 3 if you use the `default` param

db/schema.rb
```diff
@@ -615,17 +616,29 @@
...
+    t.integer  "user_id",                                null: false
     t.datetime "created_at",                             null: false
     t.datetime "updated_at",                             null: false
-    t.integer  "positive_votes",             default: 0, null: false
+    t.integer  "likes_count",                default: 0, null: false
+    t.string   "media_type",                             null: false
   end
 
-  add_index "quotes", ["anime_id"], name: "index_quotes_on_anime_id", using: :btree
+  add_index "quotes", ["media_id"], name: "index_quotes_on_media_id", using: :btree
```
NuckChorris, line 641: You probably wanna drop this index and add a compound index on `media_type, media_id` (and another on `character_id`)

db/schema.rb
```diff
@@ -873,9 +887,15 @@
...
   add_foreign_key "marathons", "library_entries"
   add_foreign_key "media_follows", "users"
+  add_foreign_key "post_likes", "posts"
+  add_foreign_key "post_likes", "users"
   add_foreign_key "posts", "users"
   add_foreign_key "posts", "users", column: "target_user_id"
-  add_foreign_key "review_likes", "reviews"
+  add_foreign_key "quote_likes", "quotes"
+  add_foreign_key "quote_likes", "users"
+  add_foreign_key "quotes", "anime", column: "media_id"
```
NuckChorris, line 896: This foreign key will prevent the polymorphic association from being a Manga... kinda

Technically as long as the ID is also used by an Anime it'll pass the constraint, but that's not the kind of constraint you want

spec/controllers/users_controller_spec.rb
```diff
@@ -1,3 +1,84 @@
...
+#
+# Indexes
+#
+#  index_users_on_email        (email) UNIQUE
+#  index_users_on_facebook_id  (facebook_id) UNIQUE
+#  index_users_on_to_follow    (to_follow)
+#  index_users_on_waifu_id     (waifu_id)
+#
+# rubocop:enable Metrics/LineLength
+
```
NuckChorris, line 81: Please don't commit these annotations because the `annotate` gem is st00pid

app/models/user.rb
```diff
@@ -104,6 +105,7 @@ class User < ApplicationRecord
   has_many :post_likes, dependent: :destroy
   has_many :review_likes, dependent: :destroy
   has_many :list_imports, dependent: :destroy
+  has_many :quotes, dependent: :destroy
```
NuckChorris, line 108: It would probably make more sense to use `dependent: :nullify` and remove the non-null constraint on `user_id`

If we add user deletion, it makes more sense to disassociate the quotes from a user than to remove the quotes entirely

app/resources/quote_resource.rb
```diff
@@ -0,0 +1,7 @@
+class QuoteResource < BaseResource
+  attributes :content, :positive_votes
```
NuckChorris, line 2: You migrated `positive_votes` to `likes_count` iirc

spec/controllers/users_controller_spec.rb
toyhammered: tell me about it... half the time it deletes the rubocop part but thankfully you teaching me `git add -p` has saved my ass :)

db/migrate/20161129222053_polymorphize_quotes.rb
toyhammered: So I can't do it with something like this?
`add_column :quotes, :media_type, :string, default: 'Anime', null: false`

db/schema.rb
toyhammered: is this not how you do compound index?
`add_index :quotes, [:media_id, :media_type]`

db/migrate/20161128200400_update_quotes.rb
toyhammered: so I should be doing something like:
`Quote.where(media_id: nil).destroy_all` (and the same for content, and media_type?) and then I can do `change_column_null`? (and I won't add foreign key constraints).

db/schema.rb
NuckChorris: Yep that's exactly how you do it

db/migrate/20161128201646_create_quote_likes.rb
toyhammered: This what you talking about (just change to quotes instead of reviews)
```
    execute <<-SQL.squish
      INSERT INTO review_likes (review_id, user_id, created_at, updated_at)
      SELECT target_id, user_id, created_at, updated_at
      FROM votes
      WHERE target_type = 'Review'
      AND positive = 't'
    SQL
```

toyhammered:
Tomorrow morning I am going to migrate this over to a new PR sven and I *might* have (100% did) screw something up with the database.

db/migrate/20161128200400_update_quotes.rb
cybrox: Well, considering `user_id` and `anime_id` already exist, I guess the easiest would be:
```ruby
# Delete all entries with incomplete dataset
Quote.where(anime_id: nil).destroy_all
Quote.where(user_id: nil).destroy_all

# Merge anime_id data over to media_id
rename_column :quotes, :anime_id, :media_id

# Add type for polymorphic relationship
add_column :quotes, :media_type, :string, default: 'Anime', null: false

# Update all existing entries
Quote.update_all(media_type: 'Anime')

# Makes sure no more invalid data can be added
change_column_null :quotes, :user_id, false

# Add reference to characters table
add_reference :quotes, :character, foreign_key: true
```

Just a write-up from the top of my head, maybe @NuckChorris can confirm/correct this real quick.

NuckChorris: I generally like to split the `add_column` from the backfill of `Anime` value, because using `default: 'Anime'` sets an actual database default... which could cause silent and confusing errors down the road if we accidentally use media_id instead of media.  You can use `change_column_null`'s third(I think?) parameter to backfill without setting a default, too

I'd also advise moving the destruction down by `change_column_null` just to have a narrower window of fuckupery

The rest seems pretty much spot-on

db/migrate/20161128201646_create_quote_likes.rb
NuckChorris: Yep

cybrox: Yep, that should work. The old model was called `Quote` as well, for filtering the polymorphic type.

db/migrate/20161128200400_update_quotes.rb
cybrox: You're right, if we leave setting a database default, it will just error out with a _not null constraint_, which is more beneficial to us than assigning the quote to the wrong media type.

Thanks for pointing that out 😅

toyhammered: is this correct?
```
  # Delete all entries with incomplete dataset
    Quote.where(anime_id: nil).destroy_all

    # Merge anime_id data over to media_id
    rename_column :quotes, :anime_id, :media_id

    # Add type for polymorphic relationship
    add_column :quotes, :media_type, :string,
    change_column_null :quotes, :media_type, false, default: 'Anime'

    # Update all existing entries
    Quote.update_all(media_type: 'Anime')

    # Delete all entries with incomplete dataset
    Quote.where(user_id: nil).destroy_all

    # Makes sure no more invalid data can be added
    change_column_null :quotes, :user_id, false

    # Add reference to characters table
    add_reference :quotes, :character, foreign_key: true
  end
```
What I don't get is why are we only adding foreign key to character?

also not sure if this line:
`change_column_null :quotes, :media_type, false, default: 'Anime'` should just be
`change_column_null :quotes, :media_type, false, 'Anime'`

NuckChorris:
What's the difference between this and #17 exactly?

cybrox:
This was the first PR. As far I remember, @toyhammered redid the migration but carried over the code, due to some problems this PR caused. I'll close this, since it was only used as a reference in the end. The new PR will need some refactoring still, to match Josh's new concept.

PR #16, opened 2016-12-01
Change factory associations to build rather than create.

toyhammered:
- can't change follows factory needs user creation.


toyhammered:
9 min 17 seconds build. A bit better I guess.

NuckChorris:
Could you rebase this on top of `the-future` so we can see if the specs pass now?

PR #17, opened 2016-12-02
Add quotes and quote likes

toyhammered:
I have not run migrations, need to make sure they are 100% correct (previous PR had screwed up database).

Keeping #15 open just for reference for now. Will be closed.


cybrox:
Status of this PR: Implementation for the original `Quote` and `QuoteLike` are done. However, this will need some changes to make `Quotes` and `Character` a `HABTM`, in order to match Josh's new concept.

toyhammered:
Does this mean we have to re-re-re do this again, because last time when I tried to change this after doing the migrations everything just went to shit... (@cybrox should remember this). The only thing that could be good is if I set a commit right before I did the `migration` so I can just rollback to that.

NuckChorris:
@cybrox How about we get this initial thing working and then a separate PR can be done later to move to multi-line quotes?  Should be a pretty simple change at any point, once the quotes have a `character_id` in 'em

Also, if I'm not mistaken, all that remains on this is waiting for me to write `Character.guess` and you guys to integrate it, right?

cybrox:
@NuckChorris I guess that'd be possible, yeah. Changing it later on is just a single migration. And yeah, as far I can see, only the guessing method needs to be added.

The question on my end boils down to how much *later* we are going to change the structure, since building the UI for quotes as it is right now and then changing it again wouldn't make much sense. Then again, I haven't seen anyone complain about missing quotes, so we can probably do that change rather quick and then take our time to create a dead-simple UI for multiline quotes.

JoshFabian:
If it isn't a considerable amount of additional work - we should just finish up with multi-line support. Haven't seen much (any) complaints about quotes missing - let's give them a reason to remember them the next time we bring them back.

cybrox:
Server-side, the additional work is really not much, once we have run all of this so we have quotes actually associated to characters.

Client side might be a tiny bit tricky to create an intuitive UI for this, but I do kind of have an idea on how to realize it.

JoshFabian:
IMDB ui for this is actually a pretty good starting point!

trmcnvn:
🏠 🚴

I think that just having a character selection / quote UI with the ability to just add another line until finished is :+1:

NuckChorris:
Rebased and moved to feature/quotes to finish this up 👌

PR #19, opened 2016-12-04
move media to correct status and set progess_limit correctly

toyhammered:
Ajax still needs to be fixed to remove it from the list (if we do this) when the status is changed.

I am also not sure why `status == :completed` does not work, so just changed to a string.

### DIFF
--- a/app/models/library_entry.rb
+++ b/app/models/library_entry.rb
@@ -102,10 +102,10 @@ class LibraryEntry < ApplicationRecord
     if status_changed? && status == 'completed' && media.progress_limit
       # When marked completed, we try to update progress to the cap
       self.progress = media.progress_limit
-    elsif self.progress == media.progress_limit
+    elsif progress == media.progress_limit
       # when in current and episodes changed to 25
       self.status = :completed
-    elsif self.progress != media.progress_limit && status == 'completed'
+    elsif progress != media.progress_limit && status == 'completed'
       # when in completed and episodes changed, strange case
       self.status = :current
     end
--- a/app/models/library_entry.rb
+++ b/app/models/library_entry.rb
@@ -72,8 +72,8 @@ class LibraryEntry < ApplicationRecord
 
   def progress_limit
     return unless progress
-    progress_cap = media.try(:progress_limit)
-    default_cap = media.try(:default_progress_limit)
+    progress_cap = media&.progress_limit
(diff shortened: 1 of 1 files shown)


toyhammered:
screw you timeout error. 🖕

app/models/library_entry.rb
```diff
@@ -99,9 +99,15 @@ def activity
   end
 
   before_save do
-    if status_changed? && status == :completed && media.progress_cap
+    if status_changed? && status == 'completed' && media.progress_limit
       # When marked completed, we try to update progress to the cap
-      self.progress = media.progress_cap
+      self.progress = media.progress_limit
+    elsif progress == media.progress_limit
+      # when in current and episodes changed to 25
```
NuckChorris: This comment is cruel, evil stuff, please don't do that.

It's not "when it changes to 25"

app/models/library_entry.rb
```diff
@@ -99,9 +99,15 @@ def activity
   end
 
   before_save do
-    if status_changed? && status == :completed && media.progress_cap
+    if status_changed? && status == 'completed' && media.progress_limit
```
NuckChorris: I remembered that you can actually just do `&& completed?` due to the way enums work!

app/models/library_entry.rb
toyhammered: sorry sorry did that because i was using sao as my example.

app/models/library_entry.rb
toyhammered: oh, of course totally forgot about that.

toyhammered: wait isn't it `status.completed?`

toyhammered: nvm you were right I thought it would need to know the column name to reference (ie: status) but I guess it is smarter than me 😏

wopian:
>Ajax still needs to be fixed to remove it from the list

~~Is this when it's in CW and you set it to 12/12, it goes to Completed? I vote against that since I also adjust the rating and content disappearing on you is 👎~~

~~Basically the way it works in V2 is 👌~~

I'm dumb

toyhammered:
uhhh that is the way it works in v2? If it is set to 12/12 it goes to completed, if you change status to completed, it goes to completed (and sets to 12/12). Both of these options remove it from the list (in v2).

PR #21, opened 2016-12-13
Scope blocks endpoint to authenticated user

trmcnvn:
So Blocks aren't queryable by everyone.


NuckChorris:
:+1: Can't believe I missed this, I s2g I said to myself "I need to make sure blocks are private"

PR #24, opened 2016-12-16
Include deleted content with reports

trmcnvn:
also add paranoid to reviews


NuckChorris:
Ah! That's a perfect solution I didn't even think of, I was barking up way more complex trees

PR #25, opened 2016-12-18
Mal Auto Updates

toyhammered:
**Feature:** Kitsu will update your mal list so it is in sync with Kitsu at all times

~~**[x]** - when a user updates a library entry it will update mal list also~~
~~**[x]** - when a user initially syncs their account (inputting username/password) we will completely sync    up their mal list to be the same as kitsu list~~
  ~~- [x] will create/update all entries~~
    - [Future] will delete any entries on mal list and not on kitsu list
~~**[x]** - write behavior tests for `get, delete, post, put` methods~~
~~**[x]** - make sure this works with anime and manga~~
  ~~- **[x]** anime~~
  ~~- **[x]** manga~~
~~**[x]** - Encrypt linked_profile tokens (all of them)~~


app/workers/my_anime_list_list_worker.rb
```diff
@@ -0,0 +1,16 @@
+class MyAnimeListListWorker
+  include Sidekiq::Worker
+
+  def perform(user_id)
```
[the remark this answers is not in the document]
toyhammered: hey.. if you read like 9 lines down in the private method IT IS RIGHT THERE?!?!! 🖕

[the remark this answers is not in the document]
NuckChorris: You do realize that the private method can't access the parameter of this, right?  You need to pass it down or set it as an instance variable or property accessor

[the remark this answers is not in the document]
toyhammered: oh I am such a dimwit sometimes...............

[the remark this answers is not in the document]
toyhammered: forgetting ruby basics 101?

toyhammered:
@NuckChorris I have no idea why Webmock is not working for my before_save verify credentials (inside `linked_profile.rb`). I had Sven _attempting_ to help me, I literally copied/pasted the request webmock gave us and it still failed. Any advice on how to fix this?

Also feel free to review what we have done (I will fix the hound errors at some point before the merge)

app/services/my_anime_list_sync_service.rb
```diff
@@ -0,0 +1,203 @@
...
+                episodes: le.progress,
+                score: format_score(le.rating),
+                rewatch_count: le.reconsume_count
+              }
+            )
+          else
+            post(
+              "animelist/#{media_type}",
+              profile,
+              {
```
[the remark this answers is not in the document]
toyhammered: I actually don't understand what this means, it won't work without the `{}`.

app/services/my_anime_list_sync_service.rb
```diff
@@ -0,0 +1,203 @@
...
+
+    case method
+    when 'delete'
+      delete(
+        "#{animelist}/#{mal_media_type}/#{mal_media_id}",
+        profile
+      )
+    else
+      # find the anime or manga
+      get("#{media_type}/#{mal_media_id}#{MINE}", profile) do |response|
```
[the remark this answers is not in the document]
toyhammered: I would get this error... Thanks hound...

app/services/my_anime_list_sync_service.rb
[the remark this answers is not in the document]
NuckChorris: really?  It _should_ work, I'm pretty sure that's valid in Ruby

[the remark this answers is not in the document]
toyhammered:
```
post(
              "animelist/#{media_type}",
              profile,
              anime_id: mal_media_id,
              status: format_status(le.status),
              episodes: le.progress,
              score: format_score(le.rating)
            )
```
looks really strange...

.env
```diff
@@ -23,3 +23,5 @@ STREAM_API_SECRET=uxupxkuw5ckrd2fe9qz88zne4qagwmdm9yaegbsm8xuymqja6ff8jr678636v3
 
 TWITTER_CONSUMER_KEY=some_key
 TWITTER_CONSUMER_SECRET=some_secret
+
+LP_TOKEN_KEY=\xBB_\xAE4\xBB\xDFG\x97\xDA\xAF\xEF\v\xF8\x843\x8B^\x13\xD1\xF5\xB6\x92Y\rT$\xA9\xB5i4\\y
```
NuckChorris: Could you use `Base64.decode64` to read the variable, so we can base64-encode this and not put binary into an environment variable?

Also I'd just call it `ATTR_ENCRYPT_KEY` so we can use it generically

app/models/library_entry.rb
```diff
@@ -104,6 +104,10 @@ def activity
     MediaActivityService.new(self)
   end
 
+  def linked_profile_exists?
```
NuckChorris: This method name doesn't reveal that we're seeking a LinkedProfile _specifically for MAL_

It really ought to have a more specific name than this, perhaps `sync_to_mal?`

app/models/library_entry.rb
```diff
@@ -132,5 +136,20 @@ def activity
         media.increment_rating_frequency(rating)
       end
     end
+
+    # Sync MAL updates if linked profile exists
+    if linked_profile_exists?
+      return unless media_type.in? %w[anime manga]
```
NuckChorris: This will never be true, because the column is capitalized

`%w[Anime Manga]`

app/models/linked_profile.rb
```diff
@@ -31,6 +31,35 @@ class LinkedProfile < ApplicationRecord
...
+    # )
+    # request.run
+
+    # response = request.response
+
+    # if response.code != 200
+    #   errors.add(:token, "#{response.code}: #{response.body}")
+    # end
+
+    # true
```
NuckChorris: Why is this all commented out?

app/services/my_anime_list_sync_service.rb
```diff
@@ -0,0 +1,203 @@
+class MyAnimeListSyncService
+  ATARASHII_API_HOST = 'https://hbv3-mal-api.herokuapp.com/2.1/'.freeze
+  MINE = '?mine=1'.freeze
+
+  attr_reader :le, :method
```
NuckChorris: I'd prefer if you used `library_entry` or `entry` for this.  `le` is not very descriptive, so should be limited to small contexts of just a few lines at most

app/models/linked_profile.rb
```diff
@@ -31,6 +31,35 @@ class LinkedProfile < ApplicationRecord
...
+  end
+
+  after_save do
+    if url == 'myanimelist'
```
NuckChorris: Why are you using `myanimelist` as the URL?  That's not a URL, that's a site name, why not just verify whether the LinkedSite is MAL?

app/services/my_anime_list_sync_service.rb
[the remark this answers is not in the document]
NuckChorris: You can probably split this up a little better, perhaps move the `put` stuff below into methods like `update_animelist(mal_media_id, le)` so that this is more clear what's going on.  Otherwise it's kind of difficult to visually track how indented you are at any given point

app/services/my_anime_list_sync_service.rb
```diff
@@ -0,0 +1,203 @@
...
+        profile
+      )
+    else
+      # find the anime or manga
+      get("#{media_type}/#{mal_media_id}#{MINE}", profile) do |response|
+        # need to check watched_status OR read_status
+        if media_type == 'anime'
+          if response['watched_status']
+            put(
+              "animelist/#{media_type}/#{mal_media_id}",
```
NuckChorris: At this point you know it's an animelist, you don't need `media_type`, you can just directly cram `animelist/anime` in there and avoid the interpolation

app/services/my_anime_list_sync_service.rb
```diff
@@ -0,0 +1,203 @@
...
+    )
+    mal_media_id = mal_media.external_id
+
+    case method
+    when 'delete'
+      delete(
+        "#{animelist}/#{mal_media_type}/#{mal_media_id}",
+        profile
+      )
+    else
```
NuckChorris: It's better if you use `when 'create/update'`, just in the off-chance we need to add more methods down the road.  Plus it's more clear what's going on

app/services/my_anime_list_sync_service.rb
```diff
@@ -0,0 +1,203 @@
...
+    # anime or manga
+    media_type = le.media_type.underscore
+    # convert kitsu data -> mal data
+    mal_media = le.media.mappings.find_by(site: "myanimelist/#{media_type}")
+    # user mal profile information
+    profile = LinkedProfile.find_by(
+      user_id: le.user_id,
+      url: 'myanimelist'
+    )
+    mal_media_id = mal_media.external_id
```
NuckChorris: These four lovelies can probably be moved to their own methods

app/services/my_anime_list_sync_service.rb
```diff
@@ -0,0 +1,203 @@
...
+
+    case method
+    when 'delete'
+      delete(
+        "#{animelist}/#{mal_media_type}/#{mal_media_id}",
```
NuckChorris: >`mal_media_type` is undefined
>`animelist` is undefined

You can probably express this as:

```ruby
delete("#{media_type}list/#{media_type}/#{mal_media_id}", profile)
```

app/services/my_anime_list_sync_service.rb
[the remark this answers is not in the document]
NuckChorris: I think it's more that the indentation you've used (putting the method name on its own line) is really strange.  Keyword parameters and Hash params without curlies are standard procedure in Ruby

app/services/my_anime_list_sync_service.rb
```diff
@@ -0,0 +1,203 @@
...
+            put(
+              "mangalist/#{media_type}/#{mal_media_id}",
+              profile,
+              {
+                status: format_status(le.status),
+                chapters: le.progress,
+                score: format_score(le.rating),
+                reread_count: le.reconsume_count
+              }
+            )
```
NuckChorris: All the things which applied to the `put` for anime above also apply here

app/services/my_anime_list_sync_service.rb
```diff
@@ -0,0 +1,203 @@
...
+            post(
+              "mangalist/#{media_type}",
+              profile,
+              {
+                manga_id: mal_media_id,
+                status: format_status(le.status),
+                chapters: le.progress,
+                score: format_score(le.rating)
+              }
+            )
```
NuckChorris: And also here

app/services/my_anime_list_sync_service.rb
```diff
@@ -0,0 +1,203 @@
...
+    end
+  end
+
+  # if you send no score in
+  # ie: &score&anythingelse
+  # it will not set the score
+  def format_score(score)
+    return nil if score.nil?
+
+    (score * 2).floor
```
NuckChorris: Might be cleaner to just say

```ruby
def format_score(score)
  (score * 2).to_i if score
end
```

There's no reason to floor it, all our scores are .0 or .5, which when multiplied by two guarantees a .0

app/services/my_anime_list_sync_service.rb
```diff
@@ -0,0 +1,203 @@
...
+      method: :post,
+      userpwd: simple_auth(profile),
+      body: body
+    )
+
+    # TODO: @nuck is there a better way to handle this?
+    request_status(request) do |response|
+      # doesn't need to do anything if success
+    end
+    request.run
```
NuckChorris: I'd personally recommend just using Typheous in synchronous mode. If you call `Typhoeus::Request.new('https://example.com').run`, it returns a `Typheous::Response`.

You can call the `response_body`, `response_headers`, or `response_code` methods on the `Typheous::Response` to get info about the response

app/services/my_anime_list_sync_service.rb
```diff
@@ -0,0 +1,203 @@
...
+        $stderr.puts('got a time out')
+      elsif response.code.zero?
+        # Could not get an http response, something's wrong.
+        $stderr.puts(response.return_message)
+      else
+        # Received a non-successful http response.
+        $stderr.puts('HTTP request failed: ' + response.code.to_s)
+      end
+    end
+  end
```
NuckChorris, line 155: It's probably better to raise an exception when we get a non-200 response so that it gets caught by Sentry and emailed to me ;)

app/services/my_anime_list_sync_service.rb
```diff
@@ -0,0 +1,203 @@
...
+    # will return request or nil
+    request.on_complete do |response|
+      if response.success?
+        # this is being sent up to either
+        # get/create/update/delete
+        # afterwards the chosen method will send it up
+        # to the parent request under my_anime_list.rb
+
+        # delete will not have a body so need try catch
+        yield response&.body
```
NuckChorris: I don't think any of this is really necessary (except the basic concept of handling error cases, which is good!), since `post` and the other methods you've created will return the `Typheous::Response` due to ending in `request.run`... you can get the body of the response just by calling `post(things).response_body`

NuckChorris (changes requested):

app/services/my_anime_list_sync_service.rb
toyhammered: yeah... I am so stupid.

app/models/linked_profile.rb
toyhammered: Sven and I were having Webmock issues with testing it, like it flat out would not work (so that code should be valid, was just having testing issues)

app/models/library_entry.rb
toyhammered: ahh yes polymorphic rules...

app/models/linked_profile.rb
toyhammered: yes yes yes... Oh god I screwed up here...

.env
toyhammered: not sure what you mean here.

app/models/library_entry.rb
```diff
@@ -104,6 +104,14 @@ def activity
     MediaActivityService.new(self)
   end
 
+  def sync_to_mal?
+    # is it better to take present? out and pass
+    # this variable into the worker also?
```
NuckChorris: No, it makes more sense to handle it here, so it won't create a bunch of async jobs which do nothing

app/models/library_entry.rb
```diff
@@ -134,5 +142,20 @@ def activity
       user.update_feed_completed!
       user.update_profile_completed!
     end
+
+    # Sync MAL updates if linked profile exists
+    if sync_to_mal?
+      return unless media_type.in? %w[Anime Manga]
```
NuckChorris: Now that this is no longer named `linked_profile?`, maybe move this check into `sync_to_mal?`

Then it's just

```ruby
MyAnimeListSyncWorker.perform_async(self, 'create/update') if sync_to_mal?
```

app/models/linked_profile.rb
```diff
@@ -31,6 +31,35 @@ class LinkedProfile < ApplicationRecord
...
+    end
+
+    true
+  end
+
+  def sync_to_mal?
+    User.find(user_id).linked_profiles.where(
+      linked_site: LinkedSite.find_by(name: 'myanimelist')
+    ).present?
+  end
```
NuckChorris: This seems a bit... roundabout

Couldn't you just do `linked_site.name == 'myanimelist' && token.present?`

app/models/linked_profile.rb
```diff
@@ -31,6 +31,35 @@ class LinkedProfile < ApplicationRecord
   belongs_to :user, required: true
   belongs_to :linked_site, required: true
 
+  attr_encrypted :token, key: Base64.encode64(ENV['ATTR_ENCRYPT_KEY'])
+
   validates_presence_of :url, if: :private?
   validates_presence_of :external_user_id
+  validate :verify_mal_credentials
+
+  def verify_mal_credentials
```
NuckChorris: Make sure you check `sync_to_mal?` before this validation — you should be able to do `validate :verify_mal_credentials, if: :sync_to_mal?`

app/services/my_anime_list_sync_service.rb
```diff
@@ -0,0 +1,166 @@
...
+        else
+          post('animelist/anime', linked_profile,
+            anime_id: mal_media_id,
+            status: format_status(library_entry.status),
+            episodes: library_entry.progress,
+            score: format_score(library_entry.rating))
+        end
+      else
+        # manga
+        if response['read_status']
```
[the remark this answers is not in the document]
NuckChorris: Yo, listen to the dog!

NuckChorris (changes requested):
Much closer, though I almost feel like putting this in `LinkedProfile` was a horrible mistake :/

Like maybe it should be... idk what to call the other model, but like, split the model backing the links on your profile from a separate model backing stuff like this and the Facebook/Twitter login?

It just seems like the logic for what *should* be a simple model (profile links) is gonna get messy really fast as we add more hookups, so if the hookups were separate from the links, it might be simpler to comprehend?

cybrox:
> Like maybe it should be... idk what to call the other model, but like, split the model backing the links on your profile from a separate model backing stuff like this and the Facebook/Twitter login?

I mean, we have iterated over multiple solutions for this in the dev chat before and I think we can agree, that we didn't find a solution that fits quite all needs. With the amount of linkable profiles that are backed by an actual integation (be it oauth or mal's crappy basic auth), is quite high, so separating it would introduce useless roundabouts imo, either for us or also the user, depending on which of the discussed scenarios we pick.

I think it will probably be best to keep linked profiles a single model but before we go too far with it, look at all the main sites we actually planned on integrating and flesh out which properties we need, in order to display and/or actually integrate it.

app/services/my_anime_list_sync_service.rb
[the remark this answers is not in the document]
toyhammered: I actually don't know a nice way to do this... Only reason why I ignored :|

app/models/profile_link.rb
```diff
@@ -0,0 +1,27 @@
+# == Schema Information
+#
+# Table name: profile_links
+#
+#  id                   :integer          not null, primary key
+#  url                  :string           not null
+#  profile_link_site_id :integer          not null, indexed, indexed => [user_id]
```
toyhammered: go... die...

app/services/my_anime_list_sync_service.rb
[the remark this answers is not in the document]
NuckChorris: Instead of doing this:

```ruby
if foo
  # do stuff
else
  if bar
    # do other stuff
  else
    # do other other stuff
  end
end
```

Do this

```ruby
if foo
  # do stuff
elsif bar
  # do other stuff
else
  # do other other stuff
end
```

app/models/linked_account.rb
```diff
@@ -0,0 +1,35 @@
...
+# Foreign Keys
+#
+#  fk_rails_166e103170  (user_id => users.id)
+#
+# rubocop:enable Metrics/LineLength
+
+class LinkedAccount < ApplicationRecord
+  belongs_to :user, required: true
+  # encyrpt the token
+  attr_encrypted :token, key: Base64.encode64(ENV['ATTR_ENCRYPT_KEY'])
```
NuckChorris: Shouldn't this be *decode64*?

app/models/linked_account.rb
```diff
@@ -0,0 +1,35 @@
...
+  # validates_presence_of :url, if: :private?
```
NuckChorris: Can you delete this instead of commenting it out?

app/models/linked_account/my_anime_list.rb
```diff
@@ -0,0 +1,27 @@
...
+      host = MyAnimeListSyncService::ATARASHII_API_HOST
+      response = Typhoeus::Request.new(
+        "#{host}account/verify_credentials",
+        method: :get,
+        userpwd: "#{external_user_id}:#{token}"
+      ).run
+
+      return true if response.code == 200
+
+      errors.add(:token, "#{response.code}: #{response.body}")
```
NuckChorris: This is a terrible error message, check if the error is 403 and give something reasonable like "Username or password was incorrect"

app/models/profile_link_site.rb
```diff
@@ -13,7 +13,7 @@
 #
 # rubocop:enable Metrics/LineLength
```
NuckChorris, line 11: Can probably drop the `share_to`, `share_from` and `link_type` columns tbh

app/resources/profile_link_resource.rb
```diff
@@ -0,0 +1,7 @@
+class ProfileLinkResource < BaseResource
+  immutable
```
NuckChorris: This probably shouldn't be immutable, that would mean nobody can edit the links on their profile

app/resources/linked_account_resource.rb
```diff
@@ -0,0 +1,17 @@
...
+
+  has_one :user
+
+  def fetchable_fields
+    if current_user == _model
+      super
+    else
+      super - [:token]
+    end
+  end
```
NuckChorris: I think this is unnecessary, since the policy can just prevent users from seeing entire LinkedAccounts of other users

app/resources/linked_account_resource.rb
```diff
@@ -0,0 +1,17 @@
+class LinkedAccountResource < BaseResource
+  model_hint model: LinkedAccount::MyAnimeList
+
+  attribute :token
+  attributes :external_user_id, :share_to, :share_from, :sync_to
+  # :type is reserved for STI
```
NuckChorris: For ListImport we do `alias_attribute :kind, :type` in the model, and then expose `kind` as an attribute here.  Also can you combine the two lines here?  There's no reason to have 'em separate

app/resources/linked_account_resource.rb
```diff
@@ -0,0 +1,17 @@
+class LinkedAccountResource < BaseResource
+  model_hint model: LinkedAccount::MyAnimeList
```
NuckChorris, line 2: I wanna say I think the class name needs to be in a string, like `"LinkedAccount::MyAnimeList"`

I'm not sure, check ListImportResource

app/policies/linked_account_policy.rb
```diff
@@ -1,12 +1,12 @@
-class LinkedProfilePolicy < ApplicationPolicy
+class LinkedAccountPolicy < ApplicationPolicy
   alias_method :create?, :is_owner?
   alias_method :update?, :is_owner?
   alias_method :destroy?, :is_owner?
 
   class Scope < Scope
     def resolve
       return scope.where(private: false) unless user
```
NuckChorris: The entire scope method I think can be replaced with `scope.where(user: user)`, since you should only be able to see your own (and admins should not be allowed to peek)

app/services/my_anime_list_sync_service.rb
```diff
@@ -0,0 +1,161 @@
...
+  MINE = '?mine=1'.freeze
+
+  attr_reader :library_entry, :method
+
+  def initialize(library_entry, method)
+    @library_entry = library_entry
+    @method = method
+  end
+
+  def execute_method
```
NuckChorris, line 12: Probably wanna check if `mal_media` is nil here, to handle stuff that's in our anime database but not MAL's

config/routes.rb
```diff
@@ -605,6 +605,7 @@
     jsonapi_resources :roles
     jsonapi_resources :user_roles
     jsonapi_resources :reports
+    jsonapi_resources :linked_accounts
```
NuckChorris, line 608: Can you also add one for `:profile_links`?

NuckChorris (changes requested):

app/resources/profile_link_resource.rb
toyhammered: oh... I mis-thought that. I forgot that someone could add a profile link and also remove it if they don't want it showing anymore... ops.

app/resources/linked_account_resource.rb
toyhammered: List import has it not as a string. Docs are the same. So the way I have it is right.

NuckChorris: Aight then :+1:

app/models/linked_account.rb
toyhammered: I _think_ what are we going to be storing in heroku (will it be an encoded or decoded key?)

NuckChorris: It *will* be encoded as base64 (so we don't store binary data directly), so you need to *decode* it here

NuckChorris: (P.S. how do I generate the key, is it just random data?)

app/models/profile_link_site.rb
toyhammered: are we not going to allow other sites to post stuff to kitsu?

NuckChorris: Sure, but that's part of LinkedAccount, not ProfileLink

app/models/linked_account.rb
toyhammered: yup just random, I went into irb originally to generate the binary data.

app/models/profile_link_site.rb
toyhammered: ahhh ok, wonder why we originally had it there then...

NuckChorris: Because, remember back to like 3 days ago

When the ProfileLink and LinkedAccount were the same system

Before you, yourself, changed it

app/policies/linked_account_policy.rb
toyhammered: yes actually forgot to change this, we already removed private column anyways.

app/services/my_anime_list_sync_service.rb
toyhammered: should i just do an `errors.add` so that we know where issues are?

app/resources/linked_account_resource.rb
toyhammered: so I should just remove the whole `fetchable_fields`? (I personally never understood the point of this, you just said copy it from somewhere lol)

app/models/profile_link_site.rb
toyhammered: such a weird thing we did prior to this, confused the shit outta me.

NuckChorris: You and me both, man.  This new split is much easier to comprehend I think

app/resources/linked_account_resource.rb
NuckChorris: `fetchable_fields` is a method describing which attributes are visible to a given user — so, for example, only you can see your own email even though everybody can see your `User` (they just get `email: nil`)

So when this was combined previously, we had to display the row to all users (if it wasn't private) but hide the `token` field (so people can't steal your MAL password)

Now that it's split, you can hide the entire row from other users using the policy scope, instead of just hiding the `token` field with `fetchable_fields`

app/services/my_anime_list_sync_service.rb
NuckChorris: Can't do that within a service object.  We'll worry about that later, for now just `return` early to skip the sync

app/resources/linked_account_resource.rb
toyhammered: ooooh interesting. So adding the policy scope pretty much replaced the `fetchable_fields`.

NuckChorris: Pretty much.  Really, I'd say what replaced it is the splitting of private (LinkedAccount) from public (ProfileLink)

app/policies/linked_account_policy.rb
```diff
@@ -0,0 +1,11 @@
+class LinkedAccountPolicy < ApplicationPolicy
+  alias_method :create?, :is_owner?
+  alias_method :update?, :is_owner?
+  alias_method :destroy?, :is_owner?
+
+  class Scope < Scope
+    def resolve
+      return scope.where(user: user)
```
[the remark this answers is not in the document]
toyhammered: fixed, just gonna wait on Travis to make sure there is nothing else.

app/models/library_entry.rb
```diff
@@ -134,5 +143,12 @@ def activity
       user.update_feed_completed!
       user.update_profile_completed!
     end
+
+    # Sync MAL updates if linked profile exists
+    MyAnimeListSyncWorker.perform_async(self, 'create/update') if sync_to_mal?
```
NuckChorris: One final thing and then I think we can merge this — you don't wanna pass the whole object into a worker, it gets nasty and messy

Instead of `self`, pass `id`, and then in the worker, do `LibraryEntry.find(library_entry_id)`

NuckChorris (approved):

PR #30, opened 2017-01-07
make sure user actually exists before importing to Anilist

toyhammered:


app/models/list_import/anilist.rb
```diff
@@ -33,6 +33,19 @@ class Anilist < ListImport
     }, presence: true
     # does not accept file uploads
     validates :input_file, absence: true
+    validate :ensure_user_exists, on: :create
+
+    def ensure_user_exists
+      return if input_text.blank?
+
+      @auth_token ||= get_auth_token
+      request = Typhoeus::Request.get(build_url(input_text))
```
NuckChorris, line 42: You know you can use the methods you already have elsewhere here?

So you could remove the `@auth_token` line and just use your `get` method from below I think

toyhammered: that returns the body though, and we need the code.

NuckChorris: Ah, alright then!  I think that's fine then

toyhammered: yeah I tried to use it, but just realized it was not worth me refactoring everything for something that is being run in the background anyway.

PR #35, opened 2017-01-10
Profile Link Changes 🚀

trmcnvn:
Implements changes needed for profile links.


toyhammered:
#36 We also need to add `LinkedAccountsController`

PR #39, opened 2017-01-11
Normalize all Profile Link urls 

toyhammered:
Adding regex to validate all `profile_link` urls to make sure they are correct

### DIFF
--- a/app/models/profile_link.rb
+++ b/app/models/profile_link.rb
@@ -26,4 +26,11 @@ class ProfileLink < ApplicationRecord
   belongs_to :profile_link_site, required: true
 
   validates :url, presence: true
+
+  def url=(value)
+    self.attributes[:url] = value.gsub!(
+      profile_link_site_id.validate_find,
+      profile_link_site_id.validate_replace
+    )
+  end
 end
--- a/app/models/profile_link.rb
+++ b/app/models/profile_link.rb
@@ -27,10 +27,10 @@ class ProfileLink < ApplicationRecord
 
   validates :url, presence: true
 
-  # def url=(value)
-  #   self.attributes[:url] = value.gsub(
-  #     profile_link_site_id.validate_find,
-  #     profile_link_site_id.validate_replace
-  #   )
-  # end
+  def url=(value)
+    self.attributes[:url] = value.gsub(
+      Regexp.new(profile_link_site_id.validate_find),
+      Regexp.new(profile_link_site_id.validate_replace)
+    )
+  end
 end
--- a/app/models/profile_link.rb
+++ b/app/models/profile_link.rb
@@ -28,7 +28,7 @@ class ProfileLink < ApplicationRecord
   validates :url, presence: true
 
   def url=(value)
-    self.attributes[:url] = value.gsub!(
+    self.attributes[:url] = value.gsub(
       profile_link_site_id.validate_find,
       profile_link_site_id.validate_replace
     )
--- a/app/models/profile_link.rb
+++ b/app/models/profile_link.rb
@@ -29,8 +29,8 @@ class ProfileLink < ApplicationRecord
 
   def url=(value)
     self.attributes[:url] = value.gsub(
-      Regexp.new(profile_link_site_id.validate_find),
-      Regexp.new(profile_link_site_id.validate_replace)
+      profile_link_site.validate_find,
+      profile_link_site.validate_replace
     )
   end
 end
--- a/app/models/profile_link.rb
+++ b/app/models/profile_link.rb
@@ -28,7 +28,7 @@ class ProfileLink < ApplicationRecord
   validates :url, presence: true
 
   def url=(value)
-    self.attributes[:url] = value.gsub(
+    super value.gsub(
       profile_link_site.validate_find,
       profile_link_site.validate_replace
     )
--- a/app/models/profile_link.rb
+++ b/app/models/profile_link.rb
@@ -27,10 +27,10 @@ class ProfileLink < ApplicationRecord
 
   validates :url, presence: true
 
-  def url=(value)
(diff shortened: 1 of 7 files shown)


toyhammered:
@cybrox @NuckChorris 
ProfileLinks that I have questions on:

1. Google+: How to best deal with custom urls vs non custom (having + for custom)
2. Battle.net: not sure of proper format (if someone can just give me example urls)
3. Tumblr: I am not even sure how/what to validate for this because of all the different formats
~4. Kickstarter: uhhhhhh guess i'll do this but it is questionable~
5. Patreon: need example urls of what we want specifically

trmcnvn:
Battle.net doesn't actually have URLs, just tags in the format of `<Name>#<Id>`

spec/factories/profile_link_sites.rb
```diff
@@ -13,5 +13,123 @@
 FactoryGirl.define do
   factory :profile_link_site do
     name { Faker::Company.name }
+    validate_find 'use trait'
+    validate_replace 'use trait'
+
+    # rubocop:disable Metrics/LineLength
+    # Twitter
+    trait :twitter do
+      validate_find '(https://)?(www.)?(twitter.com/)?(@)?(?<username>[a-zA-Z0-9_]+)'
```
NuckChorris: I'd recommend `https?` for all of these, so HTTP URLs work too

spec/factories/profile_link_sites.rb
```diff
@@ -13,5 +13,123 @@
...
+
+    # Vimeo
+    trait :vimeo do
+      validate_find '(https://)?(www.)?(vimeo.com/)?(?<username>[a-zA-z0-9_\-]+)'
+      validate_replace 'https://vimeo.com/\k<username>'
+    end
+
+    # Github
+    trait :github do
+      validate_find '(https://)?(www.)?(github.com/)?@?(?<username>[a-zA-z0-9\-]+)'
```
NuckChorris: Is github.com/@nuckchorris even legal? o_o

spec/factories/profile_link_sites.rb
```diff
@@ -13,5 +13,123 @@
...
+
+    # Battlenet
+    trait :battlenet do
+      validate_find '(.+#[0-9]+)'
+      validate_replace '\1'
+    end
+
+    # Steam
+    trait :steam do
+      validate_find '((http|https)://)?(www.)?(steamcommunity.com/id/)?(?<username>[a-zA-z0-9_\-]+)'
```
NuckChorris: see ^^^^ where `https?` is an easier way to express this

spec/factories/profile_link_sites.rb
```diff
@@ -13,5 +13,123 @@
...
+    # Discord
+    trait :discord do
+      validate_find '(.+#[0-9]+)'
+      validate_replace '\1'
+    end
+
+    # Tumblr
+    trait :tumblr do
+      validate_find '((http|https)://)?(www.)?(?<username>[a-zA-z0-9]+)(.tumblr.com)?'
+      validate_replace '\k<username>.tumbler.com'
```
NuckChorris: No E in Tumblr

spec/factories/profile_link_sites.rb
toyhammered: actually it was incase someone did something like @NuckChorris so it would get rid of the @ sign.

spec/factories/profile_link_sites.rb
toyhammered: heh fine ill change them all to `(https)?(://)?` or is `(https://)?` allowed?

NuckChorris: Neither would achieve the desired aim.

The `?` must _directly_ modify the `s`, because `?` means "previous character/grouping, 0 or 1 times"

spec/factories/profile_link_sites.rb
NuckChorris: Awesome!  I like it

app/models/profile_link.rb
```diff
@@ -26,4 +26,11 @@ class ProfileLink < ApplicationRecord
   belongs_to :profile_link_site, required: true
 
   validates :url, presence: true
+
+  def url=(value)
+    self.attributes[:url] = value.gsub(
```
NuckChorris: Try doing `super` instead of `self.attributes[:url] = ` 🤔 

See if that fixes it

toyhammered: nope same error: `* profile_link - Validation failed: Url can't be blank (ActiveRecord::RecordInvalid)`.

app/models/profile_link_site.rb
```diff
@@ -11,5 +11,5 @@
 # rubocop:enable Metrics/LineLength
```
NuckChorris, line 13: Could you reannotate here?  You should have the addition of a `validate_find` and `validate_replace` column across here, the spec, and factory

db/schema.rb
```diff
@@ -966,12 +953,8 @@
   add_foreign_key "posts", "users", column: "target_user_id"
   add_foreign_key "profile_links", "profile_link_sites"
   add_foreign_key "profile_links", "users"
-  add_foreign_key "quote_likes", "quotes"
-  add_foreign_key "quote_likes", "users"
-  add_foreign_key "quotes", "characters"
   add_foreign_key "reports", "users"
   add_foreign_key "reports", "users", column: "moderator_id"
-  add_foreign_key "review_likes", "reviews"
```
NuckChorris, line 1076: I'm not sure where all these came from, tbh, but I keep getting the same so I'll allow it

NuckChorris (commented):
Looks like all the code is fine, could you just reannotate real fast?

NuckChorris (approved):

PR #41, opened 2017-01-13
fix minor errors in Mal Auto Sync

toyhammered:
- attr_encrypted is not working, not decrypting password
  - some super long delays with active jobs (it is almost like it is waiting for another request to be sent in and then firing the previous off)

### DIFF
--- a/app/models/library_entry.rb
+++ b/app/models/library_entry.rb
@@ -200,12 +200,24 @@ class LibraryEntry < ApplicationRecord
 
     media.trending_vote(user, 0.5) if progress_changed?
     media.trending_vote(user, 1.0) if status_changed?
+  end
 
-    # Sync MAL updates if linked profile exists
-    MyAnimeListSyncWorker.perform_async(id, 'create/update') if sync_to_mal?
+  # TODO: will rename this if I think of a better one
+  after_commit :create_or_update_library_entry, on: [:create, :update]
+
+  def create_or_update_library_entry
+    MyAnimeListSyncWorker.perform_async(
+      library_entry_id: id,
+      method: 'create/update'
+    ) if sync_to_mal?
   end
 
   after_destroy do
-    MyAnimeListSyncWorker.perform_async(id, 'delete') if sync_to_mal?
+    MyAnimeListSyncWorker.perform_async(
+      user_id: user_id,
+      media_id: media_id,
+      media_type: media_type,
+      method: 'delete'
+    ) if sync_to_mal?
   end
 end
--- a/app/models/library_entry.rb
+++ b/app/models/library_entry.rb
@@ -206,22 +206,22 @@ class LibraryEntry < ApplicationRecord
   after_commit :create_or_update_library_entry, on: %i[create update]
 
   def create_or_update_library_entry
-    if sync_to_mal?
-      MyAnimeListSyncWorker.perform_async(
-        library_entry_id: id,
-        method: 'create/update'
-      )
-    end
+    return unless sync_to_mal?
+
+    MyAnimeListSyncWorker.perform_async(
+      library_entry_id: id,
+      method: 'create/update'
+    )
   end
 
   after_destroy do
-    if sync_to_mal?
-      MyAnimeListSyncWorker.perform_async(
-        user_id: user_id,
-        media_id: media_id,
-        media_type: media_type,
-        method: 'delete'
-      )
-    end
+    return unless sync_to_mal?
+
+    MyAnimeListSyncWorker.perform_async(
+      user_id: user_id,
+      media_id: media_id,
+      media_type: media_type,
+      method: 'delete'
+    )
   end
 end
--- a/app/models/library_entry.rb
+++ b/app/models/library_entry.rb
@@ -203,9 +203,9 @@ class LibraryEntry < ApplicationRecord
   end
 
   # TODO: will rename this if I think of a better one
-  after_commit :create_or_update_library_entry, on: %i[create update]
+  after_commit :sync_entry_update, on: %i[create update]
 
-  def create_or_update_library_entry
+  def sync_entry_update
     return unless sync_to_mal?
 
     MyAnimeListSyncWorker.perform_async(
--- a/app/models/library_entry.rb
+++ b/app/models/library_entry.rb
@@ -203,21 +203,25 @@ class LibraryEntry < ApplicationRecord
   end
 
   # TODO: will rename this if I think of a better one
-  after_commit :create_or_update_library_entry, on: [:create, :update]
+  after_commit :create_or_update_library_entry, on: %i[create update]
 
   def create_or_update_library_entry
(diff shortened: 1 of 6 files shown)


toyhammered:
Going to keep this PR open for a bit while I do testing for Mal Auto Sync

app/models/library_entry.rb
```diff
@@ -200,12 +200,28 @@ def sync_to_mal?
...
     media.trending_vote(user, 1.0) if status_changed?
+  end
+
+  # TODO: will rename this if I think of a better one
+  after_commit :create_or_update_library_entry, on: %i[create update]
 
-    # Sync MAL updates if linked profile exists
-    MyAnimeListSyncWorker.perform_async(id, 'create/update') if sync_to_mal?
+  def create_or_update_library_entry
+    if sync_to_mal?
```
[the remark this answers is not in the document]
toyhammered: You just told me like 2 lines above that I should not use a `return if` when it is a multi-line statement...

app/models/library_entry.rb
```diff
@@ -200,12 +200,28 @@ def sync_to_mal?
 
     media.trending_vote(user, 0.5) if progress_changed?
     media.trending_vote(user, 1.0) if status_changed?
+  end
+
+  # TODO: will rename this if I think of a better one
```
NuckChorris, line 205: How about `sync_entry_update`? 🤔

app/services/my_anime_list_sync_service.rb
```diff
@@ -11,36 +11,48 @@ def initialize(library_entry, method)
 
   def execute_method
     # TODO: add errors later on if we can't find mal data
-    return if mal_media.nil?
-
     case method
     when 'delete'
-      delete("#{media_type}list/#{media_type}/#{mal_media_id}", linked_profile)
+      media_type_d = library_entry['media_type'].underscore
```
NuckChorris: What's with the `_d` suffix?  :confused:

app/workers/my_anime_list_sync_worker.rb
```diff
@@ -1,10 +1,20 @@
 class MyAnimeListSyncWorker
   include Sidekiq::Worker
 
-  def perform(library_entry_id, method)
+  # data is a hash
+  def perform(data)
```
NuckChorris, line 5: IDK man, this seems kinda janky.  Perhaps just pass in `user_id`, `anime_id` and then `library_entry_id` which can be a special value like `'delete'` to signify a deletion?

```ruby
def perform(user_id, media_type, media_id, library_entry_id)
  if library_entry_id == 'delete'
    # Delete it
  else
    # Sync it
  end
end
```

Params list is a bit long, but at least it's not an opaque hash

app/services/my_anime_list_sync_service.rb
toyhammered: does ruby allow you to have variables names the same as method names? I wasn't sure the order it checked (whether it checked if a local variable exists, then calls method or what it does)

cybrox: Allow you to? Yes. Should you do it? No.
For reference: [check this](http://stackoverflow.com/questions/2985850/how-does-ruby-allow-a-method-and-a-class-with-the-same-name/2986338#2986338)

app/workers/my_anime_list_sync_worker.rb
toyhammered: I disagree with this.. It makes it confusing and also the delete is only being done in the service not the worker. My goal was to make it as similar to the library_entry object so I can treat it _almost_ the same.

NuckChorris: "Almost the same" can be a million times worse than "completely different".  Either make it the same, or make it different.  A muddy middle ground is confusing

app/services/my_anime_list_sync_service.rb
NuckChorris: idc what you name the variables, just don't start using hungarian notation with these inexplicable `_d` suffixes

app/workers/my_anime_list_sync_worker.rb
toyhammered: ok i'll rephrase what I said, It is identical where needed. 
`media = MyAnimeListSyncService.new(library_entry, data['method'])`
`library_entry` is going to be for delete
```
 user_id: 1,
 media_id: 3936,
 media_type: "Anime",
 method: "delete"
```

 create/update:
```
 id: 1,
 user_id: 1,
 media_id: 3936,
 status: 1,
 created_at: Thu, 19 Jan 2017 19:24:07 UTC +00:00,
 updated_at: Thu, 19 Jan 2017 19:32:30 UTC +00:00,
 progress: 4,
 rating: nil,
 private: false,
 notes: nil,
 reconsume_count: 0,
 reconsuming: false,
 media_type: "Anime",
 volumes_owned: 0,
 nsfw: false,
 anime_id: 3936,
 manga_id: nil,
 drama_id: nil>
```

so the only difference is that the method is inside the hash but it isn't even used (I can change it so we do `perform(data, method)` if that will make you happier?

db/schema.rb
```diff
@@ -950,12 +936,8 @@
...
   add_foreign_key "profile_links", "users"
-  add_foreign_key "quote_likes", "quotes"
-  add_foreign_key "quote_likes", "users"
-  add_foreign_key "quotes", "characters"
   add_foreign_key "reports", "users"
   add_foreign_key "reports", "users", column: "moderator_id"
-  add_foreign_key "review_likes", "reviews"
   add_foreign_key "review_likes", "users"
   add_foreign_key "reviews", "library_entries"
   add_foreign_key "streaming_links", "streamers"
```
NuckChorris, line 974: Can you rebase this out?  🤔 How did it even get in here?  You didn't do any migrations, why would you include `db/schema.rb`

toyhammered: oops, bad `git add -p` by me :P

app/services/my_anime_list_sync_service.rb
```diff
@@ -15,32 +15,43 @@ def execute_method
...
+        # anime does not exist in their list
+        post('animelist/anime', linked_account,
           anime_id: mal_media_id,
           status: format_status(library_entry.status),
           episodes: library_entry.progress,
           score: format_score(library_entry.rating))
-      elsif media_type == 'manga' && response['read_status']
-        put("mangalist/manga/#{mal_media_id}", linked_profile,
+      elsif media_type == 'manga' &&
+            (response['id'].nil? || response['read_status'])
```
NuckChorris, line 45: Why is manga different? 🤔  What's `response['id'].nil?` for?

toyhammered: oh forgot to comment it, it is related to the HACK: part.

toyhammered: pretty much the goal of it is, that once it gets fixed it will just continue to work.

PR #47, opened 2017-01-24
add STI automatically for LinkedAccount type

toyhammered:
@NuckChorris is this the correct way to do this?


toyhammered:
@cybrox this is not as simple as I expected (nuck explained to me so will leave this off for now)

PR #52, opened 2017-02-02
[WIP]: Library entry logs for MyAnimeList sync

toyhammered:
Some things I am worried about

1. will the `log` get created in time (like the log being created before the worker is called)
2. is the way I deal with `action_performed` ok, the `action_performed` is kind of a lie with the initial sync (so when you create a `LinkedAccount`). Personally I don't think it was a big deal but just wanted to bring to your attention.

3. I will add the controller (forgot to do that)

### DIFF
--- a/app/models/library_entry.rb
+++ b/app/models/library_entry.rb
@@ -193,21 +193,21 @@ class LibraryEntry < ApplicationRecord
     media.trending_vote(user, 1.0) if status_changed?
   end
 
-  after_commit ->() { sync_entry('create') }, on: :create
-  after_commit ->() { sync_entry('update') }, on: :update
-  after_destroy ->() { sync_entry('delete') }
+  after_commit ->() { sync_entry(:create) }, on: :create
+  after_commit ->() { sync_entry(:update) }, on: :update
+  after_destroy ->() { sync_entry(:delete) }
 
   def sync_to_mal?
     return unless media_type.in? %w[Anime Manga]
 
-    linked_account.present?
+    myanimelist_linked_account.present?
   end
 
   def sync_entry(method)
     return unless sync_to_mal?
 
     # create log
-    library_entry_log = create_log(method)
+    library_entry_log = LibraryEntryLog.create_for(method, library_entry)
 
     MyAnimeListSyncWorker.perform_async(
       # for create & update
@@ -222,24 +222,8 @@ class LibraryEntry < ApplicationRecord
     )
   end
 
-  def create_log(method)
-    LibraryEntryLog.create(
-      media_type: media_type,
-      media_id: media_id,
-      progress: progress,
-      rating: rating,
-      reconsume_count: reconsume_count,
-      reconsuming: reconsuming,
-      status: status,
-      volumes_owned: volumes_owned,
-      # action_performed is either create, update, destroy
-      action_performed: method,
-      linked_account_id: linked_account.id
-    )
-  end
-
-  def linked_account
-    @linked_account ||= User.find(user_id).linked_accounts.find_by(
+  def myanimelist_linked_account
+    @mal_linked_account ||= User.find(user_id).linked_accounts.find_by(
       sync_to: true,
       type: 'LinkedAccount::MyAnimeList'
     )
--- a/app/models/library_entry.rb
+++ b/app/models/library_entry.rb
@@ -147,15 +147,6 @@ class LibraryEntry < ApplicationRecord
     end
   end
 
-  def sync_to_mal?
-    return unless media_type.in? %w[Anime Manga]
-
-    User.find(user_id).linked_accounts.where(
-      sync_to: true,
-      type: 'LinkedAccount::MyAnimeList'
-    ).present?
-  end
-
   before_validation do
     # TEMPORARY: If media is set, copy it to kind_id, otherwise if kind_id is
     # set, copy it to media!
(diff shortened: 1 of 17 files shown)


app/models/library_entry.rb
```diff
@@ -202,24 +193,55 @@ def sync_to_mal?
     media.trending_vote(user, 1.0) if status_changed?
   end
 
-  # TODO: will rename this if I think of a better one
-  after_commit :sync_entry_update, on: %i[create update]
+  after_commit ->() { sync_entry('create') }, on: :create
+  after_commit ->() { sync_entry('update') }, on: :update
+  after_destroy ->() { sync_entry('delete') }
```
NuckChorris: It's better to use a `:symbol` for stuff like this, since symbols are immutable

app/models/library_entry.rb
```diff
@@ -202,24 +193,55 @@ def sync_to_mal?
...
+      method: method,
+      library_entry_log_id: library_entry_log.id
     )
   end
 
-  after_destroy do
-    MyAnimeListSyncWorker.perform_async(
-      user_id: user_id,
-      media_id: media_id,
+  def create_log(method)
```
NuckChorris: It might be better to put this into `LibraryEntryLog` as a class method like `LibraryEntryLog.create_for(method, library_entry)`

app/models/library_entry.rb
```diff
@@ -202,24 +193,55 @@ def sync_to_mal?
...
+      reconsuming: reconsuming,
+      status: status,
+      volumes_owned: volumes_owned,
+      # action_performed is either create, update, destroy
+      action_performed: method,
+      linked_account_id: linked_account.id
+    )
+  end
+
+  def linked_account
```
NuckChorris: This method name is too generic for what it does — this is `myanimelist_linked_account`, not a general linked account

app/models/library_entry_log.rb
```diff
@@ -0,0 +1,14 @@
+class LibraryEntryLog < ApplicationRecord
+  belongs_to :linked_account, required: true
+
+  enum sync_status: %i[pending success error]
+  enum status: {
```
NuckChorris: Would `enum status: LibraryEntry.statuses` work here?

app/workers/my_anime_list_list_worker.rb
```diff
@@ -1,19 +1,45 @@
...
       MyAnimeListSyncWorker.perform_async(
-        library_entry_id: library_entry_id,
-        method: 'create/update'
+        library_entry_id: library_entry.id,
+        # HACK: I am unable to determine the method
+        # through the initial sync because I don't access their
+        # mal list. For now everything will be a create action,
+        # but at some point we can update to get the correct action.
+        # Same issue with the LibraryEntryLog performed_action
+        method: 'create',
```
NuckChorris: Is there even a difference in logic between update and create?  Is it worth even separating them and not just treating it as an "upsert" in all cases

NuckChorris (changes requested):

app/models/library_entry_log.rb
toyhammered: I have no idea, didn't even know you could do that so I will look into.

app/workers/my_anime_list_list_worker.rb
toyhammered: I just separated them incase we decide to add more to this in the future where we require to know the difference between `create` and `update` but if you want me to merge to an `upsert` I can do that.

app/models/library_entry.rb
toyhammered: yeah I can add a `LibraryEntryLog` model and just have it go through that.

app/workers/my_anime_list_list_worker.rb
toyhammered: There is not, so I was thinking what we could do is pass in the method as `'upsert'` and then inside the `execute_method` method I can just update the `method=` setter to be specifically `update` or `create`. Would something like this work?

NuckChorris: I don't know why we can't just log it all as "update" tbh

toyhammered: Because lying isn't good?

NuckChorris: It's not lying if the code is identical

And logically we're "updating their library" either way 🤔 I guess the verbiage isn't ideal though.  I'd rather not try and split the code at log time, since that's even more confusing than having the separate create/update codes... I guess this is fine as-is then

toyhammered: Like if we _really_ wanted to be nitpicky, we would update the `method` right here:
```
      if media_type == 'anime' && response['watched_status']
      elsif media_type == 'anime'
```
so that we get a 100% accurate `action_performed` but I don't think it matters that much.

app/workers/my_anime_list_sync_worker.rb
```diff
@@ -5,16 +5,20 @@ class MyAnimeListSyncWorker
   def perform(data)
     # TODO: make a report per day for what is happening (syncing)
     case data['method']
-    when 'delete'
+    when :delete
```
NuckChorris: Sidekiq parameters get turned into JSON and back, so you won't get a symbol of `:delete`, you'll probably get a string of `'delete'`

toyhammered: so once in the worker, I should change everything back to strings?

NuckChorris (approved):

PR #59, opened 2017-02-07
Profile link validation fix

toyhammered:


app/models/profile_link.rb
```diff
@@ -26,11 +26,18 @@ class ProfileLink < ApplicationRecord
...
-      Regexp.new(profile_link_site.validate_find),
+    # updates the url to the expected output
+    # Twitter, ie: toyhammered -> https://twitter.com/toyhammered
+    self.url = url&.sub(
+      profile_link_site.validate_find,
       profile_link_site.validate_replace
     )
   end
+
+  def updated_url
```
NuckChorris: Why did you name this `updated_url`?  Are you validating that the URL is... updated? 🤔

toyhammered: I wasn't actually sure what to name it (because I suck at naming) but it is checking whether the `updated_url` is valid. Maybe it should be something like `reformatted_url`?

NuckChorris (approved):

PR #62, opened 2017-02-14
Rails Admin

txie1993:
rails_admin gem integration


app/controllers/rails_admin/main_controller.rb
```diff
@@ -0,0 +1,147 @@
+module RailsAdmin
+  class MainController < RailsAdmin::ApplicationController
```
NuckChorris: Is this file directly copy+pasted from RailsAdmin?  What's it here for?

app/controllers/sessions_controller.rb
```diff
@@ -0,0 +1,10 @@
+class SessionsController < ActionController::Base
+
+  def index
```
NuckChorris: What's the `index` method for?  If it's the page that triggers the `#create` JS, shouldn't it be `new`?

app/helpers/rails_admin/application_helper.rb
```diff
@@ -0,0 +1,179 @@
+require 'rails_admin/support/i18n'
+
+module RailsAdmin
+  module ApplicationHelper
```
NuckChorris: Is there a reason this code is, again, duplicated from rails_admin?

config/application.rb
```diff
@@ -16,6 +16,11 @@
 
 module Kitsu
   class Application < Rails::Application
+    config.middleware.use Rack::MethodOverride
```
NuckChorris: This can cause a CSRF issue, since we don't have CSRF tokens.  A malicious website could do this:

``

And boom, josh is dead.  KO'd

Does it work without this?

config/application.rb
```diff
@@ -16,6 +16,11 @@
...
+    config.relative_url_root = "/api"
+    config.action_controller.relative_url_root = "/api"
+    ENV['RAILS_RELATIVE_URL_ROOT']  = "/api"
+    ENV['ROOT_URL']  = "/api"
```
NuckChorris: Please do not edit `ENV`, it'll cause confusing bugs.  `ENV` is environment variables, they're passed in by the OS and we should pretend they're immutable (even if, for some reason, they are not)

config/initializers/session_store.rb
```diff
@@ -0,0 +1 @@
+Kitsu::Application.config.session_store :cookie_store, key: '_your_app_session'
```
NuckChorris: This should probably be something like `key: '_kitsu_admin_session'`

spec/controllers/admin_controller_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
+
+RSpec.describe AdminController, type: :controller do
```
NuckChorris: Please don't include empty specs

NuckChorris (changes requested):

app/controllers/admin_controller.rb
```diff
@@ -0,0 +1,21 @@
+class AdminController < ActionController::Base
+  include Pundit
+  include Doorkeeper
+  include DoorkeeperHelpers
+  helper AdminHelper
+  protect_from_forgery with: :null_session
+  before_action :set_paper_trail_whodunnit
+
+  rescue_from ActionController::InvalidAuthenticityToken do
+    render text: 'Token expired/invalid', status: 498
```
NuckChorris: 498 is a nonstandard HTTP status code, it's probably better to use a 403

app/models/concerns/media.rb
```diff
@@ -39,9 +39,8 @@ module Media
       inverse_of: :media
     has_many :mappings, as: 'media', dependent: :destroy
     has_many :reviews, as: 'media', dependent: :destroy
-    has_many :media_relationships, as: 'source', dependent: :destroy
-    has_many :inverse_media_relationships, as: 'destination',
-                                           dependent: :destroy
+    has_many :media_relationships, class_name: 'MediaRelationship', as: 'source', dependent: :destroy
+    has_many :inverse_media_relationships, class_name: 'MediaRelationship', as: 'destination', dependent: :destroy
```
NuckChorris: Could you rewrap these around 80 characters?  I'm surprised Hound didn't complain

NuckChorris (approved):

PR #65, opened 2017-02-28
Setup feed follows for groups

trmcnvn:
Setups follow relationships for feeds:

`group_aggr` -> `group`
`global` -> `group` (unless closed)
`timeline` -> `group`


trmcnvn:
Now that I re-think this, Should we be following the group feed as we have the unread counter widget on the dashboard?

PR #67, opened 2017-03-07
Feature: Reposting

toyhammered:
WIP: incorporate into feed. Pretty sure I did the migration correctly.

### DIFF
--- a/app/models/post.rb
+++ b/app/models/post.rb
@@ -60,7 +60,7 @@ class Post < ApplicationRecord
   has_many :comments, dependent: :destroy
   has_many :uploads, as: 'owner', dependent: :destroy
   has_one :ama, foreign_key: 'original_post_id'
-  has_many :reposts, dependent: :destroy
+  has_many :reposts, dependent: :delete_all
 
   scope :in_group, ->(group) { where(target_group: group) }
   scope :visible_for, ->(user) {
@@ -147,5 +147,12 @@ class Post < ApplicationRecord
   after_create do
     media.trending_vote(user, 2.0) if media.present?
     GroupUnreadFanoutWorker.perform_async(target_group_id, user_id) if target_group.present?
+  end
+
+  before_destroy do
+    deletions = reposts.pluck(:user_id, :id).map do |user_id, repost_id|
+      [['user', user_id], { foreign_id: "repost:#{repost_id}" }]
+    end
+    ActivityDeletionWorker.perform_async(deletions)
   end
 end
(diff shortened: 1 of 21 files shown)


app/models/post.rb
```diff
@@ -49,6 +49,8 @@ class Post < ApplicationRecord
   belongs_to :spoiled_unit, polymorphic: true
   has_many :post_likes, dependent: :destroy
   has_many :comments, dependent: :destroy
+  # TODO: what happens if the original is destroyed?
```
NuckChorris: That's what `dependent: :destroy` means, "if we destroy this Post, destroy all the Reblogs"

toyhammered: oh no, this is my question. Are we just going to go with if the original is destroyed, destroy all shared posts + comments on those shared posts?

NuckChorris: Reblogs don't have separate comments or likes, so it's really just "do we remove the reblog itself" and the answer is, I think, "hell yes"

toyhammered: oh reblogs are not like an embedded post, it is just a reference to the original, got it.

Also, what about editing the original post? Are we just going to go with once it is reblogged all editing is disabled?

NuckChorris: I think for now the "Edited at" indicator is enough to delay potential abuse, and eventually we'll have an edit history on posts too

app/models/repost.rb
```diff
@@ -0,0 +1,38 @@
...
+# == Schema Information
+#
+# Table name: reposts
+#
+#  id         :integer          not null, primary key
+#  created_at :datetime         not null
+#  updated_at :datetime         not null
+#  post_id    :integer          not null, indexed
+#  user_id    :integer          not null, indexed
+#
```
matthewdias, line 11: do we want to have a unique constraint on these so you can only repost something once?

app/models/repost.rb
```diff
@@ -0,0 +1,38 @@
...
+  belongs_to :user, required: true
+  belongs_to :post, required: true
+
+  def stream_activity
+    user.profile_feed.activities.new(
+      object: post,
+      nsfw: post.nsfw,
+      foreign_id: self,
+      verb: 'repost'
+    )
```
matthewdias, line 36: should we be sending the user here too so that we have the info in following feeds?

app/models/repost.rb
NuckChorris: I mean, that might make sense yeah :thinking_face:

app/models/repost.rb
NuckChorris: It'll be added automatically as `actor` via the `WithActivity` code

matthewdias: ah right, cool

PR #68, opened 2017-03-09
Mal Auto Sync Fixes

toyhammered:
add filters for linked_account_id to logs


toyhammered:
I actually don't think the `filters` is needed if we just use include.

NuckChorris:
Ready for merge?

toyhammered:
@NuckChorris Ready for takeoff!

PR #77, opened 2017-03-18
User Stats

toyhammered:
https://www.notion.so/User-Stats-78ece66c495e41c1929856524309957d

### DIFF
--- a/app/models/library_entry.rb
+++ b/app/models/library_entry.rb
@@ -200,9 +200,21 @@ class LibraryEntry < ApplicationRecord
     media.trending_vote(user, 1.0) if status_changed?
   end
 
-  after_commit ->() { sync_entry(:create) }, on: :create
-  after_commit ->() { sync_entry(:update) }, on: :update
-  after_destroy ->() { sync_entry(:delete) }
+  after_commit(on: :create) do
+    sync_entry(:create) # mal exporter
+    # Stat STI
+    Stat::AnimeGenreBreakdown.increment(user, media.genres)
+  end
+
+  after_commit(on: :update) do
+    sync_entry(:update) # mal exporter
+  end
+
+  after_destroy do
+    sync_entry(:delete) # mal exporter
+    # Stat STI
+    Stat::AnimeGenreBreakdown.decrement(user, media.genres)
+  end
 
   def sync_to_mal?
     return unless media_type.in? %w[Anime Manga]
@@ -234,16 +246,5 @@ class LibraryEntry < ApplicationRecord
       sync_to: true,
       type: 'LinkedAccount::MyAnimeList'
     )
-  end
-
-  after_commit ->() { create_stat }, on: :create
-  after_destroy ->() { delete_stat }
-
-  def create_stat
-    Stat::AnimeGenreBreakdown.increment_genres(user, media.genres)
-  end
-
-  def delete_stat
-    Stat::AnimeGenreBreakdown.decrement_genres(user, media.genres)
   end
 end
--- a/app/models/library_entry.rb
+++ b/app/models/library_entry.rb
@@ -202,18 +202,31 @@ class LibraryEntry < ApplicationRecord
 
   after_commit(on: :create) do
     sync_entry(:create) # mal exporter
-    # Stat STI
-    Stat::AnimeGenreBreakdown.increment(user, media.genres)
   end
 
   after_commit(on: :update) do
     sync_entry(:update) # mal exporter
   end
 
+  after_create do
+    # Stat STI
+    case kind
+    when :anime
+      Stat::AnimeGenreBreakdown.increment(user, media.genres)
+      Stat::AnimeAmountWatched.increment(user, self)
+    when :manga
+    end
+  end
+
   after_destroy do
     sync_entry(:delete) # mal exporter
     # Stat STI
-    Stat::AnimeGenreBreakdown.decrement(user, media.genres)
+    case kind
+    when :anime
+      Stat::AnimeGenreBreakdown.decrement(user, media.genres)
+      Stat::AnimeAmountWatched.decrement(user, self)
+    when :manga
+    end
   end
 
   def sync_to_mal?
--- a/app/models/library_entry.rb
+++ b/app/models/library_entry.rb
@@ -235,4 +235,15 @@ class LibraryEntry < ApplicationRecord
       type: 'LinkedAccount::MyAnimeList'
     )
   end
+
+  after_commit ->() { create_stat }, on: :create
+  after_destroy ->() { delete_stat }
+
+  def create_stat
+    Stat::AnimeGenreBreakdown.increment_genres(user, media.genres)
+  end
+
+  def delete_stat
+    Stat::AnimeGenreBreakdown.decrement_genres(user, media.genres)
+  end
 end
--- a/app/models/library_entry.rb
+++ b/app/models/library_entry.rb
@@ -212,7 +212,7 @@ class LibraryEntry < ApplicationRecord
     # Stat STI
     case kind
     when :anime
-      Stat::AnimeGenreBreakdown.increment(user, media.genres)
+      Stat::AnimeGenreBreakdown.increment(user, self)
       Stat::AnimeAmountWatched.increment(user, self)
     when :manga
     end
@@ -223,7 +223,7 @@ class LibraryEntry < ApplicationRecord
     # Stat STI
     case kind
     when :anime
-      Stat::AnimeGenreBreakdown.decrement(user, media.genres)
+      Stat::AnimeGenreBreakdown.decrement(user, self)
       Stat::AnimeAmountWatched.decrement(user, self)
     when :manga
     end
--- a/app/models/stat.rb
+++ b/app/models/stat.rb
@@ -1,4 +1,4 @@
-class Stat < ActiveRecord::Base
+class Stat < ApplicationRecord
   belongs_to :user, required: true
   # expose for jsonapi
   alias_attribute :kind, :type
--- a/app/models/stat.rb
+++ b/app/models/stat.rb
@@ -3,9 +3,5 @@ class Stat < ApplicationRecord
   # expose for jsonapi
   alias_attribute :kind, :type
 
-  validates_presence_of :type, :stats_data
-
-  def save_record
-    new_record? ? save : update_attribute(:stats_data, stats_data)
-  end
+  validates :type, :stats_data, presence: true
 end
--- a/app/models/stat.rb
+++ b/app/models/stat.rb
@@ -3,5 +3,5 @@ class Stat < ApplicationRecord
   # expose for jsonapi
   alias_attribute :kind, :type
 
-  validates_presence_of :type
+  validates_presence_of :type, :stats_data
 end
--- a/app/models/stat.rb
+++ b/app/models/stat.rb
@@ -1,0 +1,7 @@
+class Stat < ActiveRecord::Base
+  belongs_to :user, required: true
+  # expose for jsonapi
+  alias_attribute :kind, :type
+
+  validates_presence_of :type
+end
--- a/app/models/stat.rb
+++ b/app/models/stat.rb
@@ -4,4 +4,8 @@ class Stat < ApplicationRecord
   alias_attribute :kind, :type
 
   validates_presence_of :type, :stats_data
+
+  def save_record
+    new_record? ? save : update_attribute(:stats_data, stats_data)
+  end
 end
--- a/app/models/stat.rb
+++ b/app/models/stat.rb
@@ -1,7 +1,6 @@
 class Stat < ApplicationRecord
   belongs_to :user, required: true
-  # expose for jsonapi
-  alias_attribute :kind, :type
 
-  validates :type, :stats_data, presence: true
+  validates :type, presence: true
+  # removed validation for stats_data.
 end
--- a/app/models/stat/anime_amount_watched.rb
+++ b/app/models/stat/anime_amount_watched.rb
@@ -1,0 +1,72 @@
+class Stat
+  class AnimeAmountWatched < Stat
+    # fully regenerate data.
+    def recalculate!
+      entries = user.library_entries.eager_load(:anime)
+                    .where.not('anime.episode_length': nil)
+
+      stats_data['all_time'] = {
+        total_anime: entries.count, # all anime
+        total_episodes: entries.sum(:progress), # all episodes
+        total_time: entries.sum(:time_spent) # time spent
+      }
+
+      save!
+    end
+
+    def self.increment(user, library_entry)
+      record = Stat.find_or_initialize_by(
+        user: user,
+        type: 'Stat::AnimeAmountWatched'
+      )
+      record = record.default_stats if record.new_record?
+      record.increment(library_entry)
+    end
+
+    def increment(library_entry)
+      stats_data['all_time']['total_anime'] += 1
+      stats_data['all_time']['total_episodes'] +=
+        progress_difference(library_entry)
+      stats_data['all_time']['total_time'] +=
+        progress_to_time(library_entry, progress_difference(library_entry))
+
+      save!
+    end
+
+    def self.decrement(user, library_entry)
+      record = Stat.find_by(user: user, type: 'Stat::AnimeAmountWatched')
+      # TODO: do we want to raise or return?
+      raise "Stat doesn't exist" unless record
+      record.decrement(library_entry)
+    end
+
+    def decrement(library_entry)
+      stats_data['all_time']['total_anime'] -= 1
+      stats_data['all_time']['total_episodes'] -= library_entry.progress
+      stats_data['all_time']['total_time'] -=
+        progress_to_time(library_entry, library_entry.progress)
+
+      save!
+    end
+
+    def default_stats
+      stats_data['all_time'] = {
+        total_anime: 0,
+        total_episodes: 0,
+        total_time: 0
+      }
+
+      self
+    end
+
+    def progress_difference(le)
+      le.progress_changed? ? (le.progress - le.progress_was) : le.progress
+    end
+
+    def progress_to_time(le, progress)
+      return 0 if le.anime.episode_length.nil?
+
+      progress * le.anime.episode_length
+    end
+  end
+end
--- a/app/models/stat/anime_amount_watched.rb
+++ b/app/models/stat/anime_amount_watched.rb
@@ -15,8 +15,7 @@ class Stat
     end
 
     def self.increment(user, library_entry)
-      record = Stat.find_or_initialize_by(
-        user: user,
+      record = user.stats.find_or_initialize_by(
         type: 'Stat::AnimeAmountWatched'
       )
       record = record.default_stats if record.new_record?
@@ -34,7 +33,7 @@ class Stat
     end
 
     def self.decrement(user, library_entry)
-      record = Stat.find_by(user: user, type: 'Stat::AnimeAmountWatched')
+      record = user.stats.find_by(type: 'Stat::AnimeAmountWatched')
       # TODO: do we want to raise or return?
       raise "Stat doesn't exist" unless record
       record.decrement(library_entry)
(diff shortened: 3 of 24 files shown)


app/models/library_entry.rb
```diff
@@ -228,4 +228,15 @@ def myanimelist_linked_account
       type: 'LinkedAccount::MyAnimeList'
     )
   end
+
+  after_commit ->() { create_stat }, on: :create
```
toyhammered: I already know this name sucks

app/models/library_entry.rb
```diff
@@ -228,4 +228,15 @@ def myanimelist_linked_account
...
+  after_destroy ->() { delete_stat }
```
toyhammered: and this one

app/models/stat/anime_genre_breakdown.rb
```diff
@@ -2,38 +2,43 @@ class Stat
...
+        record.stats_data['total'] = 0 unless record.stats_data['total']
 
-    private
+        record.stats_data[genre.slug] += 1
+        record.stats_data['total'] += 1
+      end
 
-    def library_entries
-      user.library_entries.where(media_type: 'Anime')
+      p record
```
toyhammered: will remove

app/models/stat/anime_genre_breakdown.rb
```diff
@@ -0,0 +1,47 @@
...
+      stats_data['total'] = genres.values.reduce(:+)
+
+      if new_record?
+        save
+      else
+        update_attribute(:stats_data, stats_data)
+      end
+    end
+
+    def self.increment_genres(user, genres)
```
toyhammered: I kind of feel like that these methods can just be set to `increment` and `decrement` and then just used across all the other Stats (so they are all consistent, because we already know it is genres due to `type`)

NuckChorris: Good thinking!

app/models/library_entry.rb
NuckChorris: That's... actually not a bad name 🤔 That said, it might be better to just toss them in a block like this, depending on how much logic you'll have in there (and how many of these you need):

```ruby
after_commit(on: :create) do
  Stat::AnimeGenreBreakdown.increment_genres(user, media.genres)
end
```

app/models/stat.rb
```diff
@@ -0,0 +1,7 @@
+class Stat < ApplicationRecord
+  belongs_to :user, required: true
+  # expose for jsonapi
+  alias_attribute :kind, :type
```
NuckChorris: This is gonna need adjustment once #86 is merged

app/models/stat.rb
```diff
@@ -0,0 +1,7 @@
...
+
+  validates_presence_of :type, :stats_data
```
NuckChorris: Generally I prefer `validates :type, :stats_data, presence: true`

app/models/stat/anime_genre_breakdown.rb
```diff
@@ -0,0 +1,47 @@
...
+                   .group(:'genres.slug').count
+
+      self.stats_data = genres
+      stats_data['total'] = genres.values.reduce(:+)
+
+      if new_record?
+        save
+      else
+        update_attribute(:stats_data, stats_data)
+      end
```
NuckChorris: Wouldn't `save` work for both of these cases?

Also we might want `save!` which'll throw an error if the save fails (otherwise this might break silently)

app/models/stat/anime_genre_breakdown.rb
```diff
@@ -0,0 +1,47 @@
...
+    def self.decrement_genres(user, genres)
+      record = find_by(user: user)
+      errors.add(:stat, "Stat doesn't exist, can't go negative.") unless record
+
+      genres.each do |genre|
+        next unless record.stats_data[genre.slug]
+
+        record.stats_data[genre.slug] -= 1
+        record.stats_data['total'] -= 1
+      end
```
NuckChorris: You might be able to combine `increment` and `decrement` into an `update_genres(genres, -1)` method (where the second parameter is either `-1` or `+1` depending on whether you wanna decrement or increment. Just something to think about

app/models/stat/anime_genre_breakdown.rb
toyhammered: hmm, now that I think about it I don't know why I actually split it.

app/models/stat/anime_genre_breakdown.rb
toyhammered: I originally had it with an update_genres and was thinking about going with the (+1 -1, we talked about it) but decided that splitting them up would be best for consistency across all the Stats.

app/models/stat.rb
toyhammered: sure, looked at what you did and it seems good.

app/models/stat/anime_amount_watched.rb
```diff
@@ -0,0 +1,71 @@
...
+
+      self
+    end
+
+    def progress_difference(le)
+      le.progress_changed? ? (le.progress - le.progress_was) : le.progress
+    end
+
+    def progress_to_time(le, progress)
+      return 0 if le.anime.episode_length.nil?
```
toyhammered, line 64: I just assumed we are skipping if there is no `episode_length`?

app/resources/stat_resource.rb
```diff
@@ -0,0 +1,13 @@
+class StatResource < BaseResource
```
toyhammered, line 1: should this be `immutable`, now that I think about it `recalculate!` is run with a rake task and both `increment` + `decrement` are just inside the LibraryEntry.

app/services/regenerate_stat_service.rb
```diff
@@ -0,0 +1,17 @@
+class RegenerateStatService
+  class << self
+    def anime_genre_breakdown
+      User.where(id: LibraryEntry.select(:user_id).where(media_type: 'Anime'))
```
toyhammered: I added the `.where(media-type: 'Anime')` because no point running it if they only have manga library entries.

db/migrate/20170326222205_add_time_spent_to_library_entry.rb
```diff
@@ -0,0 +1,23 @@
...
+  using UpdateInBatches
+
+  self.disable_ddl_transaction!
+
+  def up
+    add_column :library_entries, :time_spent, :integer
+
+    say_with_time 'Filling time_spent column' do
+      LibraryEntry.where(media_type: 'Anime').joins(:anime)
+        .update_in_batches('time_spent = anime.episode_length * progress')
```
toyhammered: what happens if there is no `anime.episode_length` because there is no default (or null constraint) so wouldn't it be `nil * progress` which will cause an error?

spec/factories/stats.rb
```diff
@@ -0,0 +1,7 @@
+FactoryGirl.define do
+  factory :stat do
+    association :user, factory: :user, strategy: :build
+    type 'Stat::AnimeGenreBreakdown'
+    stats_data { { total: 0 } }
```
toyhammered: I am getting this error if `stats_data` is just a `{}`:
```
     Failure/Error: FactoryGirl::Linter.new([factory], :factory).lint!

     FactoryGirl::InvalidFactoryError:
       The following factories are invalid:

       * stat - Validation failed: Stats data can't be blank (ActiveRecord::RecordInvalid)
```

spec/models/stat_spec.rb
```diff
@@ -0,0 +1,8 @@
+require 'rails_helper'
+
+RSpec.describe Stat, type: :model do
+  subject { described_class.new }
+
+  it { should belong_to(:user) }
+  it { should validate_presence_of(:type) }
```
toyhammered, line 7: I'll add validation for `stats_data` also

app/models/library_entry.rb
```diff
@@ -193,9 +193,21 @@ def kind
     media.trending_vote(user, 1.0) if status_changed?
   end
 
-  after_commit ->() { sync_entry(:create) }, on: :create
-  after_commit ->() { sync_entry(:update) }, on: :update
-  after_destroy ->() { sync_entry(:delete) }
+  after_commit(on: :create) do
+    sync_entry(:create) # mal exporter
+    # Stat STI
+    Stat::AnimeGenreBreakdown.increment(user, media.genres)
```
NuckChorris: These probably shouldn't be in `after_commit`, that'll make them run in a separate transaction which could result in them being out-of-sync with this model.  Should be in `after_create` and `after_destroy` hooks

app/models/library_entry.rb
```diff
@@ -193,9 +193,21 @@ def kind
...
+  end
+
+  after_commit(on: :update) do
+    sync_entry(:update) # mal exporter
+  end
+
+  after_destroy do
+    sync_entry(:delete) # mal exporter
+    # Stat STI
+    Stat::AnimeGenreBreakdown.decrement(user, media.genres)
```
NuckChorris: See above

app/models/stat.rb
NuckChorris: You don't need this with the STIResource

app/models/stat/anime_amount_watched.rb
```diff
@@ -0,0 +1,71 @@
...
+        progress_difference(library_entry)
+      stats_data['all_time']['total_time'] +=
+        progress_to_time(library_entry, progress_difference(library_entry))
+
+      save!
+    end
+
+    def self.decrement(user, library_entry)
+      record = user.stats.find_by(type: 'Stat::AnimeAmountWatched')
+      # TODO: do we want to raise or return?
```
NuckChorris: Just do nothing, raising an exception might break library updates

app/models/stat/anime_amount_watched.rb
NuckChorris: Seems reasonable tbh

app/models/stat/anime_genre_breakdown.rb
```diff
@@ -0,0 +1,61 @@
...
+      # TODO: do we want to raise or return?
+      raise "Stat doesn't exist" unless record
+      record.decrement(genres)
+    end
+
+    def decrement(genres)
+      genres.each do |genre|
+        # TODO: do we want to skip, or somehow record this error?
+        # this should never happen but there is a chance of
+        # mistakes happening
```
NuckChorris: Just skip

app/resources/stat_resource.rb
NuckChorris: Actually, yes, that's a good point!

app/services/regenerate_stat_service.rb
NuckChorris: Makes sense

db/migrate/20170326222205_add_time_spent_to_library_entry.rb
NuckChorris: I'm pretty sure it'll just replace it with 0 (since this is Postgres math, not Ruby math), that's what it did last night when I was pulling numbers for josh 🤔 But it's possible that order matters, I did `progress * episode_length`

If you wanna be explicit you could use `coalesce(anime.episode_length, 0)`

spec/factories/stats.rb
NuckChorris: Is that if you do `stats_data {}` or is it with `stats_data { {} }`?

spec/models/stat_spec.rb
NuckChorris: :+1:

spec/factories/stats.rb
toyhammered: first one is uhh... wrong? I have tried with the second one, and there is a default set already to `{}` at a database level.

NuckChorris: First one is really more "ambiguous" than wrong, it could be either a block or a hash (and both are valid for FactoryGirl).

Also the problem is simple:

`{}.blank? #=> true`

You can probably drop the validation since the database provides a default

app/models/library_entry.rb
toyhammered: well the `after_create` is creating some janky timing errors with my tests now....

NuckChorris: Really?  That suggests your tests are wrong, since `after_commit` is just never run during tests (the `commit` never actually happens, instead it just runs a `rollback` after each test)

toyhammered: heh, well I know what is wrong but my _easy_ solution of just doing a `before do subject.recalculate!` fails it seems...

app/models/library_entry.rb
```diff
@@ -193,9 +193,34 @@ def kind
...
+    sync_entry(:create) # mal exporter
+  end
+
+  after_commit(on: :update) do
+    sync_entry(:update) # mal exporter
+  end
+
+  after_create do
+    # Stat STI
+    case media_type
```
NuckChorris: I'd prefer if you used `kind` for this (returns a symbol `:anime`, `:manga`, etc.), since `media_type` is deprecated and slated for eventual removal

app/models/stat/anime_amount_watched.rb
```diff
@@ -0,0 +1,72 @@
...
+      stats_data['all_time']['total_episodes'] -= library_entry.progress
+      stats_data['all_time']['total_time'] -=
+        progress_to_time(library_entry, library_entry.progress)
+
+      save!
+    end
+
+    def default_stats
+      # needs to be a string and NOT a symbol
+      # due to how it is being saved.
```
NuckChorris: I'm pretty sure Rails will handle all the conversions itself?

Also it's probably better for `default_stats` to just be a `DEFAULT_STATS` constant that you `self.stats_data = DEFAULT_STATS if stats_data.blank?`, instead of a method which modifies the instance and returns the self.

When a method has a name like `default_stats`, you expect that it'll return the default stats.  The fact that this method doesn't is a bit of a betrayal of expectations.

app/models/stat/anime_genre_breakdown.rb
```diff
@@ -0,0 +1,57 @@
...
+        next unless stats_data[genre.slug]
+
+        stats_data[genre.slug] -= 1
+        stats_data['total'] -= 1
+      end
+
+      save!
+    end
+
+    def default_stats
```
NuckChorris: See above, re: the other `default_stats` method

db/migrate/20170326222205_add_time_spent_to_library_entry.rb
```diff
@@ -0,0 +1,23 @@
...
+class AddTimeSpentToLibraryEntry < ActiveRecord::Migration
+  using UpdateInBatches
+
+  self.disable_ddl_transaction!
+
+  def up
+    add_column :library_entries, :time_spent, :integer
+
+    say_with_time 'Filling time_spent column' do
+      LibraryEntry.where(media_type: 'Anime').joins(:anime)
```
NuckChorris: I would prefer if you used `by_kind(:anime)` for this

lib/tasks/user_stats.rake
```diff
@@ -0,0 +1,10 @@
+task user_stats: 'kitsu:user_stats_setup'
```
NuckChorris: Is there a task named `user_stats_setup`, cause I ain't seeing one 🤔

lib/tasks/user_stats.rake
```diff
@@ -0,0 +1,10 @@
...
+
+namespace :kitsu do
+  task :user_stats do
```
NuckChorris: You probably want to use the `environment` dependency so that all of Rails will load for this (otherwise it probably won't be able to access any Rails models)

lib/tasks/user_stats.rake
toyhammered: oh I failed at copying the `stats.rake` :D. it should be `user_stats_setup` (like 2 lines down)

lib/tasks/user_stats.rake
toyhammered: Can I get an English translation of this please?

db/migrate/20170326222205_add_time_spent_to_library_entry.rb
toyhammered: so instead of `joins(:anime)`, use `by_kind(:anime)`? Can I have a tl;dr for what the difference is?

app/models/stat/anime_amount_watched.rb
toyhammered: Okay! I actually wasn't entirely sure how to approach the `default_stats` because I felt like a method wasn't _right_. Thinking about it now, having a constant (frozen) is prob the best method.

app/models/library_entry.rb
toyhammered: Sounds good, there are quite a few other places that I use `media_type` so should I change them to `kind`?

db/migrate/20170326222205_add_time_spent_to_library_entry.rb
toyhammered: Also for the `media_type: 'Anime'` should I change it to `.where(kind: :anime)`?

app/models/library_entry.rb
NuckChorris: Yeah, anywhere you're using `media_type` from LibraryEntry, change to `kind`, so we can gradually wean off the polymorphic column

db/migrate/20170326222205_add_time_spent_to_library_entry.rb
NuckChorris: Nonono, replace `where(media_type: 'Anime')` with `by_kind(:anime)`

As the name implies, `by_kind` gets Library Entries _by kind_ 😉

lib/tasks/user_stats.rake
NuckChorris: uhhhh why are you copying `stats.rake`? That's literally just a thing to configure how we count lines of code with `rake stats` (try running it, it'll tell you the lines of code in Kitsu!)

It shouldn't be named `user_stats_setup` unless it's doing some from of configuration setup (which it isn't)

lib/tasks/user_stats.rake
NuckChorris: TL;DR go look at `kitsu.rake`

app/models/stat/anime_amount_watched.rb
```diff
@@ -0,0 +1,66 @@
+class Stat
+  class AnimeAmountWatched < Stat
+    DEFAULT_STATS = {
```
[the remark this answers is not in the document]
toyhammered: tis ain't possible houndy :(.

lib/tasks/user_stats.rake
```diff
@@ -0,0 +1,8 @@
+namespace :kitsu do
+  task :user_stats => :environment do
```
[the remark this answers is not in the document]
toyhammered: assuming it just wants `user_stats: :environment`?

app/models/stat/anime_amount_watched.rb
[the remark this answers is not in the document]
NuckChorris: Why isn't that possible?  It looks possible to me :v

app/models/stat/anime_genre_breakdown.rb
NuckChorris: This, again.

db/schema.rb
```diff
@@ -11,11 +11,12 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20170325205929) do
 
+ActiveRecord::Schema.define(version: 20170326222205) do
```
NuckChorris: I think this might've gotten messed up in your merge, but I think the blank line is supposed to be _after_ the `ActiveRecord::Schema.define` line 🤔 

(not a big deal tho, it'll get fixed next time rails dumps the schema)

NuckChorris (commented):
Great progress, I like where this is going!

app/models/stat/anime_genre_breakdown.rb
toyhammered: oops... I only fixed 1 of them...

NuckChorris (approved):

PR #80, opened 2017-03-21
Fix next link feed serializer

jonasbleyl:
At the moment the `next` url in the links response has a double `/api` in its path for the feeds requests:

`"links": { "next":"https://kitsu.io/api/api/edge/feeds/..." }`


NuckChorris:
Thanks, I missed that!

PR #86, opened 2017-03-24
Hide STI better

NuckChorris:
So, we've discussed this a few times, fixing the leakage of Ruby class names into the API.

This fixes that by replacing `kind: "LinkedAccount::MyAnimeList"` and `kind: "ListImport::Anilist"` with `kind: "my-anime-list"` and `kind: "anilist"`, which seems a way better way of hiding it.

THAT SAID!  It's likely a breaking change for any sane clients, so we need to discuss how to do this.  I can think of two possible things:

1. Use a different field name temporarily while we switch over
2. Fuck that shit and just flip the client at the same time

I'm pretty sure no third-party clients currently support either of our STI models (ListImport and LinkedAccount), so this should be up to you and me @vevix

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -271,7 +271,7 @@ class User < ApplicationRecord
     update_profile_completed.save!
   end
 
-  after_create do
+  after_commit on: :create do
     # Send Confirmation Email
     UserMailer.confirmation(self).deliver_later
     # Set up feeds
--- a/app/resources/concerns/sti_resource.rb
+++ b/app/resources/concerns/sti_resource.rb
@@ -10,6 +10,7 @@ module STIResource
   end
 
   included do
+    attribute :kind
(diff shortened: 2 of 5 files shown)


matthewdias:
yeah i think im aware of all the public (and a few private) clients rn and none of them use those.

trmcnvn:
I'm good with option 2 but is this still serializing `kind`? We currently use that to display the type of import.

NuckChorris:
It still serializes but instead of leaking the class name directly it'll be kebab-case

NuckChorris:
Oh wait you're right, I missed the attribute call. I'll fix that in morning

trmcnvn:
@NuckChorris Ready on client pending adding the attribute back.

NuckChorris:
We good to go on this?

trmcnvn:
:+1: LGTM

PR #87, opened 2017-03-24
Fix abbreviated_titles field in rails admin

kazzarin:
I think this might fix the mistake I made with #82.


NuckChorris:
It should!  I mean, it's still not ideal (gotta do it in YAML lol) but it should hopefully work

kazzarin:
Ok, that's good to know :smile:

NuckChorris (approved):

PR #88, opened 2017-03-24
Refactor feed filtering to use split feeds

migl:
* Use split feeds to filter activity
* Creating users and media sets the correct follows between feeds
* User activity gets created into the corresponding filtered feed
* Updated Stream Sync Rake tasks to point (and repoint) feeds correctly
* BONUS: Fix code cleaning up unenriched keys (was ignoring symbols)
* BONUS: Check resource permissions based on user ID and not object identity
* BONUS: Set `activity_copy_limit` parameter to 300 as 1000 is not supported for batch follows (Stream API bug)

### DIFF
--- a/app/models/concerns/media.rb
+++ b/app/models/concerns/media.rb
@@ -64,15 +64,30 @@ module Media
     (end_date || Date.today) - start_date if start_date
   end
 
-  def feed
-    @feed ||= Feed.media(self.class.name, id)
+  def posts_feed
+    @posts_feed ||= Feed.media_posts(self.class.name, id)
+  end
+
+  def media_feed
+    @media_feed ||= Feed.media_media(self.class.name, id)
   end
 
   def aggregated_feed
     @aggregated_feed ||= Feed.media_aggr(self.class.name, id)
   end
 
+  def posts_aggregated_feed
+    @posts_aggregated_feed ||= Feed.media_posts_aggr(self.class.name, id)
+  end
+
+  def media_aggregated_feed
+    @media_aggregated_feed ||= Feed.media_media_aggr(self.class.name, id)
+  end
+
   def follow_self
-    aggregated_feed.follow(feed)
+    aggregated_feed.follow(posts_feed)
+    aggregated_feed.follow(media_feed)
+    posts_aggregated_feed.follow(posts_feed)
+    media_aggregated_feed.follow(media_feed)
   end
 end
--- a/app/models/post.rb
+++ b/app/models/post.rb
@@ -70,7 +70,7 @@ class Post < ApplicationRecord
   end
 
   def other_feeds
-    [media&.feed].compact
+    [media&.posts_feed].compact
   end
 
   def notified_feeds
@@ -82,11 +82,11 @@ class Post < ApplicationRecord
 
   def target_feed
     if target_user
-      target_user.aggregated_feed
+      target_user.posts_aggregated_feed
     elsif target_group
       target_group.feed
     else
-      user.feed
+      user.posts_feed
     end
   end
 
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -232,16 +232,36 @@ class User < ApplicationRecord
     self.title == 'Staff' || self.title == 'Mod'
   end
 
-  def feed
-    @feed ||= Feed.user(id)
+  def posts_feed
+    @posts_feed ||= Feed.user_posts(id)
+  end
+
+  def media_feed
+    @media_feed ||= Feed.user_media(id)
   end
 
   def aggregated_feed
     @aggr_feed ||= Feed.user_aggr(id)
   end
 
+  def posts_aggregated_feed
+    @posts_aggr_feed ||= Feed.user_posts_aggr(id)
+  end
+
+  def media_aggregated_feed
+    @media_aggr_feed ||= Feed.user_media_aggr(id)
+  end
+
   def timeline
     @timeline ||= Feed.timeline(id)
+  end
+
+  def posts_timeline
+    @posts_timeline ||= Feed.timeline_posts(id)
+  end
+
+  def media_timeline
+    @media_timeline ||= Feed.timeline_media(id)
   end
 
   def notifications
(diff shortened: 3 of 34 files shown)


migl:
DEPLOYMENT NOTES FOR SPLIT FEEDS
----------

* Before deploying, make sure you have the following feeds in Stream
  - `user_posts`
  - `user_media`
  - `user_aggr` (aggregated)
  - `user_posts_aggr` (aggregated)
  - `user_media_aggr` (aggregated)
  - `media_posts`
  - `media_media`
  - `media_aggr` (aggregated)
  - `media_posts_aggr` (aggregated)
  - `media_media_aggr` (aggregated)
  - `global` (aggregated)
  - `global_posts` (aggregated)
  - `global_media` (aggregated)
  - `timeline` (aggregated)
  - `timeline_posts` (aggregated)
  - `timeline_media` (aggregated)

  Check the aggregation formats in my application 'dev-mig', but they're basically
  the same as the current ones just removing the bits that do not apply, like
  an aggregation based on a `post_id` where there cannot arrive any `post_id`. I had a few
  problems with aggregation formats that were invalid but that Stream accepted and didn't
  report as wrong, which cause activity to not show up at all. We may want to take a second
  look at them.

  `user` and `media` feeds are ignored with these changes, so all the existing
  activity there should be included in the dump script and moved into the corresponding
  `user_posts` and `user_media` feeds. If not done, old activity will still show up
  in `user_aggr` under the current scheme but once users select a filtered feed, old activities
  won't show as part of it.

* After deploying, we need to repoint all the current feeds the correct way.
  The following rake tasks will take care of that:
  - `rake stream:sync:follows`
  - `rake stream:sync:auto_follows`

  Both already existed but were updated so that when ran they add all the missing follows.
  Tasks are idempotent but take into account that the high number of requests will cause Stream
  to slow down its processing so you won't see all the follows immediately setup.
  Basically, the resulting follow scheme will look like:

    `user_aggr => [user_posts, user_media]`
    `user_posts_aggr => [user_posts]`
    `user_media_aggr => [user_media]`
    `media_aggr => [media_posts, media_media]`
    `media_posts_aggr => [media_posts]`
    `media_media_aggr => [media_media]`
    `global => [user_posts, user_media]`
    `global_posts => [user_posts]`
    `global_media => [user_media]`
    `timeline => [user_posts, user_media]`
    `timeline_posts => [user_posts]`
    `timeline_media => [user_media]`

  This would be necessary just to update the current status of follows; new users and media will have
  their follows correctly setup on runtime.

* Ember client will need changes for real-time updates.

  The feed API endpoint has not changed its signature because internally what
  it's doing is to read the `filters => kind` parameter to choose the correct
  split from the requested feed. So if I'm requesting the `timeline` feed
  with the `posts` kind, internally the app will be reading from `timeline_posts`.

  Nevertheless, the client app will need to repoint the real-time updates to
  be read from the correct feed, based on the same logic the server uses.
  In the future, we could refactor this so that the client requests for the
  correct feed based on the filter the user selects.

  NOTE: This hasn't removed all the filtering from the feeds; only the `kind`
  filtering is gone as there are still NSFW and blocked stuff to remove which require the previous
  algorithm to work.

app/models/concerns/media.rb
```diff
@@ -64,15 +64,30 @@ def run_length
...
+  def posts_aggregated_feed
+    @posts_aggregated_feed ||= Feed.media_posts_aggr(self.class.name, id)
+  end
+
+  def media_aggregated_feed
+    @media_aggregated_feed ||= Feed.media_media_aggr(self.class.name, id)
+  end
+
   def follow_self
-    aggregated_feed.follow(feed)
```
NuckChorris, line 76: We probably need to maintain the existing feed groups until the data is all in the new feeds

app/models/post.rb
```diff
@@ -70,7 +70,7 @@ def feed
   end
 
   def other_feeds
-    [media&.feed].compact
+    [media&.posts_feed].compact
```
NuckChorris, line 73: These should probably be dual-write until the data is imported

app/models/user.rb
```diff
@@ -229,18 +229,38 @@ def admin?
...
   end
 
+  def posts_timeline
+    @posts_timeline ||= Feed.timeline_posts(id)
+  end
+
+  def media_timeline
+    @media_timeline ||= Feed.timeline_media(id)
+  end
+
```
NuckChorris, line 266: This list of methods is getting to some silly levels, perhaps an abstraction would help.  Maybe moving the filter-to-feed-name conversion down to the `Feed` layer, so we can do `Feed.timeline(id, filter: :media)` or something?  Would probably help clarify the logic elsewhere too.

Alternatively, and this is a bit crazy, the `Feed` object could be changed to be a "virtual feed" representing a set of filter feeds so that it still _feels_ like interacting with unsplit feeds, but under the hood it's actually using multiple feeds

app/services/feed_query_service.rb
```diff
@@ -2,6 +2,13 @@ class FeedQueryService
   MEDIA_VERBS = %w[updated rated progressed].freeze
   POST_VERBS = %w[post comment follow review].freeze
 
+  KIND_TO_FEED = {
+    media_aggr: { posts: 'media_posts_aggr', media: 'media_media_aggr' },
+    user_aggr: { posts: 'user_posts_aggr', media: 'user_media_aggr' },
+    timeline: { posts: 'timeline_posts', media: 'timeline_media' },
+    global: { posts: 'global_posts', media: 'global_media' }
+  }.freeze
```
NuckChorris, line 10: This might be good to merge into the `FEED_GROUPS` list in `Feed`

app/services/feed_query_service.rb
```diff
@@ -18,12 +25,15 @@ def list
     list = list.mark(mark) if mark
     list = list.sfw if sfw_filter?
     list = list.blocking(blocked)
-    list = list.select(kind_filter[:ratio], &kind_filter[:proc]) if kind_filter
```
NuckChorris, line 21: Probably want the split feeds to be behind a flag, so we can do dual-writes until the data is imported and then flip the switch

PR #90, opened 2017-03-25
Rails admin config fixes

kazzarin:
- Configured the chapter model so the `titles` field should show up properly
- Added the "Show in App" links which should link directly to Kitsu pages
- Added the missing episode, chapter, and volume count fields to the anime and manga models
- Excluded the `library_entries` field from the user model (I think it will fix a crash similar to the ones that were happening with manga)

### DIFF
--- a/config/initializers/rails_admin.rb
+++ b/config/initializers/rails_admin.rb
@@ -31,7 +31,7 @@ RailsAdmin.config do |config| # rubocop:disable Metrics/BlockLength
     delete
     history_show
     show_in_app do
-      only ['Anime', 'Manga', 'Groups', 'User']
+      only %w[Anime Manga Groups User]
     end
   end
 
--- a/config/initializers/rails_admin.rb
+++ b/config/initializers/rails_admin.rb
@@ -76,7 +76,14 @@ RailsAdmin.config do |config| # rubocop:disable Metrics/BlockLength
   end
   config.model('MangaCharacter') { parent Manga }
   config.model('MangaStaff') { parent Manga }
-  config.model('Chapter') { parent Manga }
+  config.model 'Chapter' do
+    parent Manga
+    field :manga_id
+    field(:titles, :serialized) { html_attributes rows: '6', cols: '70' }
+    fields :canonical_title, :number, :synopsis, :published, :volume, :length
+    include_all_fields
+    navigation_label 'Chapters'
+  end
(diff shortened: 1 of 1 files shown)


config/initializers/rails_admin.rb
```diff
@@ -76,7 +80,14 @@ def edit_user_link
   end
   config.model('MangaCharacter') { parent Manga }
   config.model('MangaStaff') { parent Manga }
-  config.model('Chapter') { parent Manga }
+  config.model 'Chapter' do
+    parent Manga
+    field :manga_id
```
NuckChorris: Would this expose `manga_id` as an integer field?  If so, that's probably not the best option 🤔

NuckChorris:
You are a saint for fixing all these issues in rails_admin 🙌

config/initializers/rails_admin.rb
kazzarin: Oh, I wasn't really sure what field type to specify, but looking through the docs now I think `:belongs_to_association ` might work?

NuckChorris: I think you can just do `field :manga` and it'll pick up that it's an association

kazzarin: Ok, I'll change that in a sec :+1:

NuckChorris:
eeeeeexcellent! ❤

kazzarin:
From testing it on staging, it seems the addition of show_in_app broke things and the *_count fields still aren't showing up :confused:

NuckChorris:
Ah yeah, `show_in_app` will probably never work for us because Rails just handles the API and not the actual site 😓 I'll disable that real fast

Also lemme check something on the `_count`, I did something for that so let me see if I can override it

kazzarin:
Ah, that's unfortunate. I'm curious about what went wrong with the `_count` fields :thinking:

NuckChorris:
I had a thing in `ApplicationRecord` that strips columns with the `_count` suffix (and a couple other suffixes), and it was probably overriding this configuration.

Also I think I might have gotten show_in_app to... vaguely work

kazzarin:
Show in app looks like it's working now, though counts still seem to be missing when editing.

NuckChorris:
Indeed, I'll probably have to add explicit exceptions to the rule 🤔

PR #96, opened 2017-04-07
rails admin fixes

matthewdias:
[bug](https://kitsu.io/feedback/bugs/p/rails-admin-config-bugs)

- [x] make list view not terrible
- [x] episodes
- [x] mappings
- [x] streaming links
- [x] media relationships

### DIFF
--- a/app/assets/javascripts/rails_admin/custom/ui.js
+++ b/app/assets/javascripts/rails_admin/custom/ui.js
@@ -1,0 +1,66 @@
+$(document).on('rails_admin.dom_ready', function(){
+  var $table = $('#bulk_form').find('table');
+  var table = $table[0];
+
+  // Abort if there's nothing to do. Don't repeat ourselves, either.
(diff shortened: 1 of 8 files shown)


matthewdias:
Trying to get this to work. https://github.com/sferik/rails_admin/pull/2461

NuckChorris:
Looks like @synthtech beat you to episodes

matthewdias:
nah, what i want to fix in this pr is the media field on episodes (and the others). right now it's just a dropdown, so it tries to load everything and fails. working on changing it to a search field like the non-polymorphic associations have. see https://kitsu.io/feedback/bugs/p/rails-admin-config-bugs

NuckChorris:
Oh, awesome!  Yeah, looks like they did some work on that upstream and then never merged it :(

NuckChorris:
Does it really need all that copy+pasted JS? 🤔 Can we reduce what's copied to the bare minimum?

matthewdias:
ill see what i can strip out.

PR #102, opened 2017-04-11
add library entry logs controller

matthewdias:


matthewdias:
probably don't need this anymore. doesn't hurt tho.

PR #105, opened 2017-04-17
Manga chapters

matthewdias:


app/models/chapter.rb
```diff
@@ -22,8 +26,20 @@
...
 class Chapter < ApplicationRecord
+  include Titleable
+
   belongs_to :manga
 
+  has_attached_file :thumbnail
+
   validates :manga, presence: true
   validates :number, presence: true
+  # validates :volume_number, presence: true
```
matthewdias: do we want to validate this? i'm thinking sometimes we might want to add chapters when volumes aren't out yet.

toyhammered: nope don't validate

NuckChorris: (probably wanna remove this line entirely)

NuckChorris (commented):
Looks awesome, let me know when this is ready for merge!

PR #107, opened 2017-04-20
Activity History Stats

toyhammered:
* add LibraryEvent
* remove all references to Marathon & MarathonEvent

Going to now flesh this out and start building the Stat next.


NuckChorris:
If you get a chance, would you mind rebasing?

app/models/library_entry.rb
```diff
@@ -207,16 +203,32 @@ def kind
...
+    end
+  end
+
+  after_update do
+    # Activity History Stat will be using this
+    # TODO: refactor this somehow?
+    library_event = LibraryEvent.create_for(:updated, self)
+    case kind
+    when :anime
+      Stat::AnimeActivityHistory.increment(user, library_event)
```
NuckChorris: Shouldn't the `AmountConsumed` ones be running on update too? 🤔

app/models/library_event.rb
```diff
@@ -0,0 +1,43 @@
...
+      rating: library_entry&.rating,
+      reconsuming: library_entry.reconsuming,
+      reconsume_count: library_entry.reconsume_count,
+      volumes_owned: library_entry.volumes_owned,
+      time_spent: library_entry.time_spent,
+      status: library_entry.status,
+      # instead of polymorphic media
+      anime_id: library_entry&.anime_id,
+      manga_id: library_entry&.manga_id,
+      drama_id: library_entry&.drama_id,
```
NuckChorris: aren't all of these duplicated in `changed_data`?

db/migrate/20170420035108_create_library_events.rb
```diff
@@ -0,0 +1,33 @@
...
+      t.text :notes
+      t.boolean :nsfw
+      t.boolean :private
+      t.integer :progress
+      t.integer :rating
+      t.boolean :reconsuming
+      t.integer :reconsume_count
+      t.integer :volumes_owned
+      t.integer :time_spent
+      t.integer :status
```
NuckChorris: Again, I really don't know that these are necessary with `changed_data` existing.

app/models/library_entry.rb
toyhammered: 🤔 Trying to think _why_ I wouldn't do that. Will go over the code I did again and get back to you on this.

db/migrate/20170420035108_create_library_events.rb
toyhammered: can you expand on what you mean? 😕

app/models/library_event.rb
toyhammered: Only the first time right? 😕

app/models/library_entry.rb
toyhammered: Ok, `AmountConsumed` should be added to update, my 1 question is what happens if they go from episode 8 (or chapter) -> episode 7. Should I have like an `if` statement to decide whether to `increment` or `decrement`?

NuckChorris: Probably a good idea to calculate the difference (progress - progress was) and use that as the incr/decrease value

app/models/library_event.rb
NuckChorris: What?

db/migrate/20170420035108_create_library_events.rb
NuckChorris: You're storing the changes in their own columns but also in changed_data, but why both?

app/models/library_event.rb
toyhammered: wait... wouldn't the changed data on store those for the first event, then any after would only have what was changed?

app/models/library_entry.rb
toyhammered: yup, can do that!

db/migrate/20170420035108_create_library_events.rb
toyhammered: because if the data doesn't change and we only have them stored in `changed_data` wouldn't we lose these?

app/models/library_event.rb
toyhammered: oh wait... new event every time, you are right.

db/migrate/20170420035108_create_library_events.rb
toyhammered: what fields _should_ I keep is the question.

app/models/library_event.rb
toyhammered: The question would be though, do we need access to anything isde those fields from JSONAPI, because removing those would kill the relationship it has.

toyhammered:
I just realized that I did some editing on favorite_years in here... oops....

app/models/library_event.rb
```diff
@@ -0,0 +1,57 @@
...
+  enum event: %i[added updated]
+  # remove validation of changed_data
+  validates :event, presence: true
+
+  def self.create_for(event, library_entry)
+    LibraryEvent.create!(
+      # instead of polymorphic media
+      anime_id: library_entry&.anime_id,
+      manga_id: library_entry&.manga_id,
+      drama_id: library_entry&.drama_id,
```
NuckChorris, line 47: You don't need `&.`, the library_entry will never be `nil`

PR #114, opened 2017-04-29
Library dates

matthewdias:
[feature request](https://kitsu.io/feedback/feature-requests/p/start-and-finish-date-for-library)


matthewdias:
for these timestamps we’ll probably need to backfill `watched_at` to equal `updated_at`, and backfill `finished_at` to equal `updated_at` for completed entries?

edit: actually i guess only `watched_at` for entries that have `progress > 0`

matthewdias:
Oh wait, can I do that in a migration? 🤔

matthewdias:
I should probably also change `watched_at` to `consumed_at` to be media-agnostic

app/models/library_entry.rb
```diff
@@ -164,13 +167,26 @@ def kind
   end
 
   before_save do
-    if status_changed? && completed? && media&.progress_limit
-      # When marked completed, we try to update progress to the cap
-      self.progress = media.progress_limit
-    elsif !status_changed? && progress == media&.progress_limit
+    if status_changed?
```
toyhammered: please... no.... :sob:

Nested if's are generally a big nono, nested nested if's are.... inhumane.... It might be best to break this out into a few methods.

spec/models/library_entry_spec.rb
```diff
@@ -133,6 +133,52 @@
     end
   end
 
+  describe 'timestamp validation' do
```
toyhammered, line 138: for all the tests, it might be better to have like a `let!(:library_entry) { create(:library_entry, media: anime) }` and then just do something `LibraryEntry.last` to find it and just continue working with that, instead of creating more and more library entries. I don't think any of your tests conflict with each other that if you change the status of 1 of them, or change the progress you will get different results than expected.

Also just thought i'd explain this to you, 
```
  library_entry = build(:library_entry, media: anime, status: :current)
  library_entry.save!
```
can just be consolidated into `library_entry = create(:library_entry, media: anime, status: :current)` for the first instance (create saves it also)

toyhammered (changes requested):
Looking good so far, seems like it just needs some refactoring and once all the code is correct as the final commit we should add a migration that will populate all of this for existing data.

db/migrate/20170430050223_backfill_library_timestamps.rb
```diff
@@ -0,0 +1,11 @@
+class BackfillLibraryTimestamps < ActiveRecord::Migration
```
toyhammered: @NuckChorris is this how you want to do it or rather with your custom `update_in_batches` module?

db/migrate/20170430050223_backfill_library_timestamps.rb
```diff
@@ -0,0 +1,11 @@
...
+
+  def change
+    LibraryEntry.where("progress > 0").find_in_batches do |batch|
```
toyhammered: @NuckChorris is this how you want to do it or rather with your custom update_in_batches module?

matthewdias: maybe I'm wrong, but it looked like `update_in_batches` doesn't take a block, so I wouldn't be able to get references to each entry's `updated_at`

toyhammered: I _think_ it does, it can access itself(self) by just doing something like `.update_in_batches(consumed_at: updated_at)` (but I could be wrong)

matthewdias: yeah I tried it like that first and it didn't work, which is why I switched to `find_in_batches`

NuckChorris: It takes SQL, you can do
```ruby
update_in_batches(<<-SQL)
  consumed_at = updated_at,
  finished_at = CASE WHEN status = #{LibraryEntry.statuses[:completed]} THEN updated_at END
SQL
```

app/models/library_entry.rb
```diff
@@ -12,8 +13,10 @@
 #  rating          :decimal(2, 1)
 #  reconsume_count :integer          default(0), not null
 #  reconsuming     :boolean          default(FALSE), not null
+#  started_at      :datetime
 #  status          :integer          not null, indexed => [user_id]
 #  volumes_owned   :integer          default(0), not null
+#  consumed_at      :datetime
```
NuckChorris: `last_progress_at` or `progress_updated_at` or `progressed_at` might be more clear.

app/models/library_entry.rb
```diff
@@ -164,13 +167,20 @@ def kind
...
+    if status_changed? && completed?
+      # update progress to the cap
+      self.progress = media.progress_limit if media&.progress_limit
+      # set finished_at and started_at for the first consume
+      self.finished_at ||= Time.now
+      self.started_at ||= Time.now
     end
+
+    # When progress equals total episodes
+    self.status = :completed if progress == media&.progress_limit
```
NuckChorris: This caused a bug previously where you couldn't rewatch, which is why we have `!status_changed?` to prevent it from triggering when the status changes iirc

Lemme check the blame to verify

NuckChorris: Just checked, and I added it very explicitly in 5026578, and I remember it being a bugfix, so we should probably not take it out

app/models/library_entry.rb
matthewdias: yeah, that would avoid confusion with reconsumes. ill go with `progressed_at`

app/models/library_entry.rb
matthewdias: ah, gotcha, ill get that back in there

app/models/library_entry.rb
```diff
@@ -164,12 +167,22 @@ def kind
...
-      self.progress = media.progress_limit
-    elsif !status_changed? && progress == media&.progress_limit
-      # When in current and progress equals total episodes
-      self.status = :completed
+    if status_changed? && completed?
+      # update progress to the cap
+      self.progress = media.progress_limit if media&.progress_limit
+      # set finished_at and started_at for the first consume
+      self.started_at ||= Time.now unless imported
+      self.finished_at ||= Time.now unless imported
```
trmcnvn, line 174: What's the reasoning behind gating behind `imported`?

matthewdias: we're importing these values too, so we don't want them to be overridden with `Time.now` based on their status

trmcnvn: Right, but what about a case when the imported data has no values for these, and the `||=` check shouldn't override an already set value since it expands to something like:

```ruby
self.started_at = self.started_at || Time.now
```

matthewdias: ah. hm. i wonder why i was getting new values on imported entries. 🤔  ill test it again, maybe i missed something.

matthewdias: I cannot for the life of me figure out what im missing to get dates to import.

app/services/my_anime_list_sync_service.rb
```diff
@@ -112,6 +120,10 @@ def format_score(score)
     (score / 2).floor if score
   end
 
+  def format_date(date)
```
toyhammered, line 124: put a check in here to make sure date exists or it will fail.

app/models/library_entry.rb
matthewdias: @vevix would it be okay to leave this `imported` check in actually? the dates are allowed to be unset normally, and setting them to import time wouldn't be very valuable.

trmcnvn: Sure :+1:

spec/models/library_entry_spec.rb
```diff
@@ -134,6 +135,49 @@
...
+    let!(:anime) { create(:anime) }
```
toyhammered: you generally don't have `let!` statements inside a describe block, normally just used at top of file. If you need specific for multiple `it` blocks you can do a `before(:each)` block.

spec/models/library_entry_spec.rb
```diff
@@ -134,6 +135,49 @@
...
+    let!(:library_entry) do
+      create(:library_entry, media: Anime.last, status: :current, progress: 1)
```
toyhammered: read above. Also don't reference Anime.last here if it is in the same block. You can just do `media: :anime` because the variable is in the same scope. You would do `Anime.last` if this was in a `before(:each)` or some other block and you didn't have access to the variable, then you would find the entry it created in the database.

spec/models/library_entry_spec.rb
```diff
@@ -34,7 +36,6 @@
 #  index_library_entries_on_user_id_and_media_type_and_media_id  (user_id,media_type,media_id) UNIQUE
 #  index_library_entries_on_user_id_and_status                   (user_id,status)
 #
-# rubocop:enable Metrics/LineLength
 
 require 'rails_helper'
```
toyhammered, line 41: Just a general comment you can create multiple let values at the top with different names. This is where they should all be (starting below `RSpec.describe`)

spec/models/list_import/my_anime_list/row_spec.rb
```diff
@@ -0,0 +1,37 @@
+require 'rails_helper'
+
+RSpec.describe ListImport::MyAnimeList::Row do
+  context 'with anime' do
+    let(:anime) { fixture('list_import/my_anime_list/toy-anime.json') }
```
toyhammered: not sure if I like the way the fixture is being used. I think it might be better to just create your own object with the anime factory?

cc @NuckChorris

spec/models/list_import/my_anime_list/row_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+
+    describe '#started_at' do
+      context 'with empty date' do
+        it 'should return nil' do
+          subject = described_class.new(JSON.parse(anime)[1])
```
toyhammered: subject can be brought to the top as `subject { described_class.new(JSON.parse(anime)) } and you can then do like `subject[0]` (this should work) but this is also related to above.

cc @NuckChorris

spec/models/list_import/my_anime_list/row_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+          expect(subject.started_at).to be_nil
+        end
+      end
+      context 'with valid date' do
+        it 'should return a date object' do
+          subject = described_class.new(JSON.parse(anime)[0])
```
toyhammered: Comment above related to all of these. Just pointing out 1 more.

toyhammered (dismissed):
Looking good! The only issues I have are with your tests.

spec/models/list_import/my_anime_list/row_spec.rb
matthewdias: This is MAL data tho. Is there a way I can do that without a fixture? I didn't think there was.

spec/models/library_entry_spec.rb
matthewdias: With `before(:each)` does it keep referencing the same object so i can modify it across tests? If not is there a way to do that without `let`? Or should I not be doing that at all and create new objects for each `it` block?

spec/models/library_entry_spec.rb
matthewdias: with `let(:anime)` at the top of the `RSpec.describe` block, using `media: :anime` throws `NoMethodError: undefined method 'primary_key' for Symbol:Class`

spec/models/list_import/my_anime_list/row_spec.rb
matthewdias: doesn't look like this works
```
Failure/Error: @obj = obj.deep_symbolize_keys
     
     NoMethodError:
       undefined method `deep_symbolize_keys' for #<Array:0x0055c15f2c5918>
```

spec/models/list_import/my_anime_list/row_spec.rb
toyhammered: Yeah overlooked that. I guess this will work not happy about the let statement where it is but w/e.

spec/models/library_entry_spec.rb
toyhammered: oh oops, it should be media: anime (not as a symbol). anime is just a variable.

spec/models/list_import/my_anime_list/row_spec.rb
toyhammered: what........ does this work if you do it as `subject =` where you have yours, and then doing `subject[0]`?

spec/models/library_entry_spec.rb
toyhammered: it will create a new object each time in the `before(:each)` and the way to reference it is `Anime.last`. If you need to keep referencing the same object you can do `before do` so it will only do it once.

spec/models/list_import/my_anime_list/row_spec.rb
matthewdias: nope, the issue here is that `described_class.new` is creating a new `ListImport::MyAnimeList::Row`, which expects one JSON object, rather than an array them. Basically that means the only way it will works is if I initialize it with an element in the array.

app/services/my_anime_list_sync_service.rb
```diff
@@ -112,6 +120,10 @@ def format_score(score)
...
+    date&.strftime('%F') if date
```
NuckChorris: You don't need both of these — `&.` serves the same purpose as `if date`

spec/models/library_entry_spec.rb
NuckChorris: Yeah I have to agree, `let!(:library_entry)` is better than using `@library_entry` in `before`

Both `let` and `let!` provide memoized "variables", but `let` will initialize on first usage while `let!` will be pre-initialized before the specs run

spec/models/library_entry_spec.rb
```diff
@@ -134,6 +135,49 @@
...
+        @library_entry.status = :on_hold
+        @library_entry.save!
+        @library_entry.status = :current
+        @library_entry.save!
+        expect(@library_entry.started_at).to eq(started_at)
+      end
+    end
+    context 'finished_at validation' do
+      before do
+        @library_entry =
```
NuckChorris: See above, re: `let`

spec/models/list_import/my_anime_list/row_spec.rb
NuckChorris: Instead of `subject = `, it's better to declare the subject in the `context` block as `subject { described_class.new(...) }`

app/services/my_anime_list_sync_service.rb
matthewdias: whoops, didn't mean to leave that in

spec/models/list_import/my_anime_list/row_spec.rb
matthewdias: See my comment here https://github.com/hummingbird-me/hummingbird-server/pull/114#discussion_r116649465

spec/models/list_import/my_anime_list/row_spec.rb
NuckChorris: I think what toy means is this:

```ruby
subject { described_class.new(JSON.parse(anime)[1]) }
```

Does that not work?

matthewdias: that would work, but i need to access two different ones. `[1]` which doesn't have dates, and `[0]` which does.

matthewdias: should i just use `let` instead, up at the top?

NuckChorris: You can have separate `subject` within each `context` or `describe` block, they're only applicable within their `do..end` boundary.  The same applies to `let` and `let!` and `before` and `after` and `around`

PR #117, opened 2017-05-02
Site-wide Announcement System

NuckChorris:


app/models/site_announcement.rb
```diff
@@ -0,0 +1,31 @@
...
+
+class SiteAnnouncement < ApplicationRecord
+  include WithActivity
+
+  belongs_to :user, required: true
+
+  validates :title, presence: true
+
+  def stream_activity
+    Feed::SiteAnnouncementsFeed.global.activities.new
```
trmcnvn: Shouldn't this be a per-user notification feed, so we can mark announcements as seen?

trmcnvn: Apparently, it is. nvm :nerd_face:

trmcnvn:
```
User.first.site_announcements_feed.setup!
# => {"target"=>["You can only follow flat feeds, not feeds of type notification"]}:
```

NuckChorris:
It should be a flat feed, we don't need agg— fuck notification feeds are aggregated only

Okay we'll need a second feed then

PR #119, opened 2017-05-07
Update user stats data

toyhammered:
going to be changing some stats_data


NuckChorris:
What's the WIP on this?  It seems to get most of the way there 🤔

NuckChorris (approved):

PR #122, opened 2017-05-12
Category models + import + trending

cinquemb:
I guess we'll need this at some point for the explore pages.

# To import:
 1) ```bundle exec rake db:migrate```
 2) ```bundle exec rake chewy:reset``` if ES is not loaded
 3) download and copy https://kitsu.slack.com/files/cinquemb/F5DK2TAH3/anidb_category.json, and https://kitsu.slack.com/files/cinquemb/F5CGS08HH/anidb_anime_category_assoc.json, to lib/anidb 
 4) ```bundle exec rake importers:anidb:categories``` to import categories
 5) ```bundle exec rake importers:anidb:media_assoc``` to associate categories with media on kitsu


app/models/anime.rb
```diff
@@ -59,6 +59,7 @@ class Anime < ApplicationRecord
   has_many :anime_productions, dependent: :destroy
   has_many :anime_characters, dependent: :destroy
   has_many :anime_staff, dependent: :destroy
+  has_and_belongs_to_many :categories
```
trmcnvn: Should move this to the media concern over each individual model.

db/migrate/20170512081514_add_join_tables_for_categories.rb
```diff
@@ -0,0 +1,7 @@
+class AddJoinTablesForCategories < ActiveRecord::Migration
+  def change
+  	create_join_table :categories, :dramas
```
trmcnvn: Indentation here is too far.

app/models/category.rb
```diff
@@ -0,0 +1,27 @@
...
+#
+#  index_categories_on_anidb_id  (anidb_id)
+#
+
+class Category < ActiveRecord::Base
+  has_many :category_favorite, dependent: :destroy
+  has_and_belongs_to_many :anime
+  has_and_belongs_to_many :manga
+  has_and_belongs_to_many :drama
+end
```
trmcnvn: We should validate that at least one title exists.

Does the titles hstore follow the same format as `Media`? We might just be able to include the `titleable` concern.

db/migrate/20170512042542_create_categories.rb
```diff
@@ -0,0 +1,15 @@
+class CreateCategories < ActiveRecord::Migration
+  def change
+    create_table :categories do |t|
+      t.string :canonical_title, required: true
+      t.integer :anidb_id, index: true
+      t.string :image_content_type, limit: 255
+      t.string :image_file_name, limit: 255
+      t.integer :image_file_size
+      t.datetime :image_updated_at
```
trmcnvn: Shouldn't add these manually.

Check out https://github.com/thoughtbot/paperclip#migrations

trmcnvn (changes requested):
Just some initial comments, I'll download the data and run the import myself when I get a chance over the next few days.

db/migrate/20170512042542_create_categories.rb
cinquemb: Cool thanks, i'll check this out

app/models/category.rb
cinquemb: it doesnt follow the same format as media, in the dump there as far as i could tell, there was just the canonical key inside like: ```"titles": {
            "canonical": "type"
        }```

so maybe we could just remove it.

db/migrate/20170512042542_create_categories.rb
cinquemb: Does paperclip allow urls? There are some existing urls in the dump I was just adding to the file name field  if they were there, and leaving the rest empty.

trmcnvn: Yes, you will be able to assign it to the virtual attribute.

```ruby
object.image = <image_path>
```

Upon saving, it'll grab the image and fill in the related fields.

cinquemb: oh ok, awesome.

app/policies/category_policy.rb
```diff
@@ -0,0 +1,2 @@
+class CategoriesPolicy < ApplicationPolicy
```
trmcnvn: Should be `CategoryPolicy`

db/migrate/20170512081514_add_join_tables_for_categories.rb
```diff
@@ -0,0 +1,7 @@
+class AddJoinTablesForCategories < ActiveRecord::Migration
+  def change
+    create_join_table :categories, :dramas
+    create_join_table :anime, :categories
+    create_join_table :categories, :manga
```
trmcnvn, line 5: Is there a reason these don't match up?

lib/tasks/importers.rake
```diff
@@ -1,4 +1,18 @@
...
+  namespace :anidb do
+    require 'anidb/category_importer'
+    require 'anidb/anime_category_assoc_importer'
+    desc 'Import Categories'
+    task :categories, [:quantity] => [:environment] do |_t|
+      puts 'Importing Categories from AniDB dump'
+      AnidbCategoryImport.new().run!
+    end
+    desc 'Import Anime Associated Categories'
+    task :anime_assoc, [:quantity] => [:environment] do |_t|
```
trmcnvn: Since we don't pass quantity, we can change these to:

```ruby
task anime_assoc: :environment do
  # ...
end
```

spec/factories/categories.rb
```diff
@@ -0,0 +1,27 @@
...
+#
+# Indexes
+#
+#  index_categories_on_anidb_id   (anidb_id)
+#  index_categories_on_parent_id  (parent_id)
+#
+
+FactoryGirl.define do
+  factory :category do
+    canonical_title Faker::Name.name
```
trmcnvn: Let's change this to

```ruby
canonical_title { Fake::Name.name }
```
Just to remain consistent with other factories

app/resources/category_resource.rb
```diff
@@ -0,0 +1 @@
+class CategoryResource < BaseResource; end
```
trmcnvn: You'll need to list out the attributes available here.

```ruby
attributes :canonical_title, :description
attribute :image, format: :attachment

has_one :parent
has_many :anime
```

lib/anidb/category_importer.rb
```diff
@@ -0,0 +1,44 @@
...
+      @data = JSON.parse(open(filename).read)
+    end
+
+    def apply!
+      data.each do |item|
+        item = item.deep_symbolize_keys
+        puts item[:titles][:canonical]
+
+        category ||= Category.find_by(anidb_id: item[:id]) || Category.new
+        category.canonical_title ||= item[:titles][:canonical]
```
trmcnvn: Anidb titles are all lowercase so lets improve that.

```ruby
item[:titles][:canonical].titleize
```

db/migrate/20170512042542_create_categories.rb
```diff
@@ -0,0 +1,12 @@
+class CreateCategories < ActiveRecord::Migration
+  def change
+    create_table :categories do |t|
+      t.string :canonical_title, required: true, null: false
```
trmcnvn: Can we just rename this to `title`?

trmcnvn (changes requested):
Import ran with no issues :100:

Just a few small changes.

db/migrate/20170512081514_add_join_tables_for_categories.rb
cinquemb: Like all starting with categories first? Cause rails wants it in alphabetical order if using the ```has_and_belongs_to_many``` method http://api.rubyonrails.org/v4.2.7/classes/ActiveRecord/Associations/ClassMethods.html#method-i-has_and_belongs_to_many

trmcnvn: Ah didn't know that, can leave it as-is.

db/migrate/20170512042542_create_categories.rb
cinquemb: Yeah, didn't realize we were using it differently than anidb was at first, so hopefully this will avoid some confusion in the future

trmcnvn (approved):

trmcnvn:
Before we merge this, we'll need to make the importer download & read the JSON import files over the web (Similar to the bcmoe import) as we don't want to commit the files.

I'll upload them and pass some URLs in here as a comment after I've done so.

trmcnvn:
https://media.kitsu.io/import_files/anidb_category.json
https://media.kitsu.io/import_files/anidb_anime_category_assoc.json

trmcnvn:
I've noticed that the Categories only have one level of depth, whereas Anidb has multiple levels of depths.

Example:

Anidb: `Elements -> Action -> Martial Arts -> Ninja`
Import: `Elements -> Action | Martial Arts | Ninja`

We want to match the Anidb style, does the data provide that info?

cinquemb:
Ok, I'll use those urls and do it like the bcmoe import.


That info isn't provided in dump i was parsing. The parent tag data was invited of the tag data with the Anime, and if it defined a parent, I pulled it. (The anidib tag data by itself didn't include any associtions, just the info about each tag)

JoshFabian:
We can have our database moderators manually sort through and apply the correct nesting to the categories from within RailsAdmin provided the structure is there for that on the backend. 

Essentially, a category should be able to have children and it's children may have children.

shane-tw:
@cinquemb Sorry, that was a mistake on my part.  
I'm pretty sure you can completely ignore all parent tags on the anime pages, and just grab child tags.  
The heirarchy can instead be obtained from [this json file](https://gist.github.com/Kenshii/d6b971f0139d68781435d39492f6478a) for anime tags. That'll let you know the parents. 
It should be the same heirarchy regardless of which anime.  
If you need the heirarchy for another media type's tags such as character, etc. [see here](https://gist.github.com/Kenshii/c7d197584eef9e6875257a7b9afc7efe).

cinquemb:
No worries @Kenshii, the dump was pretty useful (thanks!), but I was able to scrape the hierarchy for the tags a couple days ago.

NuckChorris (commented):
I'm wondering what the exact behavior of parent categories is.  For example, given a category tree of A-(parent of)->B-(parent of)->C, if I set C on something does it also set B and A?

The mockup that @JoshFabian linked me seems to imply that in some cases the parent should apply to the child category and in others it might not — for example, "Elements->Comedy" appears to be applicable to media but "Elements" is not.

lib/anidb/media_category_assoc_importer.rb
```diff
@@ -0,0 +1,91 @@
...
+      exclude_ids = []
+      puts 'Associating Anime With Categories From AniDB Dump'
+      data.each do |unfiltered_anime|
+        unfiltered_anime = unfiltered_anime.deep_symbolize_keys
+        next unless unfiltered_anime[:tags]
+        mapping_object = {
+          title: unfiltered_anime[:canonical],
+          episode_count: unfiltered_anime[:episode_count]
+        }
+        puts 'looking up -> ' + unfiltered_anime[:canonical]
```
NuckChorris: I think there's a way to get the MAL ID from the AniDB dump, which might give a better result.  I'm not gonna lie and say `Mapping.guess` doesn't suck, because it massively sucks.

lib/anidb/media_category_assoc_importer.rb
```diff
@@ -0,0 +1,91 @@
...
+        genre_map_file +
+        '?' +
+        utc
+      ).read
+      gc_data = JSON.parse(genre_categeory_map)
+      gc_data.update(gc_data) { |_, v| Category.where(title: v)[0] if v }
+      @genre_category_model = gc_data
+    end
+
+    def associate_media_categores(media)
```
NuckChorris: I think you may have typo'd `categories`

cinquemb:
>I'm wondering what the exact behavior of parent categories is. For example, given a category tree of A-(parent of)->B-(parent of)->C, if I set C on something does it also set B and A?

Right now, If you set C on something, it does not set B and A.

lib/anidb/media_category_assoc_importer.rb
cinquemb: How does it suck exactly? Has it not worked very well for finding matches?

@Kenshii , I didn't see the MAL ID for anime in the AniDB dump, was it there?

lib/anidb/media_category_assoc_importer.rb
cinquemb: Ahh, thanks

lib/anidb/media_category_assoc_importer.rb
NuckChorris: If you look under `resources` in the JSON, there's something like this:

```json
        {
            "url": "http://www.animenewsnetwork.com/encyclopedia/anime.php?id=14",
            "type": "ANN"
        },
        {
            "url": "http://myanimelist.net/anime/290",
            "type": "MAL"
        },
```

And both of those have IDs we can extract and lookup really really fast with `Mapping.lookup`

shane-tw: resources["MAL"] works in the case of anime with a single corresponding MAL ID.  
eg. `http://myanimelist.net/anime/30191 // Get id via regex match.`  
However, some anime have multiple MAL ID's, in which case it might be MAL (S1) or MAL (S2) or MAL (S1-S6) or I've even seen MAL #2 and MAL (2).  
So, by looking at resources that start with MAL, you'd get a list of MAL ID's corresponding to that anime, if known.  
I've no idea how you'd determine what specific AniDB episodes correspond to a MAL season though, because the resource names seem semi-random. All they have in common is that they all start with MAL.

NuckChorris: Might be reasonable to just apply to all of them, since this would be applying categories... let mods sort it out later?

matthewdias:
if i understand correctly, the only thing the hierarchy is used for is displaying the tree and related categories on the explore page. so we shouldn't be applying parents to anything when children are applied. unless @JoshFabian wanted to have children taken into account for trending?
>Related categories, these are the parent, sub or sibling categories. For example, the category `Action` may have the sub-categories of `Air Combat` and `Battle Royale` - these would be shown in the sidebar. If a user were to click and view the `Battle Royale` category, the sidebar would show the related categories `Action`, `Air Combat`.

https://projects.invisionapp.com/share/FDBREWT7Q#/screens/234267273

lib/anidb/media_category_assoc_importer.rb
cinquemb: Are you talking about all the json files in the ```Tags/``` dir of the anidb.7z? I looped over all of ```resources``` for each of the tags in each json file, and I don't see a single one with a type of ```MAL```, nor do I see any ```resources``` that have a key of ```MAL```.  These are the only values that the ```type``` uses across all the JSON files in ```Tags/```: ```[u'Official page (en)', u'TV Tropes', u'VNDB', u'Wiki (en)', u'Wiki (en) #3', u'Wiki (en) #2', u'Wiki (jp) #2', u'Wiki (jp)', u'Official page (jp)', u'TV Tropes #2']```

cinquemb: Ok, so 9532 of 9991 anime have at least one or more MAL urls, so i'll mod the import script to use the MAL lookup agaisnt those to find the kitsu id instead of Mapping.guess

How can I apply all the id's? I mean, I can try all of the MAL ids (if more than one for a given anime), but Kitsu only has 1 to 1 map with anime base on MAL id?

cinquemb: Yup, way faster! ~5min vs ~1hour for associating anime with categories. Yeah this way is way better, thanks!

app/models/category.rb
```diff
@@ -0,0 +1,46 @@
...
+#  anidb_id           :integer          indexed
+#  parent_id          :integer          indexed
+#
+# Indexes
+#
+#  index_categories_on_anidb_id   (anidb_id)
+#  index_categories_on_parent_id  (parent_id)
+#
+
+class Category < ActiveRecord::Base
```
matthewdias: We'll want to add a `slug` and the `Sluggable` concern, based on title.

app/models/category.rb
```diff
@@ -0,0 +1,46 @@
...
+    tiny: '-quality 90 -strip',
+    small: '-quality 75 -strip',
+    medium: '-quality 70 -strip',
+    large: '-quality 60 -strip'
+  }
+  has_and_belongs_to_many :anime
+  has_and_belongs_to_many :manga
+  has_and_belongs_to_many :drama
+  belongs_to :parent, class_name: 'Category', required: false
+
```
matthewdias, line 47: Need a `children` relationship, so that we can request the direct descendants of a category. Probably just want one level of children, not all of the sub-categories. We also need some way to load only the top-level categories, (`Setting`, `Elements`, etc) for the tree view. Not sure how to approach this.

matthewdias (changes requested):
Just a few things we need to add to support client functionality.

app/models/category.rb
trmcnvn: I don't think we need a relationship, handle this similarly to comments. Allow filtering by `parent_id` on the resource.

Then in the client you can:

```
filter[parent_id]=_none
filter[parent_id]=<parent_id>
```

app/resources/category_resource.rb
```diff
@@ -0,0 +1,9 @@
+class CategoryResource < BaseResource
```
matthewdias, line 3: Can we also add a filter by slug, and unlimited pagination?
```
require 'unlimited_paginator`
...
paginator :unlimited
```
`filters :parent_id, :slug`

app/resources/category_resource.rb
```diff
@@ -0,0 +1,15 @@
...
+  attribute :image, format: :attachment
+
+  has_one :parent
+  has_many :anime
+  has_many :drama
+  has_many :manga
+
+  paginator :unlimited
+
+  filter :parent_id, :slug
```
trmcnvn: `filter` -> `filters`

trmcnvn (changes requested):

trmcnvn:
@cinquemb Could you open a new PR targeting the newer branch?

Also we might want to add a `nsfw` boolean field to categories.

cinquemb:
Yeah, I can do both. Will close this old one when it's done

cinquemb:
@vevix, do you think it would make sense to follow `AgeRatings` concern to handle the nsfw?

trmcnvn:
I think a boolean flag is a better approach here, I don't think categories fit into a group of age ratings.

cinquemb:
Ok sounds good.

PR #127, opened 2017-05-19
Media airing status

matthewdias:


app/models/concerns/media.rb
```diff
@@ -1,6 +1,8 @@
 module Media
   extend ActiveSupport::Concern
 
+  STATUSES = %w[nya upcoming current finished].freeze
```
NuckChorris: n-nyaa?  What's this, nya?

app/models/concerns/media.rb
```diff
@@ -45,6 +47,17 @@ module Media
...
+    scope :past, -> { where('start_date <= ?', Date.today) }
+    scope :finished, -> { past.where('end_date < ?', Date.today) }
+    scope :current, -> do
+      past.where('end_date >= ? OR end_date IS ?', Date.today, nil)
+    end
+    scope :future, -> { where('start_date > ?', Date.today) }
+    scope :upcoming, -> do
+      future.where('start_date <= ?', Date.today + 3.months)
+    end
+    scope :nya, -> { future.where('start_date > ?', Date.today + 3.months) }
```
NuckChorris: And here?

app/models/concerns/media.rb
```diff
@@ -45,6 +47,17 @@ module Media
...
     delegate :year, to: :start_date, allow_nil: true
 
+    scope :past, -> { where('start_date <= ?', Date.today) }
+    scope :finished, -> { past.where('end_date < ?', Date.today) }
+    scope :current, -> do
+      past.where('end_date >= ? OR end_date IS ?', Date.today, nil)
+    end
+    scope :future, -> { where('start_date > ?', Date.today) }
+    scope :upcoming, -> do
+      future.where('start_date <= ?', Date.today + 3.months)
```
NuckChorris, line 62: What's the logic here?

app/models/concerns/media.rb
toyhammered: nyaa nyaa!

toyhammered: not yet airing maybe?

matthewdias: Yeah not yet airing. I didn't know what to call it so just went with what the client is already using. Should I change it?

app/models/concerns/media.rb
matthewdias: Upcoming = it starts in the next 3 months.

app/models/concerns/media.rb
matthewdias: Not yet aired = it starts in the future outside of the upcoming range.

matthewdias: Could maybe change this to include all future for queries, but I did it like this to keep consistency with the attribute.

crse: Do we need to be announced for anime that does not have fixed date yet?

matthewdias: could you rephrase that? im not sure what you're asking.

matthewdias: nevermind, I gotcha. yeah that would be a good addition, ill get that in.

matthewdias: or should i just roll that into "not yet aired"? cc @NuckChorris

app/models/concerns/media.rb
NuckChorris: I think `not_yet_airing` would be nicer, there's not a lot of reason to save those like, 10 bytes

PR #129, opened 2017-05-21
Fix issues with User Deletion

trmcnvn:
This still has issues I haven't addressed @NuckChorris.

Sidekiq workers are failing due to models such as `Follow` attempting to unfollow feeds and using the `default_scope` of the user.

What do you think the best way to fix that is? Out of a few tests, it seems the only way I got it to work was:

```ruby
belongs_to :follower, -> { unscope(where: :deleted_at) }, ...
```

_note: `-> { unscoped }` doesn't seem to work due to a rails bug_

But, that seems _incorrect_.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -134,7 +134,20 @@ class User < ApplicationRecord
   has_many :post_follows, dependent: :destroy
   has_many :review_likes, dependent: :destroy
   has_many :list_imports, dependent: :destroy
+  has_many :group_action_logs, dependent: :destroy
+  has_many :group_bans, dependent: :destroy
+  has_many :group_invites, dependent: :destroy
   has_many :group_members, dependent: :destroy
+  has_many :group_reports
+  has_many :group_reports_as_moderator, class_name: 'GroupReport',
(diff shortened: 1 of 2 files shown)


NuckChorris:
What's failing exactly?  Do we not just delete the follows on user deletion?

trmcnvn:
Each of our follow models has an `after_destroy` hook that references the user relation.

```ruby
# follow.rb
after_commit(on: :destroy) do
  follower.timeline.unfollow(followed.profile_feed)
end
```

Due to the `User` model having a `default_scope -> { where(deleted_at: nil) }`, it fails to find the user related to the model.

See Sidekiq:
`Couldn't find User with 'id'=<ID> [WHERE "users"."deleted_at" IS NULL]`

---

Maybe one way we could fix this is drop the default scope and apply that scope to the user policy instead?

---

Edit: Due to the workers failing to delete due to the above issue, posts aren't being reassigned (which happens after the DestructionWorker) so we end up with posts such as https://kitsu.io/posts/8751967

trmcnvn:
Updated this to move the scope to the policy so `DestructionWorker` should complete now.

PR #130, opened 2017-05-24
Bug fix on library entries enpoint

cinquemb:


trmcnvn:
This change caused some issues since `kind !== media_type`.

`media_type` is a capitalized value, while `kind` is lowercase -- this also _might_ conflict with the `#kind` method we already provide in the model.

See #131

cinquemb:
Ahh ok makes sense, i'll close this then

PR #131, opened 2017-05-25
Fix remaining bugs with Library Search

trmcnvn:
:memo:  See commit messages.

Confirmed working :+1:

Closes #130

### DIFF
--- a/app/models/library_entry.rb
+++ b/app/models/library_entry.rb
@@ -61,7 +61,7 @@ class LibraryEntry < ApplicationRecord
   end
   scope :privacy, ->(privacy) { where(private: !(privacy == :public)) }
   scope :visible_for, ->(user) {
-    scope = user && !user.sfw_filter? ? self : sfw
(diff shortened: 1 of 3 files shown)


trmcnvn:
Timeout is still an issue, also a new ES error:

`"reason":"No mapping found for [status] in order to sort on"`

NuckChorris:
I dunno if that's easily fixable, but I'll take a look. Dunno why it keeps giving timeouts, that's bizarre

trmcnvn:
Scratch that, the other error isn't a timeout but:

```
"exception": "undefined method `id' for nil:NilClass",
"backtrace": [
"/app/vendor/bundle/ruby/2.3.0/gems/jsonapi-resources-0.9.0/lib/jsonapi/resource_serializer.rb:60:in `block in serialize_to_hash'",
```

NuckChorris:
Wat. Is it giving a nil in the results? :thinking:

PR #132, opened 2017-05-28
Filter out NSFW groups

trmcnvn:
:arrow_up:

### DIFF
--- a/app/policies/group_policy.rb
+++ b/app/policies/group_policy.rb
@@ -33,7 +33,8 @@ class GroupPolicy < ApplicationPolicy
 
   class Scope < Scope
     def resolve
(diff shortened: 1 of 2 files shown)


app/policies/group_policy.rb
```diff
@@ -33,7 +33,8 @@ def group
 
   class Scope < Scope
     def resolve
-      scope.visible_for(user)
+      return scope.visible_for(user) if see_nsfw?
+      scope.sfw
```
NuckChorris: This seems like it would circumvent private groups?

trmcnvn: Yeah, will throw the visible_for check on.

PR #135, opened 2017-05-29
Fixing Mal Sync to use xml importing

toyhammered:
converting initial sync to use mal xml importer.


app/services/list_sync/my_anime_list/library_updater.rb
```diff
@@ -0,0 +1,86 @@
...
+        end
+        # Rewatches/rewatches
+        if library_entry.kind == :anime
+          field_for(:num_watched_times).value = library_entry.reconsume_count
+        elsif library_entry.kind == :manga
+          field_for(:num_read_times).value = library_entry.reconsume_count
+        end
+      end
+
+      def mal_status_key
```
toyhammered, line 61: assuming you took this from what I had (and removed the comments that mapped them nicely...)

app/workers/list_sync/destroy_worker.rb
```diff
@@ -0,0 +1,22 @@
+module ListSync
+  class DestroyWorker
+    include Sidekiq::Worker
+    sidekiq_options retry: 3, queue: 'sync'
+
+    def perform(linked_account_id, media_type, media_id)
+      pending_logs = LibraryEntryLog.for_entry(library_entry).pending
```
toyhammered: where is this `library_entry` variable defined? I only saw it in the `UpdateWorker`. The `for_entry` is only looking for `media_type` and `media_id` only anyways so maybe just pass those in direct and change the method to accept both those parameters?

app/workers/list_sync/sync_worker.rb
```diff
@@ -0,0 +1,25 @@
...
+    sidekiq_options retry: false, queue: 'sync'
+
+    def perform(linked_account_id, user_id)
+      user = User.find(user_id)
+      linked_account = LinkedAccount.find(linked_account_id)
+      entries = user.library_entries
+      entries.map do |le|
+        LibraryEntryLog.create_for(:update, le, linked_account)
+      end
+      pending_logs = LibraryEntryLog.for_entry(library_entry).pending
```
toyhammered: same with this one, where is `library_entry` defined? Also because this one has multiple entries would this actually work?

Possible logical error here?

app/workers/list_sync/update_worker.rb
```diff
@@ -0,0 +1,20 @@
+module ListSync
+  class UpdateWorker
+    include Sidekiq::Worker
+    sidekiq_options retry: 3, queue: 'sync'
+
+    def perform(linked_account_id, library_entry_id)
+      linked_account = LinkedAccount.find(linked_account_id)
+      library_entry = LibraryEntry.find(library_entry_id)
+      pending_logs = LibraryEntryLog.for_entry(library_entry).pending
```
toyhammered, line 9: `library_entry` exists here 👍

toyhammered (commented):
Wow! I am impressed at how this turned out, prob 2x better than what I would have produced tbh. Left a few questions on things I was uncertain about.

app/services/list_sync/my_anime_list/library_updater.rb
NuckChorris: I really don't think we need a comment mapping them nicely... it's pretty self-explanatory :v

app/workers/list_sync/destroy_worker.rb
NuckChorris: Nice catch, lemme fix that!

app/workers/list_sync/sync_worker.rb
NuckChorris: And ditto here

Also yes, it would work fine, it returns an active record scope, not a single row

PR #136, opened 2017-05-30
Categories (Explore Page)

cinquemb:
Will deprecate #122


app/models/category_favorite.rb
```diff
@@ -0,0 +1,43 @@
...
+  belongs_to :user, required: true
+  belongs_to :category, required: true
+
+  validates :user_id, uniqueness: {
+    scope: :category_id,
+    message: 'Cannot fave a category multiple times'
+  }
+
+  after_save do
+    client ||= Stream::CustomEndpointClient.new
```
NuckChorris: `||=` doesn't make much sense here

app/services/trending_service.rb
```diff
@@ -75,6 +77,24 @@ def trending_key(uid = nil)
...
+    key
+  end
+
+  def handle_trending_categories(id, weight = 1.0)
+    return unless %w[Anime Manga Drama].include?(namespace.name)
+    categories = namespace.includes(:categories).find_by(id: id).categories
+    categories.each do |category|
+      key = trending_category_key(category.id)
+      update_score(key, id, weight)
+      trim(key, limit: NETWORK_LIMIT) if rand < TRIM_PROBABILITY
```
NuckChorris: Should use a different constant from `NETWORK_LIMIT`, maybe `CATEGORY_LIMIT`?

lib/anidb/category_importer.rb
```diff
@@ -0,0 +1,45 @@
+class AnidbCategoryImport
+  class ImportFile
```
NuckChorris: I think we want to keep the class names matching the filename, so `anidb/category_importer.rb` would be `AniDB::CategoryImporter` and `AnidbCategoryImport::ImportFile` would be `anidb_category_import/import_file.rb` — though honestly you can just keep `ImportFile` inline, but probably want to rename the outermost class here to match the file.

lib/anidb/media_category_assoc_importer.rb
```diff
@@ -0,0 +1,94 @@
+class AnidbAssocMediaCategoryImport
```
NuckChorris: A namespace might be good? 🤔  Maybe `AnidbCategoryImport` (`lib/anidb_category_import/media_import.rb`, etc.)?

app/services/trending_service.rb
cinquemb: Make sense to me, although i'm sure what would be appropriate to set it to. Should we keep it at 20?

lib/anidb/media_category_assoc_importer.rb
cinquemb: Sounds good to me

lib/anidb/category_importer.rb
cinquemb: Ok, i'll change the outer one.

trmcnvn (approved):
:+1:

PR #141, opened 2017-06-02
adding mapping of anidb to anime from dump

cinquemb:
We will need to run the media_assoc rake task again


vevix:
Just want to make a note that Stream is returning a 500 for this endpoint currently.

NuckChorris (approved):

PR #142, opened 2017-06-02
adding "realtime" endpoint

cinquemb:


trmcnvn:
Just want to make a note that Stream is returning a 500 for this endpoint currently.

cinquemb:
According to @NuckChorris, the real-time query wont work outside of production, and i can confirm that when running ```Stream::CustomEndpointClient.new.get('realtime/5554/')``` works when running it in the local console it will fail

app/services/recommendations_service/media.rb
```diff
@@ -25,6 +25,18 @@ def recommendations_for(klass)
       reccs.map(&:first)
     end
 
+    def realtime_recommendations_for(klass)
+      res = client.get("realtime/#{user.id}/")
+      reccs = res.body.dig('results', klass.name)
+      reccs = reccs.reduce(&:merge)
+      media_ids = reccs.keys.map { |key| key.split(':').last.to_i }
+      media = klass.find(media_ids).index_by { |m| "#{m.class.name}:#{m.id}" }
```
NuckChorris: Would probably be good to split this out into an `enrich` method, so that the logic for swapping `"Anime:321"` for actual `Anime` instances can be centralized

cinquemb: Makes sense, I'll do this

config/routes.rb
```diff
@@ -73,6 +73,10 @@
       # Trending
       get '/trending/:namespace', to: 'trending#index'
       get '/recommendations/:namespace', to: 'recommendations#index'
+      get '/recommendations/:namespace/realtime', to: 'recommendations#realtime'
+
+      get '/category_recommendations/:namespace', to: 'category_recommendations#index'
+      get '/cateogry_recommendations/:namespace/realtime', to: 'category_recommendations#realtime'
```
trmcnvn: Typo here. `cateogry` -> `category`

app/services/recommendations_service/media.rb
```diff
@@ -16,6 +16,66 @@ def poll
...
+        ).find_by(id: key.to_i)
+        next unless category_object
+        j_category_object = category_object.as_json(include: field_key)
+        j_category_object[field_name].sort_by do |e|
+          media_ids.index(e[:id]) || media_ids.length
+        end
+        categories_objects << j_category_object
+      end
+      categories_objects
+    end
```
NuckChorris: Could you give an example of the input and output of this method?  I'm not sure I quite follow what's going on here

app/controllers/category_recommendations_controller.rb
```diff
@@ -0,0 +1,58 @@
...
+
+  def namespace_class
+    @namespace_class ||= namespace.safe_constantize
+  end
+
+  def resource_class
+    "#{namespace}Resource".safe_constantize
+  end
+
+  def authenticate_user
```
NuckChorris: This might make sense in `CustomControllersHelper`, and maybe named `authenticate_user!` to match Devise (since the method has the same name)?

app/services/recommendations_service/media.rb
cinquemb: The input: https://kitsu.slack.com/files/kev/F5R490W8N/-.py
The output is the json serialization of  of a category openstructs, with the anime sorted

app/controllers/category_recommendations_controller.rb
cinquemb: Makes sense to me

app/services/recommendations_service/media.rb
NuckChorris: I'm afraid I don't quite follow, could you give an actual example of the JSON that's serialized by this?  I'm sure there's gotta be an easier/more clear way to do this

cinquemb: Here's what it outputs: https://files.slack.com/files-pri/T27CM6PGW-F5T06EBA8/sample_results_recommendations.js

app/controllers/category_recommendations_controller.rb
```diff
@@ -0,0 +1,62 @@
+class CategoryRecommendationsController < ApplicationController
+  include CustomControllerHelpers
+
+  before_action :authenticate_user!
+  before_action :validate_namespace
+
+  def index
+    serializer = CategoryRecommendationSerializer.new(
```
NuckChorris, line 8: I don't see this class defined, do you mean `JSONAPI::ResourceSerializer`?

app/models/category_recommendation.rb
```diff
@@ -0,0 +1,8 @@
+class CategoryRecommendation
+  attr_accessor :category, :media, :id
+  def initialize(params)
```
NuckChorris: It's optional but if you want to be a little more explicit, you can do this:

```ruby
def initialize(category:, media:, id:)
  @category = category
  @media = media
  @id = id
end
```

Also it looks like `id` is always nil, so maybe you could default it to `category.id`, just for the sake of having an ID? 🤔

app/controllers/category_recommendations_controller.rb
cinquemb: Ahh yeah, forgot to check in the new file I created

NuckChorris (approved):

PR #143, opened 2017-06-02
Filter out NSFW groups from ES

trmcnvn:
:arrow_up: :no_entry_sign:


trmcnvn:
This will need to be merged prior to next production push.

/cc @hummingbird-me/staff

NuckChorris (approved):

PR #146, opened 2017-06-04
Add since filter to media

matthewdias:
Not sure if this needs to query anime/manga tables. Cant test from phone.


NuckChorris:
Just remove the `media.` and it'll work

PR #147, opened 2017-06-06
Media reactions

cinquemb:


app/models/library_entry.rb
```diff
@@ -48,6 +48,7 @@ class LibraryEntry < ApplicationRecord
   belongs_to :drama
   has_one :review, dependent: :destroy
   has_many :library_events, dependent: :destroy
+  has_many :media_reactions
```
trmcnvn: Shouldn't this be `has_one`?

cinquemb: I was unsure, since `has_one` from the docs says "This association indicates that each instance of a model contains or possesses one instance of another model.", which isn't true because library entries don't have to have any media reactions. 

Compared to `has_many` which implies: "This association indicates that each instance of the model has zero or more instances of another model." which I thought would be more permissive.

NuckChorris: No associations are required unless `required: true`. Only belongs_to defaults to required in Rails 5. This is probably better served as a has_one since only one is allowed

NuckChorris: Also should this be `dependent: :destroy`?

app/models/media_reaction.rb
```diff
@@ -0,0 +1,79 @@
...
+#  progress         :integer          default(0), not null
+#  reaction         :string(140)
+#  up_votes_count   :integer          default(0), not null
+#  created_at       :datetime         not null
+#  updated_at       :datetime         not null
+#  anime_id         :integer          indexed
+#  drama_id         :integer          indexed
+#  library_entry_id :integer          indexed
+#  manga_id         :integer          indexed
+#  media_id         :integer          not null, indexed => [media_type, user_id]
```
NuckChorris, line 16: Don't need media_id/type if it's got the three columns for types

spec/factories/media_reactions.rb
```diff
@@ -0,0 +1,47 @@
...
+#  fk_rails_9a5bef4caf  (manga_id => manga.id)
+#  fk_rails_bbc29d526d  (library_entry_id => library_entries.id)
+#  fk_rails_db814b132f  (anime_id => anime.id)
+#
+
+FactoryGirl.define do
+  factory :media_reaction do
+    reaction { Faker::Lorem.characters(140) }
+    association :anime, factory: :anime, strategy: :build
+    association :library_entry, factory: :library_entry, rating: 3.0,
```
NuckChorris: Probably don't need a `rating` here, I think the factory will still work without it

app/models/media_reaction.rb
cinquemb: It's easier to do validation for unique on the db level using media_id/type

NuckChorris: That's fair

app/models/library_entry.rb
```diff
@@ -48,7 +51,7 @@ class LibraryEntry < ApplicationRecord
   belongs_to :drama
   has_one :review, dependent: :destroy
   has_many :library_events, dependent: :destroy
-  has_many :media_reactions
+  has_one :media_reactions, dependent: :destroy
```
NuckChorris: Shouldn't that be `has_one :media_reaction`, with the singular form?

app/policies/media_reaction_policy.rb
```diff
@@ -0,0 +1,10 @@
+class MediaReactionPolicy < ApplicationPolicy
+  def create?
+    record.user == user
```
NuckChorris, line 3: Probably needs to have a separate `create?` from `destroy?` so that admins can delete reported reviews

cinquemb: Ahh yeah, that makes sense.

NuckChorris (approved):

app/policies/media_reaction_policy.rb
```diff
@@ -0,0 +1,13 @@
...
+  end
+
+  def update?
+    false
+  end
+
+  def destroy?
+    record.try(:user) == user || is_admin?
```
toyhammered: `record&.user` should work right?

app/models/media_reaction.rb
```diff
@@ -0,0 +1,79 @@
+# == Schema Information
+#
+# Table name: media_reactions
+#
+#  id               :integer          not null, primary key
+#  media_type       :string           not null, indexed => [media_id, user_id]
+#  progress         :integer          default(0), not null
+#  reaction         :string(140)
```
toyhammered, line 8: for what we were talking about in slack, we could technically have a `reaction_type` enum with `%i[individual, whole]` or something, but this idea would need to be approved by @JoshFabian

toyhammered (commented):
Looks good! Just left 2 quick questions.

NuckChorris:
Looks like one of the specs is failing, any idea what's happening there?

matthewdias:
whoops forgot to update the policy spec

PR #148, opened 2017-06-06
Media attributes

cinquemb:
Will need to run ```bundle exec rake importers:media_attributes:seed``` in order to seed the initial attributes


app/models/anime.rb
```diff
@@ -59,6 +59,7 @@ class Anime < ApplicationRecord
   has_many :anime_productions, dependent: :destroy
   has_many :anime_characters, dependent: :destroy
   has_many :anime_staff, dependent: :destroy
+  has_many :media_attribute, through: :anime_media_attributes
```
NuckChorris, line 62: Don't you need to also declare `has_many :anime_media_attributes` for this to work?  Also do we even need an association to `media_attribute`, it's just a table with like... one row?

app/models/media_attribute.rb
```diff
@@ -0,0 +1,34 @@
...
+# Indexes
+#
+#  index_media_attribute_on_slug   (slug)
+#  index_media_attribute_on_title  (title)
+#
+
+class MediaAttribute < ActiveRecord::Base
+  self.table_name = 'media_attribute'
+  extend FriendlyId
+  friendly_id :title, use: %i[slugged finders history]
```
NuckChorris: Is there a reason for this to have a slug?  Is there any place where we put the media attribute in a user-visible URL?

app/models/media_attribute.rb
```diff
@@ -0,0 +1,34 @@
...
+#  updated_at    :datetime         not null
+#
+# Indexes
+#
+#  index_media_attribute_on_slug   (slug)
+#  index_media_attribute_on_title  (title)
+#
+
+class MediaAttribute < ActiveRecord::Base
+  self.table_name = 'media_attribute'
```
NuckChorris: Is there a reason you've overridden the default?

app/models/media_attribute_vote.rb
```diff
@@ -0,0 +1,70 @@
...
+#  index_media_attribute_votes_on_user_id  (user_id)
+#  index_user_media_on_media_attr_votes    (user_id,media_id,media_type) UNIQUE
+#
+# Foreign Keys
+#
+#  fk_rails_39b0c09be9  (user_id => users.id)
+#
+
+class MediaAttributeVote < ActiveRecord::Base
+  enum vote: %i[unvoted low neutral high]
```
NuckChorris: Is `unvoted` even an option, I thought it was only low/neutral/high, neutral being the "unvoted"?

db/migrate/20170602065925_add_join_tables_for_media_attribute.rb
```diff
@@ -0,0 +1,46 @@
+class AddJoinTablesForMediaAttribute < ActiveRecord::Migration
+  def change
+    create_table :anime_media_attributes do |t|
+      t.references :anime, null: false, index: true,
+                           foreign_key: true, required: true
+      ActiveRecord::Base.pluralize_table_names = false
```
NuckChorris: Just let Rails do its thing, `media_attributes` is an okay table name

app/models/anime.rb
cinquemb: Yes to the first question.
And it has at least three now, and maybe more in the future.

app/models/media_attribute.rb
cinquemb: rails complained about something when i used the resource generator with `MediaAttributes`, and somethings didn't work so to get the errors to go away I overridden the default

app/models/media_attribute.rb
cinquemb: I don't know if they're will be, but i figured if someone wanted to eventually do that, they could?

app/models/media_attribute_vote.rb
cinquemb: Here was the convo around unvoted: https://kitsu.slack.com/archives/C2C1D7HT3/p1496385038886626

app/models/anime_media_attribute.rb
```diff
@@ -0,0 +1,33 @@
...
+#
+# Foreign Keys
+#
+#  fk_rails_88955ab592  (media_attribute_id => media_attributes.id)
+#  fk_rails_c8ff37f0ab  (anime_id => anime.id)
+#
+# rubocop:enable Metrics/LineLength
+
+class AnimeMediaAttribute < ActiveRecord::Base
+  self.table_name = 'anime_media_attributes'
```
NuckChorris, line 29: Lemme guess... it tries to do `anime_medium_attributes`?  If we haven't yet, we should mark `media` as uncountable in Rails' inflector, then this table should default correctly

NuckChorris (approved):

PR #149, opened 2017-06-08
Notification Settings

cinquemb:


app/models/notification_setting_state.rb
```diff
@@ -0,0 +1,29 @@
...
+#                                        (notification_setting_id)
+#  index_notification_setting_states_on_user_id                  (user_id)
+#
+# Foreign Keys
+#
+#  fk_rails_057c83c713  (user_id => users.id)
+#  fk_rails_66e9ee5396  (notification_setting_id => notification_settings.id)
+#
+
+class NotificationSettingState < ActiveRecord::Base
```
NuckChorris: Why not use an ActiveRecord `enum` column for this instead of a separate `notification_setting_id` + table?  Also perhaps it'd be good to have multiple `toggled` columns for the different methods of notification?

app/models/user.rb
```diff
@@ -47,7 +47,8 @@
 #  name                        :string(255)
 #  ninja_banned                :boolean          default(FALSE)
 #  password_digest             :string(255)      default(""), not null
-#  past_names                  :string           default([]), not null, is an Array
+#  past_names                  :string           default([]), not null,
+#                                                          is an Array
```
NuckChorris: Try to avoid manual changes to this section, it's maintained by the `annotate` gem

app/models/user.rb
```diff
@@ -300,6 +301,16 @@ def update_profile_completed!
     update_profile_completed.save!
   end
 
+  def setup_notification_setting_states!
```
NuckChorris: I think it would make more sense to have the logic primarily in the `NotificationSetting` model

db/schema.rb
```diff
@@ -93,6 +93,20 @@
...
+    t.integer  "neutral_vote_count", default: 0, null: false
+    t.integer  "low_vote_count",     default: 0, null: false
+    t.datetime "created_at",                     null: false
+    t.datetime "updated_at",                     null: false
+  end
+
+  add_index "anime_media_attributes", ["anime_id", "media_attribute_id"], name: "index_anime_media_attribute", unique: true, using: :btree
+  add_index "anime_media_attributes", ["anime_id"], name: "index_anime_media_attributes_on_anime_id", using: :btree
+  add_index "anime_media_attributes", ["media_attribute_id"], name: "index_anime_media_attributes_on_media_attribute_id", using: :btree
+
```
NuckChorris: I think some unrelated schema stuff leaked into here, you might wanna use `git add -p` to pick and choose lines

lib/notification_import/seed_notification_types.rb
```diff
@@ -0,0 +1,21 @@
...
+      end
+    end
+
+    def run!
+      ActiveRecord::Base.logger = Logger.new(nil)
+      Chewy.strategy(:bypass)
+      create_notification_types!
+    end
+  end
+end
```
NuckChorris: If you use an `enum` this can go.  Otherwise we have the `seed` directory for base data like this

cinquemb: The reason I used this is so that if we decided to add more attributes in the future we can keep in the database and allow for what kinds of media we wanted to allow this for.

app/models/notification_setting_state.rb
cinquemb: I decided to do it this way to make it easier to add more setting types in the future instead of having to add more columns to db.

db/schema.rb
cinquemb: yeah, i'll need to rebuild this schema.

app/models/notification_setting_state.rb
NuckChorris: I think this actually makes things a little more complicated... we need to write a script to add it to the DB and stuff instead of just changing the code, which would already need to be changed to add support for more notification methods

cinquemb: You don't need a script, you can do it from the rails admin.

lib/notification_import/seed_notification_types.rb
cinquemb: Ok, i'll move this seed dir.

app/models/notification_setting_state.rb
NuckChorris: And then deploys are messy and manual.  If we can't add new notification methods without adding code, it's just more trouble to have them editable in the database.

cinquemb: So rather than just needing to make changes to the code, you want to have to make changes to the code and the database for new notifications? If that's how you want it to be done, sure I can do it.

NuckChorris: Either way will require changes to both code and database, but using an enum for the types and a set of `mobile`, `web`, `email` columns would be more explicit and allow us to better version the changes and make the migrations smoother (and more self-contained)

cinquemb: I can see that, while sacrificing flexibility because I'm inclined to think that the notifications settings we have so far will become way more exhaustive/granular.

NuckChorris: Indeed, I think we'll want to be constantly modifying this, and it's much easier to add more values to an `enum` as part of a deploy than to remember to manually edit the database (and keep that synced to staging and dev environments!)

cinquemb: Ok that's fair. So I'll just need to merge these two tables together by using enums for what notifications we have defined so far?

app/models/notification_setting.rb
```diff
@@ -3,15 +3,37 @@
...
 #
-#  id           :integer          not null, primary key
-#  setting_name :string           not null
-#  created_at   :datetime         not null
-#  updated_at   :datetime         not null
+#  id                      :integer          not null, primary key
+#  is_email_toggled        :boolean          default(TRUE)
+#  is_fb_messenger_toggled :boolean          default(TRUE)
+#  is_mobile_toggled       :boolean          default(TRUE)
+#  is_web_toggled          :boolean          default(TRUE)
```
NuckChorris: prefixing these predicates with `is` is out-of-place in Ruby, generally we'd do `web_enabled`, `mobile_enabled`, etc. — and then Rails automatically gives us `web_enabled?` methods on the model

app/models/notification_setting.rb
```diff
@@ -3,15 +3,37 @@
...
+  enum setting_type: NOTIFICATION_TYPES
+  belongs_to :user
 
-  validates :setting_name, presence: true
+  def self.setup_notification_settings(user)
+    NOTIFICATION_TYPES.each do |st|
+      NotificationSetting.where(
+        setting_type: st,
+        user: user
+      ).first_or_create
```
NuckChorris: What should the defaults be for these notification channels? /cc @JoshFabian

app/models/notification_setting.rb
```diff
@@ -0,0 +1,42 @@
...
+#  fk_rails_0c95e91db7  (user_id => users.id)
+#
+# rubocop:enable Metrics/LineLength
+
+class NotificationSetting < ActiveRecord::Base
+  NOTIFICATION_TYPES = %i[mentions replies likes follows posts].freeze
+  enum setting_type: NOTIFICATION_TYPES
+  belongs_to :user
+
+  def self.setup_notification_settings(user)
```
NuckChorris, line 31: Could we get a rake task or migration to add these defaults for existing users?

db/migrate/20170616053125_add_notification_settings_for_users.rb
```diff
@@ -0,0 +1,18 @@
...
+        all_users_notifications << {
+          setting_type: st,
+          user: user,
+          fb_messenger_enabled: false,
+          mobile_enabled: false,
+          email_enabled: false
+        }
+      end
+    end
+    NotificationSetting.create(all_users_notifications)
```
NuckChorris: This is probably a huge RAM hog, since it's gonna create 800,000-1,000,000 hashes in RAM.  Probably better to do smaller batches, maybe move this into the outer loop so it flushes after each user

PR #151, opened 2017-06-10
Add status to media index

matthewdias:
idk if this works


NuckChorris:
Seems like it would work for search yeah, though you probably need to add something to the resource

trmcnvn:
Resource level should be good, it's added as an attribute already.

We'll need to update the media index on production.

trmcnvn (approved):

PR #152, opened 2017-06-11
Fix user deletion issue.

trmcnvn:
- Reparent the personal posts of the user to the `Deleted` user as well.

---

Another approach would be to really delete the records (We can just call destroy a 2nd time for that to happen).


NuckChorris:
I think it's better to actually delete them for real 🤔

NuckChorris:
Lovely!

NuckChorris (approved):

PR #163, opened 2017-06-29
Badges

NuckChorris:


wopian:
Missing rank 3 ;)

https://github.com/hummingbird-me/hummingbird-server/blob/f8155c851f5202088120584df8af9f526a9eeb6d/app/badges/feed_commenting_badge.rb#L5

NuckChorris:
Ah nice catch

app/services/badge/bestowment_service.rb
```diff
@@ -0,0 +1,55 @@
...
+      @badge = badge
+      @user = user
+    end
+
+    # Actually creates/updates Bestowments for the badge
+    def run!
+      tracked_ranks.each do |(rank, progress)|
+        bestowment = bestowment_for(rank).first_or_initialize
+        # Ratcheting - progress can only increase, never decrease.
+        bestowment.progress = [bestowment.progres, progress].max
```
cinquemb, line 18: So If someone deletes  their comment,post or likes and etc, they will maintain their progress for a badge?

app/models/user.rb
```diff
@@ -106,10 +106,13 @@ class User < ApplicationRecord
   ].freeze
 
   enum rating_system: %i[simple advanced regular]
+  enum theme: %i[light dark]
   rolify after_add: :update_title, after_remove: :update_title
   has_secure_password
   update_index('users#user') { self }
-  enum theme: %i[light dark]
+  update_bestowment_for 'ConsecutiveDaysBadge', if: :consecutive_days_changed?
+  update_bestowment_for 'TheAdventureBeginsBadge'
```
cinquemb, line 114: Will this default to after create? I can see an issue happening if it comes before the user object has been committed.

app/services/badge/bestowment_service.rb
NuckChorris: Yeah, so it basically works like this:

```
like - 1
like - 2
like - 3
unlike - 3
like - 3
like - 4
```

So it never decreases, but it also won't increase until it reaches that point again.

This is mainly to prevent the progress from dropping on badges that have already earned, but I think it'd also annoy users to see their badge progress dropping 😛 

I could probably do this differently (just prevent earned badges from being modified) if @JoshFabian thinks a different logic would be better

app/models/user.rb
NuckChorris: `update_bestowment_for` passes `on:` and `if:` directly to `after_commit`

app/services/badge/bestowment_service.rb
cinquemb: Ahh i was just thinking about from an abuse standpoint, some one could just call the api with a valid token and create as many actions in order to get some badge(s), but if thats not a something we're worried about now then cool.

app/services/badge/bestowment_service.rb
```diff
@@ -0,0 +1,55 @@
+# Handles creating and updating bestowments for a (badge, user) pair.
+class Badge
+  class BestowmentService
+    attr_reader :badge, :user
+
+    # @param [Class<Badge>] the badge to handle bestowments for
+    # @param [User] the user to bestow this badge upon
+    def initialize(badge, user)
+      @badge = badge
```
cinquemb, line 9: Does ```@badge``` need to be initialized? 
I'm not able to get anything to return when ```badge.progress_for(user)``` is called from ```ranks```

app/models/user.rb
```diff
@@ -106,10 +106,13 @@ class User < ApplicationRecord
...
+  update_bestowment_for 'StaffBadge', if: :title_changed?
```
cinquemb, line 115: I can't seem to get a worker to fire for the staff badge when changing a users title from ```user.title = ''``` to ```user.title = 'Staff'``` after calling ```user.save```

app/services/badge/bestowment_service.rb
NuckChorris: Wow I just noticed the typo on this line `bestowment.progres` (missing an S)

cinquemb: Ahh yeah, but I haven't even been able to return anything from `tracked_ranks` so I haven't ran into this issue yet.

PR #164, opened 2017-06-29
add missing timestamps

matthewdias:
- add missing timestamps
- expose timestamps on all resources
- fix model annotation wrapper


NuckChorris (approved):
Doing the lord's work right here 🙌

PR #165, opened 2017-07-03
Add tba to media and remove started_airing_date_known

matthewdias:


NuckChorris:
Is there a reason `tba` would be a string? 🤔

matthewdias:
The idea is to have a value where we can give the specifics of when it has been announced. eg. "Spring 2019", or "August 2020"

NuckChorris (approved):

PR #169, opened 2017-07-07
AMA

cinquemb:


app/models/ama.rb
```diff
@@ -0,0 +1,37 @@
...
+#
+#  fk_rails_be5e31a286  (author_id => users.id)
+#  fk_rails_ef2f48b4b8  (original_post_id => posts.id)
+#
+# rubocop:enable Metrics/LineLength
+
+class Ama < ApplicationRecord
+  belongs_to :author, required: true, class_name: 'User'
+  belongs_to :original_post, required: true, class_name: 'Post'
+  has_many :posts, dependent: :destroy
```
NuckChorris: Would an AMA have multiple posts? 🤔

app/models/ama.rb
```diff
@@ -0,0 +1,37 @@
...
+#  index_amas_on_original_post_id  (original_post_id)
+#
+# Foreign Keys
+#
+#  fk_rails_be5e31a286  (author_id => users.id)
+#  fk_rails_ef2f48b4b8  (original_post_id => posts.id)
+#
+# rubocop:enable Metrics/LineLength
+
+class Ama < ApplicationRecord
```
NuckChorris: You might be able to get classnames like `AMA` (all uppercase) if you configure the inflector in `config/initializers/inflections.rb` to be aware that it's an acronym (we already have a few acronyms in there)

app/models/ama_subscriber.rb
```diff
@@ -0,0 +1,44 @@
...
+  belongs_to :user, required: true
+  belongs_to :ama, required: true, counter_cache: true
+
+  validates :ama_id, uniqueness: { scope: :user_id }
+
+  after_commit(on: :create) do
+    ama.posts.each do |post|
+      user.notifications.follow(post.comments_feed, scrollback: 0)
+    end
+    user.notifications.follow(ama.original_post.comments_feed, scrollback: 0)
```
NuckChorris: Won't this notify the user for every single comment made on an AMA they're subscribed to?  Seems like that might get a little annoying for major AMAs

app/models/comment.rb
```diff
@@ -84,12 +85,24 @@ def no_grandparents
...
+    return unless post.ama
+    ama = post.ama
+    return if ama.author == user
+    now_time = Time.now
+
+    unless ama.start_date <= now_time && ama.end_date >= now_time
+      errors.add(:post, 'cannot make any more comments on this ama')
+    end
+  end
+
```
NuckChorris: Perhaps it would make more sense to model this whole thing outside of the Post/Comment structure (since you have the AMA model already), so that it can have more logic specifically geared towards that?  At the very least, this should go on the AMA, so you can just ask `ama.closed?`

app/models/comment_like.rb
```diff
@@ -29,6 +29,19 @@ class CommentLike < ApplicationRecord
...
+    return unless comment.post
+    return unless comment.post.ama
+    ama = comment.post.ama
+    return if ama.author == user
+    now_time = Time.now
+
+    unless ama.start_date <= now_time && ama.end_date >= now_time
+      errors.add(:comment, 'cannot like comments on this ama')
+    end
+  end
```
NuckChorris: (And then the logic could be reused here too)

app/models/ama.rb
cinquemb: Ok, ill try this

app/models/ama.rb
cinquemb: I interpreted "AMA Questions + Replies, these are standard posts from the community and replies from the AMA author." from the specs to mean that they AMA's could have multiple posts on top of the original author post

app/models/comment.rb
cinquemb: We still want to retain the functionality that posts/comments have for ama, and I'm not sure if we want a separate feeds for posts/comments that are under amas, but yeah it makes sense to condense this to something like `ama.closed?`

app/models/ama_subscriber.rb
cinquemb: Yeah I can see that, @JoshFabian, what do you think should happen when people "subscribe" to an AMA?

app/models/ama.rb
NuckChorris: Hmm, it seems like he would've meant comments 🤔

cinquemb: Yeah, idk. @JoshFabian, do you mean posts and comments, or just comments on the original author post?

app/models/comment.rb
NuckChorris: Feeds can have a lot more than just posts/comments, we can send anything to a feed, just by mixing in `WithActivity` into the model.  I guess you'd also have to duplicate PostLikes/CommentLikes though, so it might get a little excessive to split it into its own model

app/models/ama.rb
JoshFabian: Yeah, I think they're essentially comments on a post. Sort of like this: https://staging.kitsu.io/posts/8690875

app/models/ama_subscriber.rb
JoshFabian: I think we just want to be able to send them notifications before the AMA starts.

cinquemb: Then it seems like we'll need a background worker checking for ama's that are "about to start" ( == 1 hour before? ok) and then send out notifications to all the subscribers and not have them follow the original post comments feed.

NuckChorris: https://github.com/mperham/sidekiq/wiki/Scheduled-Jobs you can schedule a job to execute at that time with Sidekiq

cinquemb: Yeah that would make more sense vs checking in the background since we know the times.

app/models/ama.rb
```diff
@@ -0,0 +1,61 @@
...
+# rubocop:enable Metrics/LineLength
+
+class AMA < ApplicationRecord
+  belongs_to :author, required: true, class_name: 'User'
+  belongs_to :original_post, required: true, class_name: 'Post'
+  has_many :ama_subscribers, dependent: :destroy
+
+  scope :for_original_post, ->(post) { where(original_post: post) }
+
+  def send_ama_notifciation
```
NuckChorris: *notification

app/models/ama.rb
```diff
@@ -0,0 +1,61 @@
...
+#
+#  fk_rails_be5e31a286  (author_id => users.id)
+#  fk_rails_ef2f48b4b8  (original_post_id => posts.id)
+#
+# rubocop:enable Metrics/LineLength
+
+class AMA < ApplicationRecord
+  belongs_to :author, required: true, class_name: 'User'
+  belongs_to :original_post, required: true, class_name: 'Post'
+  has_many :ama_subscribers, dependent: :destroy
```
NuckChorris, line 31: I generally like to drop the prefix, so that we don't end up with `AMA.ama_subscribers`

app/models/ama.rb
```diff
@@ -0,0 +1,61 @@
...
+  def send_ama_notifciation
+    subscriber_notifications = ama_subscribers.map(&:user).map(&:notifications)
+    original_post.feed.activities.new(
+      target: original_post,
+      actor: author,
+      object: self,
+      foreign_id: self,
+      verb: self.class.name.underscore,
+      time: Time.now,
+      to: subscriber_notifications
```
NuckChorris: This won't work past 100 users

app/models/ama.rb
```diff
@@ -0,0 +1,61 @@
...
+    )
+  end
+
+  def open?
+    now_time = Time.now
+    start_date <= now_time && end_date >= now_time
```
NuckChorris: `start_date..end_date.include?(time.now)` might be a better expression of this

app/models/ama.rb
```diff
@@ -0,0 +1,61 @@
...
+  end
+
+  before_validation do
+    self.end_date = start_date + 1.hour
```
NuckChorris: Should this be `||=` so we can manually make longer ones?

app/models/comment_like.rb
```diff
@@ -29,6 +29,11 @@ class CommentLike < ApplicationRecord
   belongs_to :comment, required: true, counter_cache: :likes_count, touch: true
 
   validates :comment, uniqueness: { scope: :user_id }
+  validates :comment, active_ama: {
```
NuckChorris, line 32: This is a bit odd to read, I'm not sure what the intent is?

app/policies/ama_policy.rb
```diff
@@ -0,0 +1,11 @@
+class AMAPolicy < ApplicationPolicy
+  def update?
+    record.author == user
+  end
+
+  def destroy?
+    record.author == user || is_admin?
+  end
+
+  alias_method :create?, :update?
```
NuckChorris: I think we want staff to be the only creators

app/workers/ama_starting_worker.rb
```diff
@@ -0,0 +1,7 @@
+class AMAStartingWorker
+  include Sidekiq::Worker
+
+  def perform(ama)
+    ama.send_ama_notifciation
```
NuckChorris: Would this be better as a service instead? :thinking:

NuckChorris (changes requested):

app/models/ama.rb
cinquemb: so if in the future we have subscribers for another model, we'll have to change this to something else? If there were an `ama_follow` model, you would want that to be just `follow`?

app/models/ama.rb
cinquemb: Is this a stream limitation? What would you suggest? It didn't seem too different than from `WithMedia`, except for not doing every commit.

app/models/comment_like.rb
cinquemb: The intent is to check if this is happening under an active ama

app/policies/ama_policy.rb
cinquemb: Is this true @JoshFabian, if so ill just check `is_admin?`

app/workers/ama_starting_worker.rb
cinquemb: Idk, you tell me lol what are the pros/cons?

app/policies/ama_policy.rb
JoshFabian: @cinquemb @NuckChorris Uh... this is a good question. For now I think that yes, only an admin can create an AMA, but they'll be doing it as a registered user.

For example, if we're doing an AMA for George Clooney, an admin will create the AMA for George, including any introduction text for the post body that George provides. George will then be able to respond to questions in the AMA using his account without issue.

Does that make sense?

app/models/ama.rb
```diff
@@ -46,12 +46,12 @@ def send_ama_notifciation
...
   def open?
-    now_time = Time.now
-    start_date <= now_time && end_date >= now_time
+    (start_date..end_date).cover?(Time.now)
   end
 
   before_validation do
-    self.end_date = start_date + 1.hour
+    self.end_date = start_date + 1.hour if start_date.to_i >= end_date.to_i
+    self.end_date ||= start_date + 1.hour
```
NuckChorris: It might be best to tweak this to one conditional, so like

```ruby
if end_date.blank? || start_date >= end_date
  self.end_date = start_date + 1.hour
end
```

Also you shouldn't need those `to_i` calls — Ruby has fantastic operator overloading, so you should be able to compare Dates directly

spec/models/post_follow_spec.rb
```diff
@@ -35,6 +35,7 @@
       post = create(:post)
       ama = create(:ama, original_post: post)
       ama.start_date = 6.hours.ago
+      ama.end_date = ama.start_date + 1.hour
```
NuckChorris: These shouldn't be necessary because of the default, right?

NuckChorris:
Looks like there's conflicts, could you rebase?

app/models/ama.rb
NuckChorris: Nah, remember these names are scoped just to this model, the class name is separate.  So we can have `subscribers` on an AMA results in AMA Subscribers, which just makes a lot of sense!

app/models/ama.rb
NuckChorris: Yeah, Stream caps the `to` field at 100.  We'll need a feed group for AMAs, and then make user notification feeds follow that when they subscribe

app/models/post_like.rb
```diff
@@ -22,6 +22,11 @@ class PostLike < ApplicationRecord
   belongs_to :post, required: true, counter_cache: true, touch: true
 
   validates :post, uniqueness: { scope: :user_id }
+  validates :post, active_ama: {
+    message: 'cannot like this AMA',
+    post: true,
+    user: :user
+  }
```
NuckChorris: @JoshFabian Should we allow liking an AMA after it's closed?

app/validators/active_ama_validator.rb
```diff
@@ -0,0 +1,15 @@
+class ActiveAMAValidator < ActiveModel::EachValidator
+  def validate_each(record, attribute, value)
+    ama = if options.key?(:comment)
```
NuckChorris: This logic might better be expressed as `value.respond_to?(:post) ? value.post : value` or something else along those lines, so that we don't just have flags controlling it?  At the very least I think a `for: :comment` option might be more clear than `comment: true`.  A good rule of thumb I've found is to always second-guess boolean options — would an enum be better?

app/models/ama.rb
```diff
@@ -0,0 +1,62 @@
...
+
+  def send_ama_notification
+    feed.activities.new(
+      target: original_post,
+      actor: author,
+      object: self,
+      foreign_id: self,
+      verb: self.class.name.underscore,
+      time: Time.now
+    )
```
NuckChorris: You need to persist the activity with `feed.activities.new().create`

app/models/ama_subscriber.rb
```diff
@@ -0,0 +1,42 @@
...
+  belongs_to :user, required: true
+  belongs_to :ama, required: true, counter_cache: true
+
+  validates :ama_id, uniqueness: { scope: :user_id }
+  validates :ama, active_ama: {
+    message: 'can not subscribe to this AMA',
+    user: :user
+  }
+
+  after_create do
```
NuckChorris: These should probably be in `after_commit on: :create` and `after_commit on: :destroy` hooks instead, so that they don't run if there's an error and the db transaction is rolled back

cinquemb: yeah good idea

app/controllers/ama_controller.rb
```diff
@@ -0,0 +1,2 @@
+class AmaController < ApplicationController
```
NuckChorris: Should this file actually be `AmasController` due to Rails' plurality rules?

NuckChorris (approved):

app/models/ama.rb
```diff
@@ -0,0 +1,62 @@
+# rubocop:disable Metrics/LineLength
+# == Schema Information
+#
+# Table name: amas
+#
+#  id                    :integer          not null, primary key
+#  ama_subscribers_count :integer          default(0), not null
+#  description           :string           not null
+#  end_date              :datetime         not null
+#  start_date            :datetime         not null
```
matthewdias, line 10: Shouldn't this be nullable? As I understand we'll want to be able to create AMAs without dates and original posts, so that they can be added later when AMAs are scheduled.
```
- Upcoming AMAs (Shown as Current Month, Next Month, The Following Month and More Upcoming)
```

app/models/ama.rb
```diff
@@ -0,0 +1,62 @@
...
+# Foreign Keys
+#
+#  fk_rails_be5e31a286  (author_id => users.id)
+#  fk_rails_ef2f48b4b8  (original_post_id => posts.id)
+#
+# rubocop:enable Metrics/LineLength
+
+class AMA < ApplicationRecord
+  belongs_to :author, required: true, class_name: 'User'
+  belongs_to :original_post, required: true, class_name: 'Post'
```
matthewdias, line 30: see above

app/resources/ama_resource.rb
```diff
@@ -0,0 +1,10 @@
+class AMAResource < BaseResource
+  attributes :description, :start_date, :end_date,
+    :ama_subscribers_count
+
+  has_one :author
+  has_one :original_post
+  has_many :ama_subscribers
+
+  filters :end_date, :start_date, :ama_subscribers_count
```
matthewdias, line 9: Can you add a way to filter by future and past end dates?

app/resources/ama_subscriber_resource.rb
```diff
@@ -0,0 +1,4 @@
+class AMASubscriberResource < BaseResource
+  has_one :user
+  has_one :AMA
```
matthewdias: Add a filter by AMA id

matthewdias (changes requested):
A question, and some changes needed for client hookup

app/resources/ama_resource.rb
cinquemb: What do you mean? If you want past end_dates, can you do a less/greater than end_date query from the client with the current date?

app/models/ama.rb
cinquemb: I dont know, @JoshFabian: do we want to create AMAs that dont have dates or original posts?

JoshFabian: Dates / Original Posts are fine so long as they're editable.

cinquemb: Ok cool yeah, they're all editable.

matthewdias: how will we have "more upcoming" then? set the date to something after the next 3 months? If we forget to edit it before that date enters the 3 month range it'll show up as a date instead of "upcoming".

app/resources/ama_resource.rb
matthewdias: I don't think JR provides this by default. We've implemented it manually in `MediaResource`. for AMAs tho we don't need it to be that precise, so I think simple past and future filters would suffice.

app/models/ama.rb
cinquemb: That makes sense to me. If the ama date is that far out in the future when you are creating it, and you forget, it will complain because a date wasn't given.

app/resources/ama_resource.rb
cinquemb: Ahh i see, ok cool

app/models/ama.rb
```diff
@@ -31,6 +31,8 @@ class AMA < ApplicationRecord
   has_many :ama_subscribers, dependent: :destroy
 
   scope :for_original_post, ->(post) { where(original_post: post) }
+  scope :past_ama, -> { where('end_date <= ?', Date.today) }
+  scope :future_ama, -> { where('end_date > ?', Date.today) }
```
matthewdias: Could we maybe use `>=` to here instead? I realize the logic is weird, but I think we want to be able to display an ama that happened today on the main page as well. That's the way Product Hunt does it at least.

matthewdias (approved):

PR #170, opened 2017-07-10
Filter library entries by followers and media

matthewdias:
so we can show people you are following that have something in their library from media pages


NuckChorris (commented):
Would it be possible to express this as a scope on User that you then `merge()` onto LibraryEntry for this? That way we could reuse the following logic for filtering other stuff

matthewdias:
We already had code for this 😆

PR #171, opened 2017-07-12
interfacing onesignal with notification settings

cinquemb:


app/models/notification_setting.rb
```diff
@@ -28,6 +28,18 @@ class NotificationSetting < ApplicationRecord
...
   belongs_to :user
 
+  def self.get_stream_kitsu_notification_type(type, stream_mentions, feed_id)
+    stream_kitsu_notification_setting_map = {
+      follow: :follows,
+      post: :posts,
+      post_like: :likes,
+      comment_like: :likes,
+      invited: nil,
+      comment: 'mentions'.to_sym ? stream_mentions.include?(feed_id.to_i) : :posts
```
NuckChorris: Won't this ternary always resolve to true, since `'mentions'.to_sym` always returns `:mentions`?  I feel like maybe you swapped the condition and the true clause 🤔

app/models/notification_setting.rb
```diff
@@ -28,6 +28,18 @@ class NotificationSetting < ApplicationRecord
...
+    }
+    stream_kitsu_notification_setting_map[type.to_sym]
```
NuckChorris: This might be cleaner to express as a `case..when`

```ruby
case type
when :follow then :follows
when :post then :posts
when :post_like, :comment_like then :likes
when :comment then stream_mentions.include?(feed_id.to_i) ? :mentions : :posts
else nil
end
```

app/models/notification_setting.rb
```diff
@@ -28,6 +28,18 @@ class NotificationSetting < ApplicationRecord
   enum setting_type: NOTIFICATION_TYPES
   belongs_to :user
 
+  def self.get_stream_kitsu_notification_type(type, stream_mentions, feed_id)
```
NuckChorris: Generally method names prefixed with `get_` are frowned upon in Ruby, as we don't draw a distinction between property getters and method calls.  Also this method name is quite a mouthful lol

Though truthfully I have no ideas on a better name... `type_for_activity`? 🤔 Nothing really quite seems to describe it terribly well

app/models/one_signal_player.rb
```diff
@@ -24,4 +24,26 @@ class OneSignalPlayer < ApplicationRecord
...
+    filtered_player_ids = []
+
+    player_objects = OneSignalPlayer.includes(
+      user: :notification_settings
+    ).where(
+      id: player_ids
+    )
+
+    player_objects.each do |player|
+      user_setting = player.user.notification_settings.where(
```
NuckChorris: I don't think this will use the `includes` data because you included a `where` clause, so this is probably an N+1 query

app/models/one_signal_player.rb
```diff
@@ -24,4 +24,26 @@ class OneSignalPlayer < ApplicationRecord
...
+  def self.filter_player_ids(player_ids, notif_type)
+    filtered_player_ids = []
+
+    player_objects = OneSignalPlayer.includes(
+      user: :notification_settings
+    ).where(
+      id: player_ids
+    )
+
+    player_objects.each do |player|
```
NuckChorris: It's more Ruby-ish to do `each_with_object` or (more commonly) `select` and then `map` to do this task of filtering and then getting IDs

NuckChorris (changes requested):

app/models/one_signal_player.rb
cinquemb: Ahh yeah, i'll just `select` from vs `where`

app/models/notification_setting.rb
cinquemb: Totally right!

app/models/notification_setting.rb
cinquemb: type_for_activity makes sense to me, plus shorter lol

NuckChorris (approved):

PR #172, opened 2017-07-13
Fix mal-sync dates overwritting issue

toyhammered:
couldn't check due to local db not working but this _might_ work....


toyhammered:
hmmm, don't think what I was saying is the issue (according to our tests).....

PR #173, opened 2017-07-13
adding custom enpoint to get issues for library entries for a user

cinquemb:
Should be able to call it like this:

```
curl 'http://localhost:3000/api/edge/library-entries/_issues' -H 'Host: localhost:3000' -H 'User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_11_6) AppleWebKit/601.7.7 (KHTML, like Gecko) Version/9.1.2 Safari/601.7.7' -H 'Accept: application/vnd.api+json' -H 'Accept-Language: en-US,en;q=0.5' --compressed -H 'Authorization: Bearer 5e33becfa3c2ab149def9b05e01d1b007db35a0b60555547a27b968226e95efc'
```

and get back

```
{"rating_ids":[],"reaction_ids":[],"attributes_ids":[]}
```

### DIFF
--- a/app/controllers/library_entries_controller.rb
+++ b/app/controllers/library_entries_controller.rb
@@ -15,24 +15,9 @@ class LibraryEntriesController < ApplicationController
 
   def issues
     authenticate_user!
-    missing_library_data = {
-      rating_ids: [],
-      reaction_ids: [],
-      attributes_ids: []
-    }
-
-    library_entries = LibraryEntry.includes(
-      :media_reaction,
-      media: :media_attribute_votes
-    ).where(user: user)
-
-    library_entries.each do |entry|
-      missing_library_data[:rating_ids] << entry.id if entry.rating.nil? && entry.status.completed?
-      missing_library_data[:reaction_ids] << entry.id if entry.media_reaction.nil?
-      votes = entry.media.media_attribute_votes.select { |mv| mv.vote == :unvoted }
-      missing_library_data[:attributes_ids] << entry.id unless votes.empty?
-    end
-    render json: missing_library_data
+    entries = LibraryEntry.where(user: user)
+    missing = LibraryGapsService.new(entries).missing_engagement_ids
+    render json: missing
   end
 
   def bulk_delete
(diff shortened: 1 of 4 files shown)


app/controllers/library_entries_controller.rb
```diff
@@ -13,6 +13,28 @@ def authorize_operation(operation)
...
+      attributes_ids: []
+    }
+
+    library_entries = LibraryEntry.includes(
+      :media_reaction,
+      media: :media_attribute_votes
+    ).where(user: user)
+
+    library_entries.each do |entry|
+      missing_library_engagement[:rating_ids] << entry.id if entry.rating.nil?
```
trmcnvn: We should also check that the library entry is in its completed state.

app/controllers/library_entries_controller.rb
```diff
@@ -13,6 +13,28 @@ def authorize_operation(operation)
...
+      missing_library_engagement[:reaction_ids] << entry.id if entry.media_reaction.nil?
+      votes = entry.media.media_attribute_votes.select { |mv| mv.vote == :unvoted }
+      missing_library_engagement[:attributes_ids] << entry.id unless votes.empty?
```
trmcnvn: Should check for completed state here too? /cc @JoshFabian

trmcnvn (dismissed):

app/controllers/library_entries_controller.rb
```diff
@@ -13,6 +13,28 @@ def authorize_operation(operation)
...
+      :media_reaction,
+      media: :media_attribute_votes
+    ).where(user: user)
+
+    library_entries.each do |entry|
+      missing_library_data[:rating_ids] << entry.id if entry.rating.nil? && entry.status.completed?
+      missing_library_data[:reaction_ids] << entry.id if entry.media_reaction.nil?
+      votes = entry.media.media_attribute_votes.select { |mv| mv.vote == :unvoted }
+      missing_library_data[:attributes_ids] << entry.id unless votes.empty?
+    end
```
NuckChorris: This is gonna be really really slow (30s+ slow for many libraries), you probably need to do this via joins instead

PR #174, opened 2017-07-14
Interest Timelines

NuckChorris:
# What It Is

Discussion spaces, customized to match the stuff you're interested in!  Interest feeds are a combination of interest-general discussion, discussion of media in your library (excluding Dropped), and discussion of episodes you've seen or chapters you've read.

# How it Works

We create a timeline for each user for each interest, have it automatically follow the interest-global feed.  Then, when they update their library, we generate a MediaFollow (so they can unfollow discussion of specific series, but get it all by default).  When a post is tagged with a unit, we send it to the corresponding unit feed. The interest timeline automatically follows all episodes you've seen in past.

# Potential Improvements

 * ~It might get messy to follow all episode discussion.  If this becomes a problem we could probably limit to the last N episodes you've seen~ ✅ Done!
 * ~The automatic MediaFollow might be better if we invert it, tracking *unfollowed* media instead. That way, we wouldn't try to recreate a follow you deleted~ ✅ Done!
 * ~Perhaps MediaFollow should control episode discussion too?~ ✅ Done!

# What's Done

Everything!

# What Needs Doing

* [x] Build-breaking issue
* [x] Change interest-global discussion into a flat feed
* [x] Backfill MediaFollows
* [x] Dump interest-global follows


JoshFabian:
Damn, this is a nice PR. Well-played!

app/models/library_entry.rb
```diff
@@ -149,12 +149,9 @@ def activity
   end
 
   def kind
```
toyhammered, line 151: what would happen if `kind` is `nil`? I don't think that is _actually_ possible but gonna throw it out there.

app/models/library_entry.rb
```diff
@@ -218,6 +211,18 @@ def kind
     end
   end
 
+  after_save if: :progress_changed? do
```
toyhammered: should this really be an `after_save`? Might be better to do `after_commit on: :update` and then wrap the statement in an `if progress_changed?` Only saying this because `after_save` deals with both `create` and `update` and you already have a create below.

toyhammered (commented):
I understood about 50% of the code (I have not really dug deeply into feeds), but as for any code errors didn't really see many (left some comments on questionable stuff).

app/models/library_entry.rb
NuckChorris: Explosions, probably.  But it's something that should never happen, so I'm not terribly worried about that

app/models/library_entry.rb
NuckChorris: Ah, you're right

PR #181, opened 2017-07-18
public groups

matthewdias:
- properly hides posts in private groups
- allows posting, commenting, and liking by non-members (that are not banned)

### DIFF
--- a/app/models/post.rb
+++ b/app/models/post.rb
@@ -55,6 +55,11 @@ class Post < ApplicationRecord
   has_many :comments, dependent: :destroy
 
   scope :in_group, ->(group) { where(target_group: group) }
+  scope :visible_for, ->(user) {
+    joins(:target_group).merge(Group.visible_for(user))
+                        .or(joins(:target_group)
+                        .where(target_group_id: nil))
+  }
 
   validates :content, :content_formatted, presence: true
   validates :media, presence: true, if: :spoiled_unit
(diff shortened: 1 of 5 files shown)


JoshFabian:
What happens if I'm a non-member that has never been banned and I'm behaving in a way that should get me banned?

matthewdias:
Ah, thanks for the catch. We need to have a way to ban non-members.

NuckChorris:
Do we really want the ability for drive-by Posts?

NuckChorris:
Also I think GroupBan can be applied to nonmembers with some minor frontend tweaks?

matthewdias:
Yeah looking at it now I think it can. Regarding drive-bys, I think with the changes being made to feeds, groups will become more important for discussion that doesn't align with Interests. I really don't want to have to join a group to make one post, and then have all the posts for that group in my following feed, for example. Basically the way I see groups after this change is: A group is a topic-focused feed that anyone can interact with, provided they are not banned, and if I am active enough in a group that I want to see all its posts, I can join the group and have them appear in my following feed. Happy to revert if it's a dumb idea.

JoshFabian:
I will say that's it's a little bit strange for a non-group member to participate in a group, and it might hurt some of the intimacy of what a group offers. That said, I'm happy to give this a try - if it becomes problematic, we can always revert!

JoshFabian:
I will say that's it's a little bit strange for a non-group member to participate in a group, and it might hurt some of the intimacy of what a group offers. That said, I'm happy to give this a try - if it becomes problematic, we can always revert!

NuckChorris (approved):

NuckChorris (approved):

PR #182, opened 2017-07-19
WIP: Move user stats to background worker.

toyhammered:
not really sure if this is the best way to do it. Also just fixed some rubocop errors.


toyhammered:
merge #188 first and then I will update this + fix category tests better.

app/workers/user_stats_worker.rb
```diff
@@ -0,0 +1,70 @@
+# HACK: will want to try and find a way to abstract this into something nicer.
+# Need to find a similarity between these stats.
+
+class UserStatsWorker
+  include Sidekiq::Worker
+
+  def perform(action, kind, user, library_entry)
+    library_event = LibraryEvent.create_for(action, library_entry) unless action == :destroyed
+
+    case type
```
matthewdias: Shouldn't this be `case action` ?

JoshFabian:
@toyhammered what's going on here?

app/models/library_entry.rb
```diff
@@ -241,60 +239,15 @@ def kind
...
-      Stat::AnimeAmountConsumed.increment(user, self)
-      Stat::AnimeFavoriteYear.increment(user, self)
-      Stat::AnimeActivityHistory.increment(user, library_event)
-    when :manga
-      Stat::MangaCategoryBreakdown.increment(user, self)
-      Stat::MangaAmountConsumed.increment(user, self)
-      Stat::MangaFavoriteYear.increment(user, self)
-      Stat::MangaActivityHistory.increment(user, library_event)
-    end
+    Stat::AddWorker.perform_async(kind, user, id)
```
NuckChorris: We may add non-library related stats, so this should be named something like `Stat::LibraryCreateWorker`

app/workers/stat/update_worker.rb
```diff
@@ -0,0 +1,31 @@
...
+    def perform(kind, user, library_entry_id)
+      library_entry = LibraryEntry.find(library_entry_id)
+      library_event = LibraryEvent.create_for(:updated, library_entry)
+
+      case kind
+      when :anime
+        # TODO: Change this before merging PR 201
+        Stat::AnimeActivityHistory.increment(user, library_event)
+        # special case checking if progress was increased or decreased
+        if progress > progress_was
```
NuckChorris: Where did these variables even come from

toyhammered: oh god... Good catch, need to pass those in I think....

app/models/library_entry.rb
toyhammered: yup, will rename all 3 of them.

app/workers/stat/library_create_worker.rb
```diff
@@ -0,0 +1,25 @@
+class Stat
+  class LibraryCreateWorker
+    include Sidekiq::Worker
+
+    def perform(kind, user, library_entry_id)
+      library_entry = LibraryEntry.find(library_entry_id)
+      library_event = LibraryEvent.create_for(:added, library_entry)
```
NuckChorris: uhhhhh I don't think this'll work

iirc `LibraryEvent.create_for` uses the dirty tracking from ActiveRecord, but since this is in a background worker, it won't have the changes that the frontend made

app/models/concerns/stat/amount_consumed.rb
```diff
@@ -39,38 +39,38 @@ def increment(user, library_entry, updated = false)
...
+            progress_to_time(library_entry, progress_difference(options))
         end
 
         record.save!
       end
 
-      def progress_difference(le)
-        le.progress_changed? ? (le.progress - le.progress_was) : le.progress
+      def progress_difference(options)
+        options[:progress_changed] ? (options[:progress] - options[:progress_was]) : options[:progress]
```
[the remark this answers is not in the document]
toyhammered: 🖕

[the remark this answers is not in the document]
NuckChorris: If the bot says wrap it you gotta wrap it

spec/models/stat/anime_activity_history_spec.rb
```diff
@@ -0,0 +1,76 @@
+# rubocop:disable Metrics/LineLength
```
NuckChorris: There's no reason to comment out code — your ~mistake~ _code_ is forever recorded in the git history, where we can recover it 😄

app/models/library_entry.rb
```diff
@@ -241,54 +239,36 @@ def kind
...
-      Stat::AnimeCategoryBreakdown.decrement(user, self)
-      Stat::AnimeAmountConsumed.decrement(user, self)
-      Stat::AnimeFavoriteYear.decrement(user, self)
-    when :manga
-      Stat::MangaCategoryBreakdown.decrement(user, self)
-      Stat::MangaAmountConsumed.decrement(user, self)
-      Stat::MangaFavoriteYear.decrement(user, self)
-    end
+    Stat::LibraryDestroyWorker.perform_async(
+      kind, user, self,
```
NuckChorris: Can't do `self`, need to pass `id` instead

app/workers/stat/library_destroy_worker.rb
```diff
@@ -0,0 +1,22 @@
+class Stat
+  class LibraryDestroyWorker
+    include Sidekiq::Worker
+
+    def perform(kind, user, library_entry, options)
+      case kind
```
NuckChorris: And then load the `LibraryEntry.find(id)` here

app/models/library_entry.rb
toyhammered: uhhh... the id won't exist because it would be deleted.

NuckChorris: Then neither will `self`

PR #183, opened 2017-07-21
Feature/uploads

cinquemb:


app/controllers/uploads_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+
+  def bulk_create
+    files_to_upload = params[:files].map { |file| { user: user, content: file } }
+    uploads = Upload.create(files_to_upload)
+    render json: serialize_entries(uploads)
+  end
+
+  def update
+    upload = Upload.find(id: params[:id], user: user)
+    render_jsonapi serialize_error(401, 'Not permitted'), status: 401 unless upload
```
trmcnvn: Probably need a `return` here, no?

app/controllers/uploads_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+
+    if params.key?(:post)
+      upload.post = params[:post]
+    elsif params.key?(:comment)
+      upload.comment = params[:comment]
+    else
+      render_jsonapi serialize_error(400, 'Needs related post or comment field'), status: 400
```
trmcnvn: :arrow_up:

app/resources/upload_resource.rb
```diff
@@ -0,0 +1,10 @@
+class UploadResource < BaseResource
+  include ScopelessResource
+  attribute :content
```
trmcnvn: Should this be using the attachment formatter?

app/controllers/uploads_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+    end
+    upload.save
+    render json: serializer.serialize_to_hash(UploadResource.new(upload, context))
+  end
```
NuckChorris: Is this necessary?  Doesn't jsonapi-resources give us this for free?

app/models/upload.rb
```diff
@@ -0,0 +1,39 @@
...
+#  fk_rails_127111e6ac  (post_id => posts.id)
+#  fk_rails_15d41e668d  (user_id => users.id)
+#  fk_rails_62b822a2d6  (comment_id => comments.id)
+#
+# rubocop:enable Metrics/LineLength
+
+class Upload < ApplicationRecord
+  belongs_to :user, required: true
+  belongs_to :post
+  belongs_to :comment
```
NuckChorris: Probably best to just go with polymorphic here, there's no significant negative effects on performance or features in this situation.  Also, a corresponding `has_many :uploads, as: 'owner'` on Post and Comment is probably needed

NuckChorris (changes requested):

app/policies/upload_policy.rb
```diff
@@ -0,0 +1,5 @@
+class UploadPolicy < ApplicationPolicy
+  def destroy?
+    false
```
NuckChorris: I feel like we should allow destruction so that, if the user clicks the `(X)` on the upload item in the frontend after it's already uploaded, it'll just immediately delete it instead of waiting to be cleaned out by the automated task

app/resources/upload_resource.rb
```diff
@@ -0,0 +1,9 @@
+class UploadResource < BaseResource
+  include ScopelessResource
+  attribute :content, format: :attachment
+  attributes :owner_id, :owner_type
```
NuckChorris: This should be `has_one :owner, polymorphic: true`, and you _might_ need an empty `OwnerResource` similar to our existing `ItemResource`.  Also needs a `has_many :uploads` on Comment and Post resources

app/workers/upload_removal_worker.rb
```diff
@@ -0,0 +1,12 @@
+class UploadRemovalWorker
+  include Sidekiq::Worker
+
+  def perform
+    Upload.where(
+      post: nil,
+      comment: nil
+    ).where(
+      ['created_at > ?', 11.hours.ago]
```
NuckChorris: Probably should be a scope (`orphaned` maybe?) on `Upload`, it's best to avoid directly using `where` outside of the model.

NuckChorris (changes requested):

JoshFabian:
Posts should also belong to a user. In the future, we may want to do similar to Facebook / Twitter and show media in the sidebar that a user has uploaded (As posts, not comment replies)

NuckChorris:
@JoshFabian Not sure exactly what you mean, if you mean Uploads should be tied to User, don't worry — that's already in there

app/resources/comment_resource.rb
```diff
@@ -9,6 +9,7 @@ class CommentResource < BaseResource
   has_one :parent
   has_many :likes
   has_many :replies
+  has_many :uploads
```
NuckChorris, line 12: I think you're forgetting `as: :owner` here

NuckChorris (changes requested):

app/resources/comment_resource.rb
cinquemb: hmm, i dont see that option here: jsonapi-resources.com/v0.9/guide/resources.html

NuckChorris: Ah shit you're right, I mistook this for the model

NuckChorris (approved):

PR #184, opened 2017-07-22
Nest episodes in anime admin page

matthewdias:
are nested attributes okay?

### DIFF
--- a/app/models/concerns/episodic.rb
+++ b/app/models/concerns/episodic.rb
@@ -30,6 +30,7 @@ module Episodic
 
   included do
     has_many :episodes, as: 'media', dependent: :destroy, inverse_of: 'media'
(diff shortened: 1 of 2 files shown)


NuckChorris:
I don't think we need `accepts_nested_attributes_for`, do we?  I thought rails_admin handled all that stuff via AJAX

matthewdias:
this is apparently the only way to get an interface that allows deleting episodes from the media page.

NuckChorris:
How well does it scale to like, 1000 episodes?

Does it crash?

matthewdias:
doesn't crash, but it takes forever to generate all the html, and we get the gateway timeout, so long shows are inaccessible basically. i think ill revert this, and just see if i can tweak it so that episodes are displayed in order, or with their number in the info pane.

PR #185, opened 2017-07-22
Fix policies for follows

matthewdias:


app/policies/group_member_policy.rb
```diff
@@ -3,25 +3,16 @@ class GroupMemberPolicy < ApplicationPolicy
 
   def update?
     required_permission = record.pleb? ? :members : :leaders
-    has_group_permission?(required_permission)
+    is_owner? || has_group_permission?(required_permission)
```
matthewdias, line 6: Will this allow users to give themselves permissions that they shouldn't be able to? @NuckChorris

app/policies/follow_policy.rb
```diff
@@ -1,15 +1,7 @@
...
   def create?
     record.follower == user
   end
+  alias_method :update?, :create?
   alias_method :destroy?, :create?
-
-  def visible_attributes(all)
-    is_owner? ? all : all - %i[hidden]
-  end
-  alias_method :editable_attributes, :visible_attributes
```
NuckChorris, line 14: This was intentional, I don't think `hidden` should be public knowledge

app/policies/group_member_policy.rb
NuckChorris: Yeah it will, you'll need to use `editable_attributes` to protect against that

NuckChorris (changes requested):

app/policies/follow_policy.rb
matthewdias: It was failing because follow doesn't have a `user` so `is_owner` doesn't work. trying to do something like `record.follower == user` gave `no such method follower for <#Class...>`

NuckChorris: So I just read my own code and it looks like `visible_attributes` is run with a `record` of the target row but `editable_attributes` is run with the class, so... I guess we can honestly just remove `editable_attributes` on Follow (the only person who can edit is owner, and they can edit anything about it idgaf), but we need to figure something out for GroupMember

matthewdias:
closed by https://github.com/hummingbird-me/hummingbird-server/commit/3d6ea622e3cd2dbddd77cadb304072628acf5acd

PR #187, opened 2017-07-23
Set nsfw on library entries

matthewdias:


app/models/library_entry.rb
```diff
@@ -164,6 +164,9 @@ def kind
       kind = media_type&.underscore
       send("#{kind}=", media) if kind
     end
+
+    self.nsfw = AgeRatings::UNSAFE_AGE_RATINGS.include? media.age_rating if media_id_changed?
```
NuckChorris: This can just be `media.nsfw?`

NuckChorris (changes requested):

NuckChorris (approved):

NuckChorris (approved):

PR #190, opened 2017-07-25
Merge data ourselves before MAL Sync

NuckChorris:


cinquemb:
Do we need to change `started_at` and `my_finish_date` in `ListImport::MyAnimeListXML::Row` to mimic the same format as the dates incoming on mal?

PR #191, opened 2017-07-25
reaction skips

matthewdias:
- sort ids by `reaction_skipped`, `finished_at`
- if `reaction_skipped` is `ignored`, don't send the id


db/migrate/20170725220827_add_reaction_skipped_to_library_entry.rb
```diff
@@ -0,0 +1,5 @@
+class AddReactionSkippedToLibraryEntry < ActiveRecord::Migration
+  def change
+    add_column :library_entries, :reaction_skipped, :integer, null: false, default: 0
```
NuckChorris: We can't do this, it'll lock the LibraryEntry table for the duration.  Instead, we need three migrations:

1. add the column without null constraints or default values
2. backfill the default value with UpdateInBatches
3. add the default value and null constraint

NuckChorris (changes requested):

NuckChorris (approved):

PR #193, opened 2017-07-28
Feature/community recommendations

cinquemb:


app/models/community_recommendation.rb
```diff
@@ -0,0 +1,71 @@
...
+  belongs_to :anime
+  belongs_to :manga
+  belongs_to :drama
+  belongs_to :media, polymorphic: true, required: true
+  belongs_to :community_recommendation_request, required: true
+  has_many :reasons, class_name: 'Post'
+
+  before_validation do
+    self.media = retrieve_media
+  end
```
NuckChorris: Now might be a good time to break this syncing-media-with-triple-column thing into a concern 🤔

app/models/community_recommendation.rb
```diff
@@ -0,0 +1,71 @@
...
+      &:user
+    ).map(&:notifications)
+    community_recommendation_request.user.profile_feed.activities.new(
+      target: reason,
+      actor: reason.user,
+      object: self,
+      foreign_id: self,
+      verb: self.class.name.underscore,
+      time: Time.now,
+      to: follower_notifications
```
NuckChorris: As with AMAs, you're gonna have to create a new Feed instead and have the notification feeds follow it, since Stream limits us to 100 feeds in the `to` field.

In that case, you probably won't need this method or the corresponding worker either, and `WithActivity` should have you covered.

In future, too, you should be aware that `new` doesn't actually save the activity, it merely generates one.  You need to call `save` on it to actually save

app/models/community_recommendation_follow.rb
```diff
@@ -0,0 +1,29 @@
...
+# Foreign Keys
+#
+#  fk_rails_798884902b  (community_recommendation_request_id => community_recommendation_requests.id)
+#  fk_rails_caa37f6fd5  (user_id => users.id)
+#
+# rubocop:enable Metrics/LineLength
+
+class CommunityRecommendationFollow < ApplicationRecord
+  belongs_to :user, required: true
+  belongs_to :community_recommendation_request, required: true
```
NuckChorris, line 26: You might be able to save some sanity with:

```ruby
belongs_to :request, required: true, class_name: 'CommunityRecommendationRequest'
```

Which would allow you to do stuff like `where(request: request)` instead 😛

NuckChorris (changes requested):

NuckChorris (approved):

JoshFabian:
If this is good to go, can it be merged until we're ready to work on front-end, in order to avoid conflicts rebasing later?

cinquemb:
I dont see why not, i'll rebase and merge it in a bit

PR #194, opened 2017-07-28
Add media_reactions counter cache to User

matthewdias:
Do these migrations look okay?

### DIFF
--- a/db/migrate/20170726230852_fill_reactions_counter_cache.rb
+++ b/db/migrate/20170726230852_fill_reactions_counter_cache.rb
@@ -1,7 +1,9 @@
+require_dependency 'counter_cache_resets'
+
 class FillReactionsCounterCache < ActiveRecord::Migration
   disable_ddl_transaction!
 
   def change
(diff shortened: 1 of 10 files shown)


db/migrate/20170726230852_fill_reactions_counter_cache.rb
```diff
@@ -0,0 +1,7 @@
+class FillReactionsCounterCache < ActiveRecord::Migration
+  disable_ddl_transaction!
+
+  def change
+    User.all.each { |user| User.reset_counters(user.id, :media_reactions) }
```
NuckChorris: You probably wanna use `find_each` in this scenario because `each` will load all Users into RAM at once, while `find_each` loads them in batches of 1000.

Alternatively, the `CounterCacheResets` module in `lib/counter_cache_resets.rb` can actually do this a gajillion times faster with a zany sequence of generated SQL statements

NuckChorris (changes requested):

NuckChorris (approved):

PR #196, opened 2017-07-29
Add absolute numbers to episodes

matthewdias:
@NuckChorris do you think we should:
- do any syncing between `number` and `absolute_number`
- switch all the logic to use `absolute_number` instead? (probably cause polymorphism issues)

actually thinking about the implications of the above, i feel like it would be best to just overwrite `number` with the values of `absolute_number`, and not add the new column.

### DIFF
--- a/app/controllers/feeds_controller.rb
+++ b/app/controllers/feeds_controller.rb
@@ -68,8 +68,8 @@ class FeedsController < ApplicationController
   end
 
   def feed_visible?
-    case params[:group]
-    when 'media', 'media_aggr'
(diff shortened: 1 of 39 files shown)


NuckChorris:
Hmm, maybe migrate from `number` to `relative_number`? 🤔 Normally renaming a column is a long, arduous process, but we can probably get away with:

1. Add `relative_number`
2. Copy stuff from `number` into it
3. Generate new absolute `number` values

NuckChorris (approved):

PR #197, opened 2017-07-29
Fix post visible_for scope

matthewdias:
yeah, its ugly, but this is all i could come up with after hours of trying, so its beautiful to me!

### DIFF
--- a/app/models/post.rb
+++ b/app/models/post.rb
@@ -63,11 +63,30 @@ class Post < ApplicationRecord
   has_one :ama, foreign_key: 'original_post_id'
   has_many :reposts, dependent: :delete_all
 
+  scope :sfw, -> { where(nsfw: false) }
   scope :in_group, ->(group) { where(target_group: group) }
   scope :visible_for, ->(user) {
-    joins(:target_group).merge(Group.visible_for(user))
-                        .or(joins(:target_group)
-                        .where(target_group_id: nil))
+    scope = user && !user.sfw_filter? ? all : sfw
+
+    left_join = <<~SQL
+      LEFT OUTER JOIN groups g
+      ON g.id = target_group_id
+    SQL
+    group_visible = <<~SQL
+      posts.target_group_id IS NULL
+      OR (g.privacy = 0 OR g.privacy = 2)
+    SQL
+
+    return scope.joins(left_join).distinct.where(group_visible) unless user
+
+    scope.joins(left_join).distinct.where(<<~SQL)
+      #{group_visible}
+      OR g.id IN (
+        SELECT group_members.group_id
+        FROM group_members
+        WHERE group_members.user_id = #{user.id}
+      )
+    SQL
   }
 
   validates :content, :content_formatted, presence: true, unless: :uploads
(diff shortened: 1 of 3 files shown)


app/models/post.rb
```diff
@@ -55,11 +55,30 @@ class Post < ApplicationRecord
...
+    return scope.joins(left_join).distinct.where(group_visible) unless user
+
+    scope.joins(left_join).distinct.where(<<~SQL)
+      #{group_visible}
+      OR g.id IN (
+        SELECT group_members.group_id
+        FROM group_members
+        WHERE group_members.user_id = #{user.id}
+      )
+    SQL
```
NuckChorris: Hmm, was there an issue which prevented implementing these within the existing `.or`?

matthewdias: yup, iirc you can't use `.or` with joins unless they're nearly exactly the same. merging with `Group.visible_for` breaks that.

NuckChorris: Yeah I think you gotta have identical joins on either side of the `or`, hence the `joins(:target_group)` on either side previously

NuckChorris: Instead of `.merge(Group.visible_for)` could we do `where(target_group_id: Group.visible_for`? I think that'll generate an `IN` with a subquery

matthewdias: i'll give it a try

app/models/post.rb
```diff
@@ -55,12 +55,9 @@ class Post < ApplicationRecord
...
 
-  scope :in_group, ->(group) { where(target_group: group) }
-  scope :visible_for, ->(user) {
-    joins(:target_group).merge(Group.visible_for(user))
-                        .or(joins(:target_group)
-                        .where(target_group_id: nil))
-  }
+  scope :sfw, (-> { where(nsfw: false) })
+  scope :in_group, (->(group) { where(target_group: group) })
+  scope :visible_for, (->(user) { where(target_group_id: Group.visible_for(user)) })
```
NuckChorris: Is there a reason for these parens? I mean, they're not hurting anyone, they're just kind of... odd?

matthewdias: rubocop was complaining about em. at least i think it was... 🤔  lemme check my editor

NuckChorris: Also, if I'm not mistaken, aren't you missing a call to add the "or not in group"?

matthewdias: yeah rubocop asked for the parens. and, yes, yes i am.

app/policies/post_policy.rb
```diff
@@ -36,8 +36,9 @@ def group
 
   class Scope < Scope
     def resolve
-      scope.where.not(user_id: blocked_users)
-      #scope.visible_for(user).where.not(user_id: blocked_users)
+      return scope if is_admin?
+      scope.visible_for(user).where.not(user_id: blocked_users) if see_nsfw?
+      scope.sfw.visible_for(user)
```
NuckChorris: Won't this circumvent blocking if I have my SFW filter enabled

NuckChorris (approved):

PR #198, opened 2017-07-30
Unit Prediction

matthewdias:


app/models/anime.rb
```diff
@@ -103,6 +104,12 @@ def season_year
     end
   end
 
+  def update_episode_count_guess(guess)
+    self.episode_count_guess = guess
+    save!
```
NuckChorris: `update(episode_count_guess: guess)` makes this a bit shorter

app/models/library_entry.rb
```diff
@@ -209,6 +209,21 @@ def kind
...
       user.update_profile_completed!
     end
+
+    if progress_changed?
+      case kind
+      when :anime
+        guess = [(progress + 1), anime.default_progress_limit].min
+        anime.update_episode_count_guess(guess) if
+          !anime.episode_count &&
+          (!anime.episode_count_guess || anime.episode_count_guess <= guess)
```
NuckChorris: Does this conditional logic really belong in LibraryEntry or would it make more sense in the media?

app/models/library_entry.rb
```diff
@@ -209,6 +209,21 @@ def kind
       user.update_feed_completed!
       user.update_profile_completed!
     end
+
+    if progress_changed?
+      case kind
+      when :anime
+        guess = [(progress + 1), anime.default_progress_limit].min
+        anime.update_episode_count_guess(guess) if
```
NuckChorris: The `guess` line could probably be unified for all media types as `guess = [(progress + 1), media.default_progress_limit].min`, but it'd be nice to get `media.update_unit_count_guess(guess)` too.

I generally try to make it a rule of keeping as many method names common between media types as possible, so that we don't need conditionals to control behavior (it's just inherent as part of the different model type)

db/migrate/20170729210024_fill_unit_count_guess_for_media.rb
```diff
@@ -0,0 +1,20 @@
...
+  disable_ddl_transaction!
+
+  def change
+    Anime.where(episode_count: nil)
+         .joins('INNER JOIN library_entries le ON le.anime_id = anime.id')
+         .where('le.progress = (SELECT MAX(le.progress) FROM library_entries le WHERE le.media_id = anime.id)')
+         .pluck('anime.id, le.progress').each do |a, pr|
+           anime = Anime.find(a)
+           anime.update_episode_count_guess([(pr + 1), anime.default_progress_limit].min)
+         end
```
NuckChorris: Damn, you're getting good at this 😱

app/models/anime.rb
```diff
@@ -103,6 +104,14 @@ def season_year
     end
   end
 
+  def update_unit_count_guess(guess)
+    if !episode_count &&
```
NuckChorris: Would this be easier to express as a bunch of guard clauses?

```
return if episode_count
return if episode_count_guess && episode_count_guess > guess
```

matthewdias: yes

NuckChorris (approved):

PR #202, opened 2017-08-05
Determine date format before import

matthewdias:


app/models/list_import/my_anime_list.rb
```diff
@@ -50,13 +50,25 @@ def ensure_list_is_public
...
 
+    def date_format
+      @date_format ||= data.each do |row|
+        format = nil
+        row.fetch_values('start_date_string', 'finish_date_string').each do |date|
+          place1, place2 = date.split('-').map(&:to_i)
+          break format = '%d-%m-%y' if place1 > 12
+          break format = '%m-%d-%y' if place2 > 12
+        end
+        break format if format
```
NuckChorris: Shouldn't this be a `return`? 🤔

matthewdias: `return` would exit the method and wouldn't set the value of `@date_format`. at least I thought it would.

NuckChorris: Ah, that's true.  But I think `each` will return the list it iterated over, not the value of the last `break`

matthewdias: `break` will return the value for the `each`, but this made me notice that if you go through the whole list without finding a date format, it will return the array. so maybe I should set `nil` if we've iterated through the whole list? This already works as is, I think because it will just fail when trying to parse the date in the row.

NuckChorris (approved):

PR #203, opened 2017-08-05
Hashtag Parsing

NuckChorris:
Add HashtagsFilter to parse hashtags out of content.

Thanks to Twitter, it has some of the best specs in the entire codebase, and handles pretty much any language you throw at it 👌

### DIFF
--- a/lib/html/pipeline/hashtags_filter.rb
+++ b/lib/html/pipeline/hashtags_filter.rb
@@ -1,0 +1,76 @@
+module HTML
+  class Pipeline
+    class HashtagsFilter < Filter
+      # These regexex are reproduced from the twitter/tweet-text library, which is licensed under
+      # the Apache 2.0 license.  The originals can be found at the following URL:
+      #
+      # https://github.com/twitter/twitter-text/blob/master/java/src/com/twitter/Regex.java
+
+      # For these, we don't worry about the *exact* details of the character classes, so we just use
+      # \p{} groups directly.  Ruby 2.3, in fact, has nearly-perfect sets for these.  For more
+      # information about the complexities and nuances of using \p{} groups across platforms,
+      # Twitter has kindly documented that here:
+      #
+      # https://github.com/twitter/twitter-text/blob/master/unicode_regex/README
+      HASHTAG_LETTERS_AND_MARKS = '\p{L}\p{M}'.freeze
+      HASHTAG_NUMERALS = '\p{Nd}'.freeze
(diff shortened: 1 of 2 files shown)


spec/lib/html/pipeline/hashtags_filter_spec.rb
```diff
@@ -0,0 +1,583 @@
...
+
+  it 'should autolink a hiragana hashtag preceded by a space and followed by a bracket' do
+    twitter_example(
+      text: 'ひらがな #ひらがな」ひらがな',
+      expected: 'ひらがな <a href="https://kitsu.io/posts?query=%23ひらがな" title="#ひらがな" class="hashtag">#ひらがな</a>」ひらがな',
+      hashtags: %w[ひらがな]
+    )
+  end
+
+  it 'should autolink a hiragana hashtag preceded by a space and followed by a edge' do
```
toyhammered, line 406: 'an edge'

lib/html/pipeline/hashtags_filter.rb
```diff
@@ -0,0 +1,76 @@
...
+        result[:hashtags] ||= []
+        doc.xpath('./text()').each do |node|
+          # Skip if we're inside a link
+          next unless node.ancestors('a').empty?
+          # Otherwise linkify hashtags
+          new_text = node.text.gsub(VALID_HASHTAG) do
+            # rubocop:disable Style/PerlBackrefs
+            text = [$2, $3].join
+            result[:hashtags] << $3
+            "#{Sanitize.fragment($1)}#{link_for($3, text)}"
```
matthewdias, line 54: what's happening on these lines?

NuckChorris: the `$1/2/3/4/5/6/7/8/9` variables are backreferences into the regex, matching first/second/third/etc. groups

matthewdias: yeah i get that, but i don't know what the values of $1, 2 and 3 are supposed to be in this case 😅

NuckChorris: $2 is the hash symbol and $3 is the text of the hashtag, but the complicated part is... $1 is a reference to the character or characters leading up to it, which are included I think so that it can match based on what those are

matthewdias (approved):
👌 I assume we'll add it to the pipeline when we're ready to launch?

NuckChorris:
Yep, if we added it now, it'd start linking things to nowhere

PR #205, opened 2017-08-07
Set current when progress changed

matthewdias:
I think we should do this server-side. It causes confusing behavior when we do it on the client.

### DIFF
--- a/app/models/library_entry.rb
+++ b/app/models/library_entry.rb
@@ -189,17 +189,17 @@ class LibraryEntry < ApplicationRecord
       self.progress = media.progress_limit if media&.progress_limit
     end
 
-    return if imported
+    unless imported
+      self.progressed_at = Time.now if status_changed? || progress_changed?
 
-    self.progressed_at = Time.now if status_changed? || progress_changed?
+      if status_changed?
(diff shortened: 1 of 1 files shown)


NuckChorris (approved):

app/models/library_entry.rb
```diff
@@ -187,16 +187,18 @@ def kind
...
-      self.started_at ||= Time.now if current?
+      self.progressed_at = Time.now if status_changed? || progress_changed?
+
+      if status_changed?
+        self.started_at ||= Time.now if current?
+        self.started_at ||= Time.now if completed?
+        self.finished_at ||= Time.now if completed?
+      end
+
+      self.status = :current if progress_changed? && !status_changed?
```
NuckChorris, line 201: Isn't this basically "set status to current if the progress has changed and the status hasn't" and won't that just constantly set things to current?

matthewdias: is that not okay? or do we want to check that it's not current?

NuckChorris: Y'know, now that I think more it does seem like it would be fine 🤔

PR #207, opened 2017-08-11
Import/hulu

cinquemb:


app/services/hulu_mapping_service.rb
```diff
@@ -0,0 +1,75 @@
+# Service to sync Hulu API with Kitsu
+class HuluMappingService
+  HULU_URL = 'https://distribution.hulu.com/api/v1/'.freeze
+
+  attr_reader :since
+
+  # Initialize services with GetStream webhook request
```
NuckChorris: I think the comment here is wrong?

matthewdias (dismissed):
Looks good, I think this just needs to be updated for the mapping changes.

app/services/hulu_mapping_service.rb
```diff
@@ -0,0 +1,75 @@
...
+
+  # Initialize service with optional datetime
+  def initialize(since = nil)
+    @since = since
+  end
+
+  # Grabs all available series from hulu, maps them as wel as their episodes
+  def sync_series_and_episodes
+    ActiveRecord::Base.logger = Logger.new(nil)
+    Chewy.strategy(:bypass)
```
NuckChorris: Are these two lines intentional or are they something from development that leaked in?

app/services/hulu_mapping_service.rb
```diff
@@ -0,0 +1,75 @@
...
+    offset = 0
+    limit = 10_000
+    series_res = Typhoeus.get(
+      "#{HULU_URL}/series?guid=#{guid}&primary_category=animation&offset=#{offset}&limit=#{limit}"
+    )
```
NuckChorris: Might I suggest splitting retrieval/deserialization from actually processing?

You could just make a method which yields the JSON of each series, I think that would make this code a bit easier to read

app/services/hulu_mapping_service.rb
```diff
@@ -0,0 +1,75 @@
...
+      }
+      kitsu_anime = Mapping.guess('Anime', mapping_object)
+      next unless kitsu_anime
+      get_hulu_episodes_for_series(kitsu_anime, series_id)
+    end
+  end
+
+  def get_hulu_episodes_for_series(kitsu_anime, series_id)
+    # creating mapping with series to kitsu anime
+    kitsu_anime.mappings.where(
```
NuckChorris: This seems to be doing something unrelated to the name of the method, which could startle future developers

app/services/hulu_mapping_service.rb
```diff
@@ -0,0 +1,75 @@
...
+      external_site: 'hulu',
+      external_id: series_id
+    ).first_or_create
+
+    # grab available episodes from hulu with anime
+    since_param = since.advance(hours: -24).strftime('%F') if since
```
NuckChorris: It might be nicer to say `(since - 24.hours).strftime('%F')`, since Rails offers us that syntactical sugar

app/services/hulu_mapping_service.rb
```diff
@@ -0,0 +1,75 @@
...
+    ep_url += "&since=#{since_param}" if since
+    episode_res = Typhoeus.get(
+      ep_url
+    )
+    return unless episode_res.sucess?
+    hulu_episodes = JSON.parse(episode_res.body)
+    sync_episodes_for_anime(hulu_episodes)
+  end
+
+  def sync_episodes_for_anime(hulu_episodes)
```
NuckChorris: Generally if a method says `for_x`, then the parameter will be of type `x`

Thus, it seems surprising to take a list of episodes as a parameter

config/application.rb
```diff
@@ -38,6 +38,8 @@ class Application < Rails::Application
     # Eager-load list_import files
     import_dirs = Dir['lib/list_import'].map { |d| File.expand_path(d) }
     config.eager_load_paths += import_dirs
+    # Include all services files
+    config.eager_load_paths += Dir["#{config.root}/app/services/**/"]
```
NuckChorris: I don't think you need this, Rails 4+ automatically adds any directories directly under `app` to this, so `app/services` is eager loaded by default (and we add `app/*/concerns` ourselves)

Doesn't look like you have anything that would rely on this, if I'm not mistaken

NuckChorris (changes requested):

app/services/hulu_mapping_service.rb
```diff
@@ -0,0 +1,83 @@
...
+
+  attr_reader :since
+
+  # Initialize service with optional datetime
+  def initialize(since = nil)
+    @since = since
+  end
+
+  # Grabs all available series from hulu
+  def get_series_for_episodes(limit)
```
NuckChorris: This name seems like a pretty long name, is there a reason for the `for_episodes` part?

app/services/hulu_mapping_service.rb
```diff
@@ -0,0 +1,83 @@
...
+    hulu_anime = get_series_for_episodes(limit)
+    return if hulu_anime.nil?
+    hulu_anime.each do |item|
+      item = item.deep_symbolize_keys
+      series_id = item[:id]
+      title = item[:name].titleize
+      mapping_object = {
+        title: title
+      }
+      kitsu_anime = Mapping.guess('Anime', mapping_object)
```
NuckChorris: `Mapping.guess` should generally be preceded by an attempt to find an existing mapping with `Mapping.lookup`

Also you can probably just do `Mapping.guess('Anime', title: title)` instead of building `mapping_object`

app/services/hulu_mapping_service.rb
```diff
@@ -0,0 +1,83 @@
...
+      next unless kitsu_anime
+      create_kitsu_hulu_series_mapping(kitsu_anime, series_id)
```
NuckChorris: This could probably be shortened to `create_mapping`, no?

app/services/hulu_mapping_service.rb
```diff
@@ -0,0 +1,83 @@
...
+      hulu_episodes = get_hulu_episodes_for_series(series_id, limit)
```
NuckChorris: `get_episodes`?

NuckChorris (changes requested):

app/services/hulu_mapping_service.rb
```diff
@@ -0,0 +1,81 @@
...
+    # creating mapping with series to kitsu anime
+    kitsu_anime.mappings.where(
+      external_site: 'hulu',
+      external_id: series_id
+    ).first_or_create
+  end
+
+  def get_episodes(series_id, limit)
+    # grab available episodes from hulu with anime
+    since_param = (since - 24.hours).strftime('%F') if since
```
NuckChorris: So... wait, does `since` actually not mean "since"?  Like, "since" as a parameter generally means "get me everything newer than this", not "get me anything newer than 24 hours before this"

Should the `24.hours.ago` be maybe part of the worker?

app/services/hulu_mapping_service.rb
```diff
@@ -0,0 +1,81 @@
...
+    series_res = Typhoeus.get(
+      "#{HULU_URL}/series?guid=#{guid}&primary_category=animation&offset=#{offset}&limit=#{limit}"
+    )
+    return unless series_res.success?
+    JSON.parse(series_res.body)['results']
+  end
+
+  # Start of syncing series and episodes on hulu with kitsu anime
+  def sync_series_and_episodes
+    limit = 10_000
```
NuckChorris: Is this basically a constant?

app/services/hulu_mapping_service.rb
```diff
@@ -0,0 +1,81 @@
...
+    hulu_anime = get_series(limit)
+    return if hulu_anime.nil?
+    hulu_anime.each do |item|
```
NuckChorris: `get_series(limit)&.each` probably a bit cleaner

app/services/hulu_mapping_service.rb
cinquemb: yeah, i can move that to the worker

app/services/hulu_mapping_service.rb
cinquemb: yeah, i can put it in the url now if you want to remove this

NuckChorris: Yeah I'd do that or at least move it to like `LIMIT = 10_000`

NuckChorris (approved):

PR #208, opened 2017-08-16
Import/viz

cinquemb:


lib/tasks/importers.rake
```diff
@@ -21,6 +21,14 @@ namespace :importers do
       MediaImporter.new.run!
     end
   end
+  namespace :viz do
+    require 'viz_import/volume_chapter_importer'
+    desc 'Import Viz Volumes and Chapters'
+    task volume_assoc: :environment do |_t|
+      puts 'Importing Viz Volumes and Chapters'
+      VolumeChapterImporter.new.run!
```
NuckChorris, line 29: You're calling `new` with no params but the `initialize` method requires one

cinquemb: Nope, that `initialize` method is for the `ImportFiles` sub class, `VolumeChapterImporter` only has the method `run!`. Had to rebase this anyways for the new mapping stuff

app/models/volume.rb
```diff
@@ -0,0 +1,27 @@
...
+#
+# Foreign Keys
+#
+#  fk_rails_4ab5f3f2e5  (manga_id => manga.id)
+#
+# rubocop:enable Metrics/LineLength
+
+class Volume < ApplicationRecord
+  has_many :chapters
+  has_many :mappings, as: 'media', dependent: :destroy
```
NuckChorris: This needs to be `as: 'item'` now

db/schema.rb
```diff
@@ -1499,6 +1500,16 @@
 
   add_index "videos", ["episode_id"], name: "index_videos_on_episode_id", using: :btree
 
+  create_table "volumes", force: :cascade do |t|
+    t.string   "title",      null: false
+    t.string   "isbn",       null: false
```
NuckChorris, line 1571: Teeeeeeechnically, Postgres has a module with an ISBN type, but reading over the docs I guess barcodes are a mess and we'll probably need a library for that when we wanna do more with it

So string is probably a good choice until we actually know more about what we're doing :+1:

lib/viz_import/volume_chapter_importer.rb
```diff
@@ -0,0 +1,78 @@
+class VolumeChapterImporter
+  class ImportFiles
+    attr_reader :data
+
+    def initialize(filename)
```
NuckChorris, line 5: This logic might be a lot cleaner if you divvy it up into smaller `VizVolume`, `VizDump`, and `VizChapter` classes

As it is, I'm having a lot of trouble following the flow

NuckChorris (changes requested):

db/migrate/20170814023144_add_volume_reference_on_chapters.rb
```diff
@@ -0,0 +1,6 @@
+class AddVolumeReferenceOnChapters < ActiveRecord::Migration
+  def change
+    add_reference :chapters, :volume, index: true,
+                                      foreign_key: true, null: true, default: nil, required: true
```
NuckChorris: Is there a reason for `null: true, default: nil, required: true`?

I'm pretty sure that's the default value for `null` and `default` and then `required` isn't even an option on this

NuckChorris (changes requested):

PR #209, opened 2017-08-17
TVDB sync

toyhammered:
Still has some refactoring and a bunch of small checks to do.


app/services/the_tvdb_service.rb
```diff
@@ -0,0 +1,95 @@
...
+  end
+
+  def weekly_import!
+    return if missing_data.empty?
+
+    missing_data['Anime'].each do |media_ids|
+      media_id = media_ids[0]
+
+      # check if the external_sites exist.
+      series_id = media_ids[1].try(:[], 'thetvdb/series')
```
NuckChorris: You might be better off just doing `next unless media_ids[1]` before these so you don't need to do that `try(:[]` stuff

app/services/the_tvdb_service.rb
```diff
@@ -0,0 +1,95 @@
...
+      season_id = media_ids[1].try(:[], 'thetvdb/season')
+
+      # will get the data from tvdb
+      # TODO: need to figure out how to skip if it returns nothing (404)
+      # TODO: will also need to add a check if episode_count and data.count are equal.
+      # will most likly break this out of 1 method.
+      get(build_episode_path(series_id, season_id)).try(:[], 'data')&.each do |tvdb_episode|
```
NuckChorris: It's possible that they're gonna force pagination on this, so you might need to do multiple requests for this.

It's possible that a sane solution might be to have one class which handles getting the TVDB IDs to do, then another class to handle using a series/season ID to get the episodes, then another to import the episodes.

That way, you keep the three pieces of logic separate: finding the Mappings, grabbing the series data, and updating the episode data

app/services/the_tvdb_service.rb
```diff
@@ -0,0 +1,95 @@
...
+                     acc[media_type] ||= {}
+                     acc[media_type][media_id] ||= {}
+                     acc[media_type][media_id][external_site] = external_id
+                   }
+  end
+
+  def api_token
+    body = { apikey: API_KEY }.to_json
+
+    request = Typhoeus::Request.new(build_url('/login'),
```
NuckChorris: I think you can use `Typhoeus.post(` and it'll directly hand you the response object

app/services/the_tvdb_service/row.rb
```diff
@@ -0,0 +1,34 @@
...
+    def update_episode!
+      episode = Episode.first_or_initialize(
+        media_id: media_id,
+        media_type: media_type,
+        number: data['absoluteNumber']
+      )
+
+      episode.season_number ||= data['airedSeason']
+      episode.relative_number ||= data['airedEpisodeNumber']
+      # was returning an empty string and not updating.
```
NuckChorris: You can use a `.blank?` check instead of `.nil?` to treat empty strings as missing

app/services/the_tvdb_service.rb
toyhammered: ahhhh, was missing a class.

app/services/the_tvdb_service.rb
toyhammered: oops, forgot to change this, will do once you review my current push and tell me anything else to fix.

app/services/the_tvdb_service.rb
```diff
@@ -0,0 +1,122 @@
...
+  API_KEY = ENV['THE_TVDB_API_KEY'].freeze
+
+  attr_accessor :media_id
+
+  def initialize
+    @token = api_token
+  end
+
+  # This will update all mappings to have a season number
+  # This should be run ONLY ONCE, unless we import data from somewhere else.
```
NuckChorris: Is it *safe* to run multiple times?

app/services/the_tvdb_service.rb
```diff
@@ -0,0 +1,122 @@
...
+      season_number = response['data'].count.zero? ? 1 : response['data'].first['airedSeason']
+
+      m = Mapping.where(
+        external_site: 'thetvdb/season',
+        media_id: media_id,
+        media_type: 'Anime'
+      ).first_or_initialize
+
+      m.external_id = season_number.to_s
+      m.save! if m.changed?
```
NuckChorris: Is `m.changed?` necessary?

app/services/the_tvdb_service.rb
```diff
@@ -0,0 +1,122 @@
+class TheTvdbService
+  BASE_URL = 'https://api.thetvdb.com'.freeze
+  API_KEY = ENV['THE_TVDB_API_KEY'].freeze
+
+  attr_accessor :media_id
+
+  def initialize
+    @token = api_token
```
NuckChorris: It's better to just have `api_token` return the same token if it's called again and then use that directly

app/services/the_tvdb_service.rb
toyhammered: wouldn't it just prevent an unnecessary save?

app/services/the_tvdb_service.rb
toyhammered: so something like this?
```
  def api_token
    return @token if @token.present?
```

app/services/the_tvdb_service.rb
toyhammered: yup, I ran it twice in a row and causes no issues. It is more for documentation because there won't be a cron-job for this method.

app/services/the_tvdb_service.rb
toyhammered: nvm, we don't need it.

app/services/the_tvdb_service.rb
NuckChorris: Pretty much, yep

app/services/the_tvdb_service.rb
NuckChorris: Alright

app/services/the_tvdb_service.rb
```diff
@@ -0,0 +1,124 @@
...
+    mapping_data['Anime'].each do |media_ids|
+      # set the media_id (this is for our database)
+      self.media_id = media_ids[0]
+      series_id = media_ids[1]['thetvdb/series'][1]
+      season_id = media_ids[1]['thetvdb/season'][1] if media_ids[1].try(:[], 'thetvdb/season')
+
+      response = get(build_episode_path(series_id, season_id))
+
+      unless response.code == 200
+        # will be deleted the mappings if their is a 404 response code
```
NuckChorris: Wait, what?

app/services/the_tvdb_service.rb
```diff
@@ -0,0 +1,124 @@
...
+  end
+
+  # This will update all mappings to have a season number
+  # This should be run ONLY ONCE, unless we import data from somewhere else.
+  def update_mapping_seasons!
+    return if mapping_data.blank?
+
+    mapping_data['Anime'].each do |media_ids|
+      # set the media_id (this is for our database)
+      self.media_id = media_ids[0]
```
NuckChorris: Why in heaven's name are you setting something on `self` from within a loop!?

app/services/the_tvdb_service.rb
```diff
@@ -0,0 +1,124 @@
...
+      series_id = media_ids[1]['thetvdb/series'][1]
+      season_id = media_ids[1]['thetvdb/season'][1] if media_ids[1].try(:[], 'thetvdb/season')
```
NuckChorris: Wouldn't it make more sense to destructure this in the parameters like `|(anime_id, mappings)|` and then use `mappings['thetvdb/series']`

app/services/the_tvdb_service.rb
NuckChorris: How about something like this?

```ruby
next if response.code == 404
raise 'TVDB Error' unless response.success?

# Code to create new mapping
```

toyhammered: truthfully, now that we changed it to merge into 1, I can do that because if they don't exist it won't create the new mapping. At the end of the day you are deleting the season and series mapping.

app/services/the_tvdb_service.rb
toyhammered: not sure I follow, but I realized the `.try` was pointless.

NuckChorris (approved):

PR #211, opened 2017-08-19
Feature: Custom Profile URLs

NuckChorris:
# Goal

Loosen up username requirements by separating the URL from the username.

# How it works

Introduces a new `slug` column representing the user's choice of URL.

Once again, stealing Facebook's behavior here.  So by default, users are at `/users/5554` (their ID number) but they can choose a Custom URL such as `nuck` and get `/users/nuck`.

I pulled in the `citext` (case-insensitive text) Postgres extension because it'll make our life easier on this.  I checked, and RDS supports it, so we're good 👌 The only downside to `citext` is that it interferes with index usage during `LIKE` queries, but we don't use `LIKE` queries for this so... who cares?

# Eventually Needed on Frontend
This should be deployable as-is, but will need the following frontend changes to actually expose it:

* Switch to `filter[slug]=nuck` instead of `filter[name]=nuck`
* Stop redirecting `/users/5554` to `/users/name` if the user doesn't have a slug
* Provide an interface to pick your slug

Once these changes are made, we can loosen the requirements for the `name` column and ship this fully

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -159,7 +159,7 @@ class User < ApplicationRecord
   has_many :one_signal_players, dependent: :destroy
   has_many :reposts, dependent: :destroy
   has_many :ip_addresses, dependent: :destroy, class_name: 'UserIpAddress'
-  validates :email, :name, :password, absence: true, if: :unregistered?
+  validates :email, :name, :password, :slug, absence: true, if: :unregistered?
   validates :email, :name, :password_digest, presence: true, if: :registered?
   validates :email, uniqueness: { case_sensitive: false },
                     if: ->(user) { user.registered? && user.email_changed? }
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -163,24 +163,23 @@ class User < ApplicationRecord
   validates :email, :name, :password_digest, presence: true, if: :registered?
   validates :email, uniqueness: { case_sensitive: false },
                     if: ->(user) { user.registered? && user.email_changed? }
-  validates :slug, uniqueness: { case_sensitive: false },
-                   format: {
-                     with: /\A[_A-Za-z0-9]+\z/,
-                     message: <<-EOF.squish
-                       can only contain letters, numbers, and underscores.
-                     EOF
-                   },
-                   if: :slug_changed?
-  validates :slug, format: {
-    with: /\A[A-Za-z0-9]/,
-    message: 'must begin with a letter or number'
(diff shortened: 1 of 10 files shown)


app/resources/user_resource.rb
```diff
@@ -1,14 +1,12 @@
...
-  attributes :name, :past_names, :about, :bio, :about_formatted, :location,
-    :waifu_or_husbando, :followers_count, :facebook_id, :following_count,
-    :life_spent_on_anime, :birthday, :gender, :comments_count, :favorites_count,
-    :likes_given_count, :reviews_count, :likes_received_count, :posts_count,
-    :ratings_count, :media_reactions_count, :pro_expires_at, :title,
-    :profile_completed, :feed_completed, :website
+  attributes :name, :past_names, :slug, :about, :location, :waifu_or_husbando, :followers_count,
+    :following_count, :life_spent_on_anime, :birthday, :gender, :comments_count, :favorites_count,
+    :likes_given_count, :reviews_count, :likes_received_count, :posts_count, :ratings_count,
+    :media_reactions_count, :pro_expires_at, :title, :profile_completed, :feed_completed, :website
```
NuckChorris, line 9: Yes, I did drop `bio` and `about_formatted`, since we stopped using those

db/migrate/20170819034722_add_slug_to_users.rb
```diff
@@ -0,0 +1,6 @@
+class AddSlugToUsers < ActiveRecord::Migration
+  def change
+    enable_extension 'citext'
+    add_column :users, :slug, :citext
```
matthewdias, line 4: should we have this be not null?

NuckChorris: Nah, we want it to be null by default in future

app/models/user.rb
```diff
@@ -313,6 +316,9 @@ def update_profile_completed!
     update_profile_completed.save!
   end
 
+  # TODO: remove once slugs are live on frontend
+  before_validation(if: :name_changed?) { self.slug = name }
```
cinquemb: Are we allowing spaces inside of slugs and other chars? Will that be ok for how we want to use this?

NuckChorris: Ah, you're right, I should add some validations on `slug` itself 🤔

toyhammered (approved):

cinquemb (approved):

PR #213, opened 2017-08-21
Add notification when users like your reactions

NuckChorris:


matthewdias (approved):
cool once build is fixed

PR #215, opened 2017-08-31
Enhancement/perf for db locking

cinquemb:
Should address:
- locking on user table related to ip_address
- locking on media due to updating rating frequencies

### DIFF
--- a/app/models/concerns/rateable.rb
+++ b/app/models/concerns/rateable.rb
@@ -29,8 +29,7 @@ module Rateable
           COALESCE(rating_frequencies->'#{rating}', '0')::integer + #{diff}
         )::text)
     EOF
-    self.class.where(id: id).update_all(update_query)
-    touch
+    UpdateRatingFrequencyWorker.perform_async(self, update_query)
   end
 
   def decrement_rating_frequency(rating)
--- a/app/models/concerns/rateable.rb
+++ b/app/models/concerns/rateable.rb
@@ -29,7 +29,8 @@ module Rateable
           COALESCE(rating_frequencies->'#{rating}', '0')::integer + #{diff}
         )::text)
     EOF
-    UpdateRatingFrequencyWorker.perform_async(self, update_query)
+    class_name = self.class.name.to_s
+    UpdateRatingFrequencyWorker.perform_async(class_name, id, update_query)
   end
 
   def decrement_rating_frequency(rating)
--- a/app/models/concerns/rateable.rb
+++ b/app/models/concerns/rateable.rb
@@ -23,14 +23,8 @@ module Rateable
 
   def update_rating_frequency(rating, diff)
     return if rating.nil?
-    update_query = <<-EOF
-      rating_frequencies = rating_frequencies
-        || hstore('#{rating}', (
-          COALESCE(rating_frequencies->'#{rating}', '0')::integer + #{diff}
-        )::text)
-    EOF
     class_name = self.class.name.to_s
-    UpdateRatingFrequencyWorker.perform_async(class_name, id, update_query)
+    UpdateRatingFrequencyWorker.perform_async(class_name, id, rating, diff)
   end
 
   def decrement_rating_frequency(rating)
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -37,7 +37,6 @@
 #  import_error                :string(255)
 #  import_from                 :string(255)
 #  import_status               :integer
-#  ip_addresses                :inet             default([]), is an Array
 #  language                    :string
 #  last_backup                 :datetime
 #  last_recommendations_update :datetime
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -46,7 +46,6 @@
 #  mal_username                :string(255)
 #  media_reactions_count       :integer          default(0), not null
 #  name                        :string(255)
-#  never_signed_in_email_sent  :boolean          default(FALSE), not null
 #  ninja_banned                :boolean          default(FALSE)
 #  password_digest             :string(255)      default(""), not null
 #  past_names                  :string           default([]), not null, is an Array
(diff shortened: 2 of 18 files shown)


app/models/user.rb
```diff
@@ -47,6 +46,7 @@
 #  mal_username                :string(255)
 #  media_reactions_count       :integer          default(0), not null
 #  name                        :string(255)
+#  never_signed_in_email_sent  :boolean          default(FALSE), not null
```
NuckChorris: hm?

app/models/user.rb
```diff
@@ -157,7 +157,7 @@ class User < ApplicationRecord
   has_many :notification_settings, dependent: :destroy
   has_many :one_signal_players, dependent: :destroy
   has_many :reposts, dependent: :destroy
-
+  has_many :user_ipaddresses, dependent: :destroy
```
NuckChorris: `has_many :ip_addresses, dependent: :destroy, class_name: 'UserIpAddress'` is probably nicer

app/models/user_ipaddress.rb
```diff
@@ -0,0 +1,27 @@
+# rubocop:disable Metrics/LineLength
+# == Schema Information
+#
+# Table name: user_ipaddresses
```
NuckChorris: Probably should be `user_ip_addresses`, since `ipaddress` is not one word

app/models/user.rb
```diff
@@ -239,15 +236,12 @@ def one_signal_player_ids
   end
 
   def add_ip(new_ip)
-    unless ip_addresses.include?(new_ip)
-      ips = [new_ip, *ip_addresses].compact.first(PAST_IPS_LIMIT)
-      update_attribute(:ip_addresses, ips)
-    end
-    ip_addresses
+    UserIpaddress.where(user: self, ip_address: new_ip).first_or_create
```
NuckChorris: This could be `ip_addresses.where(ip_address: new_ip).first_or_create` since the association will provide `user: self` automagically

app/models/user.rb
```diff
@@ -239,15 +236,12 @@ def one_signal_player_ids
...
+    UserIpaddress.where(user: self).map(&:ip_address)
   end
 
   def alts
-    User.alts_of(self)
+    UserIpaddress.where(user: self).map(&:user)
```
NuckChorris: This is a really complicated way of returning `self`, and won't actually find alt accounts for the user

Alts are accounts which share an IP address with the `self`, so you need to find all IPs for the user and then find all Users which also have that IP.  Can probably be done with a subquery, something like this:

```ruby
UserIpAddress.where(ip_address: ip_addresses.select(:ip_address)).includes(:user).map(&:user)
```

app/models/user.rb
```diff
@@ -239,15 +236,12 @@ def one_signal_player_ids
   end
 
   def add_ip(new_ip)
-    unless ip_addresses.include?(new_ip)
-      ips = [new_ip, *ip_addresses].compact.first(PAST_IPS_LIMIT)
-      update_attribute(:ip_addresses, ips)
-    end
-    ip_addresses
+    UserIpaddress.where(user: self, ip_address: new_ip).first_or_create
+    UserIpaddress.where(user: self).map(&:ip_address)
```
NuckChorris: I'm not sure this is necessary, I don't think the `add_ip` method needs to return anything

app/workers/update_rating_frequency_worker.rb
```diff
@@ -0,0 +1,10 @@
+class UpdateRatingFrequencyWorker
+  include Sidekiq::Worker
+
+  def perform(class_name, class_id, query)
+    class_ = class_name.safe_constantize
```
NuckChorris: The usual Rubyism for this is to name the variable `klass`

app/workers/update_rating_frequency_worker.rb
```diff
@@ -0,0 +1,10 @@
...
+    class_obj = class_.find(class_id)
+    class_.where(id: class_id).update_all(query)
```
NuckChorris: Is there a reason for this roundabout approach?  Couldn't you just call `class_obj.update(query)`?

app/models/concerns/rateable.rb
```diff
@@ -29,8 +29,8 @@ def update_rating_frequency(rating, diff)
           COALESCE(rating_frequencies->'#{rating}', '0')::integer + #{diff}
         )::text)
     EOF
-    self.class.where(id: id).update_all(update_query)
-    touch
+    class_name = self.class.name.to_s
+    UpdateRatingFrequencyWorker.perform_async(class_name, id, update_query)
```
NuckChorris: It might make more sense if the worker takes `rating` and `diff` instead of passing the SQL as a parameter

spec/models/library_entry_spec.rb
```diff
@@ -46,6 +46,7 @@
 # rubocop:enable Metrics/LineLength
 
 require 'rails_helper'
+require 'sidekiq/testing'
```
NuckChorris: We should require this in `spec/support/sidekiq.rb` tbh

NuckChorris (changes requested):

app/workers/update_rating_frequency_worker.rb
cinquemb: Well for some reason, running `Sidekiq::Testing.inline!`, converts the class into a hash in rspec, so a way around that was doing this

db/migrate/20170822072811_insert_existing_user_ip_address_and_remove_column_from_users.rb
```diff
@@ -0,0 +1,13 @@
+class InsertExistingUserIpAddressAndRemoveColumnFromUsers < ActiveRecord::Migration
+  disable_ddl_transaction!
+  def change
+    User.find_each do |user|
+      all_users_notifications = []
+      user.ip_addresses.each do |ip|
+        all_users_notifications << {ip_address: ip, user: user}
+      end
+      UserIpaddress.create(all_users_notifications)
```
NuckChorris: Looks like you have the old capitalization here

NuckChorris (changes requested):
Once that's fixed, feel free to merge :+1:

db/migrate/20170822072811_insert_existing_user_ip_address_and_remove_column_from_users.rb
cinquemb: Ahh good catch

PR #216, opened 2017-08-31
Create default units when guess is changed

matthewdias:
This will allow db mods to change the guess to generate default units

### DIFF
--- a/app/models/anime.rb
+++ b/app/models/anime.rb
@@ -111,7 +111,8 @@ class Anime < ApplicationRecord
   alias_method :year_changed?, :season_changed?
 
   def update_unit_count_guess(guess)
-    update(episode_count_guess: guess) unless episode_count
+    return if episode_count || (episode_count_guess && episode_count_guess < guess)
+    update(episode_count_guess: guess)
   end
 
   def self.fuzzy_find(title)
--- a/app/models/anime.rb
+++ b/app/models/anime.rb
@@ -114,7 +114,6 @@ class Anime < ApplicationRecord
     return if episode_count
     return if episode_count_guess && episode_count_guess > guess
     update(episode_count_guess: guess)
-    episodes.create_defaults(episode_count_guess) if episodes.length < episode_count_guess
   end
 
   def self.fuzzy_find(title)
@@ -137,7 +136,9 @@ class Anime < ApplicationRecord
   end
 
   after_save do
-    episodes.create_defaults(episode_count) if
-      episode_count_changed? && episodes.length != episode_count
+    %w[episode_count episode_count_guess].each do |count|
+      target = send(count)
+      episodes.create_defaults(target) if send("#{count}_changed?") && episodes.length != target
+    end
   end
 end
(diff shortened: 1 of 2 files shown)


app/models/anime.rb
```diff
@@ -137,7 +135,10 @@ def self.fuzzy_find(title)
   end
 
   after_save do
-    episodes.create_defaults(episode_count) if
-      episode_count_changed? && episodes.length != episode_count
+    if episode_count_guess_changed? && episodes.length != episode_count_guess
+      episodes.create_defaults(episode_count_guess || 0)
+    elsif episode_count_changed? && episodes.length != episode_count
+      episodes.create_defaults(episode_count || 0)
+    end
```
NuckChorris, line 142: Won't this order cause the guess to be used instead of the count if they're both set (which ideally they shouldn't be, but it's possible they will be)

app/models/anime.rb
```diff
@@ -111,10 +111,8 @@ def season_year
   alias_method :year_changed?, :season_changed?
 
   def update_unit_count_guess(guess)
-    return if episode_count
-    return if episode_count_guess && episode_count_guess > guess
+    return if episode_count || episode_count_guess != guess
```
NuckChorris: Won't `episode_count_guess != guess` always be the case?  I mean, isn't the whole point of this method to modify it?

So like, if `episode_count_guess` isn't set, it'll be `nil != guess`, and if it's already set this won't update it

matthewdias: whoops, yeah

app/models/anime.rb
matthewdias: they won't ever be both set because we nil out the guess in `before_save` if count is set.

app/models/anime.rb
```diff
@@ -111,10 +111,7 @@ def season_year
   alias_method :year_changed?, :season_changed?
 
   def update_unit_count_guess(guess)
-    return if episode_count
-    return if episode_count_guess && episode_count_guess > guess
-    update(episode_count_guess: guess)
-    episodes.create_defaults(episode_count_guess) if episodes.length < episode_count_guess
+    update(episode_count_guess: guess) unless episode_count
```
NuckChorris: Now won't it overwrite mod guesses?  The old `episode_count_guess < guess` check was preventing the number from going down automatically (could only increase automatically)

matthewdias: indeed 😞

app/models/anime.rb
```diff
@@ -111,10 +111,8 @@ def season_year
   alias_method :year_changed?, :season_changed?
 
   def update_unit_count_guess(guess)
-    return if episode_count
-    return if episode_count_guess && episode_count_guess > guess
+    return if episode_count || (episode_count_guess && episode_count_guess < guess)
```
NuckChorris: I think you got this backwards, look at the removed lines in the diff (they use `>` while this uses `<`)

NuckChorris (approved):

PR #217, opened 2017-09-02
Allow null content for post and comment

matthewdias:
So that you can make posts/comments with only uploads.

Do we want to do any extra validation here?


spec/models/comment_spec.rb
```diff
@@ -41,9 +41,27 @@
   it { should belong_to(:user) }
   it { should have_many(:replies).class_name('Comment').dependent(:destroy) }
   it { should have_many(:likes).class_name('CommentLike').dependent(:destroy) }
-  it { should validate_presence_of(:content) }
   it { should validate_length_of(:content).is_at_most(9_000) }
 
+  context 'with content or uploads' do
+    subject { build(:comment, content: nil) }
+
+    context 'with content' do
```
NuckChorris: This is gonna result in `Comment with content or uploads with content` — you might wanna drop that `with content or uploads`

PR #219, opened 2017-09-04
Tvdb Sync - weekly and daily

toyhammered:
The weekly query works. The daily query "seems" to work, but because it only returns 5 results and we don't actually have any mappings for newly aired shows I can't say if it is right.


config/sidekiq.yml
```diff
@@ -19,3 +19,7 @@ production:
     every: 12h
   UploadRemovalWorker:
     every: 12h
+  TheTvdbDailyWorker:
```
toyhammered: @NuckChorris I am not sure if this is valid actually.

app/services/the_tvdb_service.rb
```diff
@@ -2,6 +2,37 @@ class TheTvdbService
...
 
+  # daily or weekly
+  def import!(worker_type)
+    data = query_data(worker_type)
+    return if data.blank?
+
+    data['Anime'].each do |media_ids|
+      # set the media_id (this is for our database)
+      media_id = media_ids[0]
+      series_id, season_id = media_ids[1]['thetvdb'].split('/')
```
NuckChorris: Destructure the `media_ids` in the parameter list

```ruby
data['Anime'].each do |(anime_id, mappings)|
  series_id, season_id = mappings['thetvdb'].split('/')
```

app/services/the_tvdb_service.rb
```diff
@@ -2,6 +2,37 @@ class TheTvdbService
...
+      next unless anime.episode_count == tvdb_episodes_count \
+        || anime.episode_count_guess == tvdb_episodes_count
+
+      # creating/updating the episode
+      process_episode_data(response, media_id, series_id)
+    end
+  end
+
   # This will create new mappings that combine series/season together.
   def create_tvdb_mapping!
```
NuckChorris: Feel free to delete this now that it's been run

app/services/the_tvdb_service.rb
```diff
@@ -2,6 +2,37 @@ class TheTvdbService
...
+      # set the media_id (this is for our database)
+      media_id = media_ids[0]
+      series_id, season_id = media_ids[1]['thetvdb'].split('/')
+
+      response = get(build_episode_path(series_id, season_id))
+
+      next if response.code == 404
+      raise 'TVDB Error' unless response.success?
+
+      response = JSON.parse(response.body)
```
NuckChorris: The JSON parsing and error handling should probably be in the `get` method or something

app/services/the_tvdb_service.rb
```diff
@@ -2,6 +2,37 @@ class TheTvdbService
...
+
+      # update the episode count if it does not exist.
+      tvdb_episodes_count = response['data'].count
+      anime = update_anime_episode_count(media_id, tvdb_episodes_count)
+
+      # we don't want to add their data if the episode counts don't match up
+      # TODO: we should log this because it would be good to double check.
+      next unless anime.episode_count == tvdb_episodes_count \
```
NuckChorris: Listen to Hound

app/services/the_tvdb_service.rb
```diff
@@ -38,6 +69,29 @@ def mapping_data
                    end
   end
 
+  def missing_thumbnail_data
+    @mtd ||= Mapping.where(external_site: 'thetvdb')
+                    .joins('LEFT OUTER JOIN episodes ON mappings.item_id = episodes.media_id AND mappings.item_type = episodes.media_type')
```
NuckChorris: Use a heredoc to wrap the line

```ruby
.joins(<<-SQL)
  LEFT OUTER JOIN episodes
  ON mappings.item_id = episodes.media_id
  AND mappings.item_type = episodes.media_type
SQL
```

app/services/the_tvdb_service.rb
```diff
@@ -38,6 +69,29 @@ def mapping_data
...
+                    .each_with_object({}) do |(item_type, item_id, external_site, external_id), acc|
+                      acc[item_type] ||= {}
+                      acc[item_type][item_id] ||= {}
+                      acc[item_type][item_id][external_site] = external_id
+                    end
+  end
+
+  def currently_airing_data
+    @cad ||= Mapping.where(external_site: %w[thetvdb/series thetvdb/season])
+                    .joins('LEFT OUTER JOIN anime ON mappings.item_id = anime.id').merge(Anime.current)
```
NuckChorris: Move the merge down to a new line

app/services/the_tvdb_service.rb
```diff
@@ -38,6 +69,29 @@ def mapping_data
...
+                    .pluck(:item_type, :item_id, :external_site, :external_id)
+                    .each_with_object({}) do |(item_type, item_id, external_site, external_id), acc|
```
NuckChorris: You repeat this logic in multiple places, DRY it out

app/services/the_tvdb_service.rb
```diff
@@ -70,6 +124,13 @@ def get(url)
     )
   end
 
+  def query_data(worker_type)
+    case worker_type
+    when 'daily' then return currently_airing_data
+    when 'weekly' then return missing_thumbnail_data
```
NuckChorris: You don't need `return` if the `case` statement is the last thing in the method

app/services/the_tvdb_service/row.rb
```diff
@@ -0,0 +1,52 @@
...
+      return unless episode_number.present?
+
+      episode.season_number ||= data['airedSeason']
+      episode.relative_number ||= data['airedEpisodeNumber']
+      # was returning an empty string and storing that in database.
+      episode.synopsis = data['overview'].presence || nil if episode.synopsis.blank?
+      episode.thumbnail_file_name ||= thumbnail_path
+      episode.airdate ||= data['firstAired']
+      # HACK: due to some corrupt db entries having {"en_jp" => nil} which causes validation errors
+      episode.titles['en_jp'] ||= data['episodeName'] || "Episode #{episode.number}"
```
NuckChorris: Most of these would do well to be put in brief methods of their own, such as this:

```ruby
def episode_name
  data['episodeName'] || "Episode #{episode.number}"
end
```

app/services/the_tvdb_service/row.rb
```diff
@@ -0,0 +1,52 @@
...
+      # only set en_us if the tvdb has it.
+      episode.titles['en_us'] ||= data['episodeName'] if data['episodeName'].present?
+      episode.canonical_title = 'en_us' if episode.titles.key?('en_us')
+
+      # to get the imdbId we would have to call the /episodes/{episode_number} endpoint
+      # instead of the series/episdoes.
+      # TODO: implement getting imdbId once we can store it.
+
+      episode.save! if episode.changed?
```
NuckChorris: Why use `.changed?`, won't Rails just ignore `save!` if there's no changes?

Also it might be smart to break this apart: `def update_episode` to apply attributes, `def update_episode!` to apply+save (so just `update_episode.save!`)

app/services/the_tvdb_service/row.rb
```diff
@@ -0,0 +1,52 @@
...
+    def update_episode!
+      return unless episode_number.present?
+
+      episode.season_number ||= data['airedSeason']
+      episode.relative_number ||= data['airedEpisodeNumber']
+      # was returning an empty string and storing that in database.
+      episode.synopsis = data['overview'].presence || nil if episode.synopsis.blank?
+      episode.thumbnail_file_name ||= thumbnail_path
+      episode.airdate ||= data['firstAired']
+      # HACK: due to some corrupt db entries having {"en_jp" => nil} which causes validation errors
```
NuckChorris: This really doesn't explain the line below it _at all_

app/services/the_tvdb_service/row.rb
```diff
@@ -0,0 +1,52 @@
...
+    end
+
+    def update_episode!
+      return unless episode_number.present?
+
+      episode.season_number ||= data['airedSeason']
+      episode.relative_number ||= data['airedEpisodeNumber']
+      # was returning an empty string and storing that in database.
+      episode.synopsis = data['overview'].presence || nil if episode.synopsis.blank?
+      episode.thumbnail_file_name ||= thumbnail_path
```
NuckChorris: Close, but no cigar.

`episode.thumbnail ||= thumbnail_path`

app/services/the_tvdb_service/row.rb
```diff
@@ -0,0 +1,52 @@
...
+      @tvdb_series_id = tvdb_series_id
+      @media_type = 'Anime'
+    end
+
+    def update_episode!
+      return unless episode_number.present?
+
+      episode.season_number ||= data['airedSeason']
+      episode.relative_number ||= data['airedEpisodeNumber']
+      # was returning an empty string and storing that in database.
```
NuckChorris: What was?  When?

Others reading this code will have none of the context you did when writing it, you need to remember that while writing comments

app/services/the_tvdb_service/row.rb
```diff
@@ -0,0 +1,52 @@
...
+      episode.synopsis = data['overview'].presence || nil if episode.synopsis.blank?
```
NuckChorris: This is another great candidate for being broken into a method

app/services/the_tvdb_service.rb
```diff
@@ -2,6 +2,37 @@ class TheTvdbService
   BASE_URL = 'https://api.thetvdb.com'.freeze
   API_KEY = ENV['THE_TVDB_API_KEY'].freeze
 
+  # daily or weekly
+  def import!(worker_type)
```
NuckChorris: Why not pass `worker_type` to `def initialize()`?  Also you've leaked the concept of being part of a "worker" into this service, which isn't good; what happens if we change the intervals?  Instead, maybe try taking a reference to a set, like instead of `daily`, do `currently_airing` (you could name the variable `set_name` or something idfk)

If you do that, then you could have this for `query_data`:

```ruby
def query_data
  return unless %w[currently_airing missing_thumbnails].include? @set_name
  send(@set_name)
end
```

NuckChorris (changes requested):

app/services/the_tvdb_service.rb
toyhammered: I don't think it is possible, the next cannot be inside a separate method from the loop, and it has to be the first thing to execute (because response.success? checks for 2xx but we don't want to raise on a 404). Maybe yielding (but not sure how to set it up).

app/services/the_tvdb_service.rb
toyhammered: not sure how to DRY this type of thing out tbh.

NuckChorris: See that big part where you do `pluck` and then `each_with_object`, which is duplicated?  Stick it in a method.

app/services/the_tvdb_service/row.rb
toyhammered: 🤔

app/services/the_tvdb_service.rb
```diff
@@ -1,41 +1,59 @@
...
+  def initialize(set_name)
+    @set_name = set_name
+  end
 
-    mapping_data['Anime'].each do |media_ids|
-      # set the media_id (this is for our database)
-      media_id = media_ids[0]
-      series_id = media_ids[1]['thetvdb/series']
-      season_id = media_ids[1]['thetvdb/season']
+  # daily or weekly
```
NuckChorris: This comment is probably meaningless now tbh

app/services/the_tvdb_service.rb
```diff
@@ -84,24 +124,25 @@ def build_episode_path(series_id, season_number)
...
-      external_site: 'thetvdb',
-      item_id: media_id,
-      item_type: 'Anime'
-    ).first_or_initialize
+    if anime.episode_count.blank?
+      if anime.episode_count_guess.blank?
+        anime.episode_count_guess = tvdb_episodes_count
+      elsif anime.episode_count_guess < tvdb_episodes_count
+        anime.episode_count_guess = tvdb_episodes_count
+      end
```
NuckChorris: Doesn't `update_unit_count_guess` on a media do this logic?

app/services/the_tvdb_service/row.rb
```diff
@@ -0,0 +1,62 @@
...
+
+    def initialize(media_id, data, tvdb_series_id)
+      @media_id = media_id
+      @data = data
+      @episode_id = data['id']
+      @tvdb_series_id = tvdb_series_id
+      @media_type = 'Anime'
+    end
+
+    def update_episode
```
NuckChorris, line 13: I'd suggest giving this method more love, it seems to do too much by itself

app/services/the_tvdb_service.rb
toyhammered: uhhh didn't know that existed, will look into.

app/services/the_tvdb_service/row.rb
toyhammered: hmmm will see if I can think of something.

app/services/the_tvdb_service.rb
toyhammered: yup

app/workers/the_tvdb_weekly_worker.rb
```diff
@@ -0,0 +1,7 @@
+class TheTvdbWeeklyWorker
+  include Sidekiq::Worker
+
+  def perform
+    TheTvdbService.new('missing_thumbnails').import!
```
NuckChorris: Probably makes more sense to use a symbol for this

config/sidekiq.yml
NuckChorris: I just confirmed that it's fine

app/services/the_tvdb_service/row.rb
```diff
@@ -0,0 +1,87 @@
...
+      return unless episode_number.present?
+
+      %i[
```
NuckChorris: You should put a comment here explaining that all of these methods will modify the `episode` object in-place

app/services/the_tvdb_service/row.rb
```diff
@@ -0,0 +1,87 @@
...
+        season_number relative_number synopsis
+        thumbnail airdate titles canonical_title
+      ].each { |k| public_send(k) }
+
+      # to get the imdbId we would have to call the /episodes/{episode_number} endpoint
+      # instead of the series/episdoes.
+      # TODO: implement getting imdbId once we can store it.
```
NuckChorris: (1) you should be more clear in the comment that the IMDB ID you're referring to is an *episode* ID
(2) you should specify what is meant by "once we can store it" (i.e. "once we can store mappings for episodes")

app/services/the_tvdb_service.rb
```diff
@@ -1,41 +1,58 @@
...
-  def mapping_data
-    @md ||= Mapping.where(external_site: %w[thetvdb/series thetvdb/season])
-                   .pluck(:item_type, :item_id, :external_site, :external_id)
-                   .each_with_object({}) do |(item_type, item_id, external_site, external_id), acc|
-                     acc[item_type] ||= {}
-                     acc[item_type][item_id] ||= {}
-                     acc[item_type][item_id][external_site] = external_id
-                   end
+  def missing_thumbnails
+    return @mtd if @mtd
```
NuckChorris: Please don't use hngrn_ntn (hungarian notation) for variables.  Use whole words.

```ruby
return @missing_thumbnails if @missing_thumbnails
```

app/services/the_tvdb_service.rb
```diff
@@ -84,24 +123,16 @@ def build_episode_path(series_id, season_number)
...
-      external_site: 'thetvdb',
-      item_id: media_id,
-      item_type: 'Anime'
-    ).first_or_initialize
-
-    mapping.external_id = "#{series_id}/#{season_number}"
-    mapping.save!
+  def update_anime_episode_count(media_id, tvdb_episodes_count)
+    anime = Anime.find(media_id)
+    anime.update_unit_count_guess(tvdb_episodes_count)
```
NuckChorris: Is there a good reason for this to be considered a "guessed" episode count?

app/services/the_tvdb_service.rb
```diff
@@ -1,41 +1,58 @@
...
 
-      next if response.code == 404
-      raise 'TVDB Error' unless response.success?
+      response = get(build_episode_path(series_id, season_id))
+      next unless response
 
-      response = JSON.parse(response.body)
+      # update the episode count if it does not exist.
+      tvdb_episodes_count = response['data'].count
+      anime = update_anime_episode_count(anime_id, tvdb_episodes_count)
```
NuckChorris: `update_anime_episode_count` is doing more than the name implies, it's also acting as `Anime.find`, which makes this really odd to read.  Why would we call `update_anime_episode_count` with an `anime_id` and `episodes_count` and then... store the result as `anime`????

It's probably best to just do that right here instead of in a method

app/services/the_tvdb_service.rb
toyhammered: @matthewdias and I had a conversation about this. Being as I wrote that like 2 weeks ago and left myself no notes I honestly don't remember...

matthewdias: Iirc the reason was tvdb episode counts just match up with how many episode pages have been added to the DB. For example, if a show is gonna be 26 eps, but info has only been submitted for 6, we'd get 6 as a count. So if ours are nil and we imported those as actual counts we'd have a lot of wrong counts. (At least last I checked this is the case. Might wanna verify that)

toyhammered: that sounds right from what I remember.

app/services/the_tvdb_service/row.rb
```diff
@@ -0,0 +1,89 @@
...
+      # All these methods will be modifying the episode object in-place
+      %i[
+        season_number relative_number synopsis
+        thumbnail airdate titles canonical_title
+      ].each { |k| public_send(k) }
+
+      # to get the imdbId we would have to call the /episodes/{episode_number} endpoint
+      # instead of the series/episdoes.
+      # TODO: implement getting imdbId (which is an episode id)
+      # once we can store mappings for an episode.
```
NuckChorris: We can now store mappings for an episode

`Mapping.create(item: episode, external_site: 'imdb/episodes', external_id: imdbId)`

toyhammered: for `item:` are we taking the kitsu episode_id or the tvdb episode_id? I have no idea.

NuckChorris: It's an association my dude

config/sidekiq.yml
NuckChorris: Could you change these to `cron:` declarations to match #231?

[This cron generator site](http://www.cronmaker.com) may come in handy, but the weekly one will probably be like `0 0 * * SUN` and the daily one will probably be like `0 0 * * *`

PR #220, opened 2017-09-05
Enhancement/meta image dimensions

cinquemb:
We'll need to run `bundle exec rake attachments:resave`, I had some trouble with this because of the Algolia indexes on  the models now.


matthewdias (approved):
👌

app/formatters/attachment_value_formatter.rb
```diff
@@ -5,6 +5,17 @@ def self.format(value)
 
     urls = value.styles.keys.map { |style| [style, value.url(style)] }
     urls << [:original, value.url]
+    urls << [
```
matthewdias: Does this need to be applied for the other sizes (styles?) as well?

NuckChorris (approved):

PR #221, opened 2017-09-07
Enhancement/mapping and library search with algolia

cinquemb:


app/models/mapping.rb
```diff
@@ -28,6 +28,10 @@ def self.lookup(site, id)
     find_by(external_site: site, external_id: id).try(:item)
   end
 
+  def self.guess_algolia(type, query)
```
NuckChorris: Algolia is just how things work from now on, so there's no need to have an `_algolia` suffix.  Probably the best solution here is to just do something like:

```ruby
def self.guess(type, query)
  query = query[:title] if query.is_a?(Hash) # Backwards compat
  AlgoliaMediaIndex.search(query, type).first
end
```

So the old calls to `.guess()` keep working until we strip it down (or alternatively we can bring back the count matching if Algolia doesn't match perfectly)

app/indices/algolia_media_index.rb
```diff
@@ -24,4 +24,16 @@ class AlgoliaMediaIndex < BaseIndex
...
+    filter_hash = {
+      filters: filter,
+      attributesToRetrieve: %w[id kind]
+    }
+    res = index.search(
+      search_query,
+      filter_hash
+    ).deep_symbolize_keys
+    res[:hits]
+  end
```
NuckChorris: This probably belongs in the `LibrarySearchService` 🤔

app/indices/base_index.rb
```diff
@@ -44,6 +44,19 @@ def index_name=(value)
       self._index = value
     end
 
+    def search(search_query, model)
```
NuckChorris: Probably better to do `model` as a keyword parameter, then we can add more options for the `.search` method over time and not deal with refactoring calls or providing backwards compatibility in here

app/services/library_search_service.rb
```diff
@@ -19,7 +19,7 @@ def to_a
   #
   # @return [Integer] the total number of results
   def total_count
-    result_media.total_count
+    algolia_result_media.length
```
NuckChorris: I think this will give the wrong number due to pagination; Algolia should probably give a _total results count_ somewhere in the response data

app/services/library_search_service.rb
```diff
@@ -61,59 +59,23 @@ def result_entries
...
-            fields: %w[titles.* abbreviated_titles],
-            type: 'phrase_prefix',
-            query: _queries[:title].join(' '),
-            boost: 1.2
-          } }
-        ]
-      }
-    }
+  # @return [Hash<String, *>] from algolia response
+  def algolia_result_media
```
NuckChorris: Another method that probably doesn't need an `algolia_` prefix

NuckChorris (changes requested):

NuckChorris (approved):

PR #222, opened 2017-09-15
Add media reaction vote notif settings

matthewdias:
I think we'll need to rerun `20170616053125_add_notification_settings_for_users.rb` after deploying this. Edit: just run the same script but only for the new setting otherwise we'll generate duplicates of the existing ones.


cinquemb (commented):
Should we just dupe `20170616053125_add_notification_settings_for_users.rb` with just the new type we want to add into a new migration?

matthewdias:
Yeah

cinquemb (approved):

PR #224, opened 2017-09-22
mal sync volumes, tags, and comments

matthewdias:


cinquemb (approved):

NuckChorris (approved):

app/models/list_import/my_anime_list_xml/row.rb
```diff
@@ -52,11 +52,12 @@ def reconsume_count
       end
 
       def notes
-        node.css('my_comments, my_tags').map(&:content).reject(&:blank?)
-            .join(';')
+        comments = node.at_css('my_comments').content
+        tags = node.at_css('my_tags').content
+        tags.present? ? [comments, tags].join("\n=== MAL Tags ===\n") : comments
```
NuckChorris, line 57: The only thing I see is that maybe this would be better handled by compacting the array and then joining

matthewdias: Would that still insert the tag prefix if no comments exist? We need it there to detect for export.

NuckChorris: ah, you're right.  :+1: LGTM then

PR #225, opened 2017-09-22
Only import existing entries, and prefer kitsu attributes

matthewdias:
Added a new `merge` strategy to use for initial sync of mal export. This only tries to import data for entries that already exist on kitsu, and will only assign attributes that haven't already been set on kitsu entries. So, your existing kitsu ratings will remain unchanged, but if you don't have any notes, your tags and comments from mal will be added to the entry. That way they aren't deleted from mal during initial sync.

### DIFF
--- a/app/models/list_import.rb
+++ b/app/models/list_import.rb
@@ -28,7 +28,7 @@ class ListImport < ApplicationRecord
 
   belongs_to :user, required: true, touch: true
 
-  enum strategy: %i[greater obliterate]
+  enum strategy: %i[greater obliterate merge]
   enum status: %i[queued running failed completed]
   has_attached_file :input_file, s3_permissions: :private
   alias_attribute :kind, :type
@@ -43,9 +43,7 @@ class ListImport < ApplicationRecord
   def type_is_subclass
     in_namespace = type.start_with?('ListImport')
     is_descendant = type.safe_constantize <= ListImport
-    unless in_namespace && is_descendant
-      errors.add(:type, 'must be a ListImport class')
-    end
+    errors.add(:type, 'must be a ListImport class') unless in_namespace && is_descendant
   end
 
   # Apply the ListImport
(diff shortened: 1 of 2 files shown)


NuckChorris:
hmmmmm I'm not sure that's a useful thing, like I feel `greater` already does a better job of that? 🤔

matthewdias:
When the progress and status are the same, MAL data will just win out, which doesn't make a whole lot of sense when you're just trying to export your kitsu data to MAL. same reason we want to skip entries that flat out don't exist in the kitsu library. Basically when you're exporting to MAL, i don't think MAL data should be overwriting any Kitsu data at all, unless its empty on Kitsu there and we need it to not erase MAL stuff.

NuckChorris:
Hmm, but if they progressed on a series on MAL it won't sync that into Kitsu and that progress will be lost forever

matthewdias:
Isn't that the point? I thought it was supposed to be a one-way sync. If that's not the case then yeah this PR is useless.

NuckChorris:
Hmm... IDK, I think this is something we'll need to discuss

JoshFabian:
The original point was to allow users to keep their MAL updated while using Kitsu. Encouraging users to make MAL the backup, not us. Two-way sync potentially puts us into the backup position again.

NuckChorris:
It's just the initial sync.  When you enable MAL import, the current behavior is to sync your MAL library to Kitsu so that _we_ can handle merging the two libraries (because MAL is bad at it)

After that sync, it's all one way (Kitsu->MAL)

Matt is proposing in this PR that we change the behavior of this import from picking the data with higher progress numbers to picking Kitsu and only using MAL if we don't have that library entry at all

I feel like that might be a bit heavy-handed, what if we change the merge behavior for text columns so that if the "winning" (higher progress) side has it empty we'll use the other?

JoshFabian:
Oh, I see. Yeah - I kinda see both sides, but it would be annoying to have my primary source of truth (kitsu) altered unexpectedly.

NuckChorris:
Yeah but on the other hand what if it ISN'T your primary source of truth? I don't expect Kitsu-loyal users are really the target market for this

We've used the greater-progress thing for past imports because it's usually the newer data but I guess that's a problem if the numbers are the same

matthewdias:
>only using MAL if we don't have that library entry at all

Actually, it proposes we don't import if we don't have the entry, that way entries that don't exist on kitsu aren't unexpectedly added from MAL. Is this okay for export? Mal won't delete entries if they don't exist in the xml upload, right?

>what if we change the merge behavior for text columns so that if the "winning" (higher progress) side has it empty we'll use the other?

That sounds good to me

matthewdias:
Closing this as stale, lets revisit the issue next time we change imports/exports significantly

PR #228, opened 2017-09-25
updating policy/validations for uploads

cinquemb:


matthewdias (approved):

app/models/upload.rb
```diff
@@ -44,6 +44,7 @@ class Upload < ApplicationRecord
     )
   }
 
+  validates :upload_order, absence: true, if: :owner_id?
```
NuckChorris: Hmm, isn't this backwards?  Shouldn't we be requiring the order if there's an owner, not the opposite?

spec/policies/upload_policy_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  permissions :show? do
+    it('should allow users') { should permit(user, upload) }
+    it('should allow anons') { should permit(nil, upload) }
+  end
+
+  permissions :update?, :destroy? do
+    it('should not allow anons') { should_not permit(nil, upload) }
+    it('should allow for yourself') { should permit(user, upload) }
+    it('should not allow for others') { should_not permit(user, other) }
```
NuckChorris: Might wanna add a test that admins should be able to destroy them

NuckChorris (approved):

PR #231, opened 2017-10-04
Simplify sidekiq queue structure

NuckChorris:
This PR does three things:

* Simplifies to four priority-based queues
* Changes our scheduled tasks to use cron-style declarations
* Splits average rating update into separate Anime and Manga tasks for Parallelism™

# New Queues
 - `now` — run within a few seconds (metrics alert @ 30 seconds)
 - `soon` — run within a few minutes (metrics alert @ 10 minutes)
 - `later` — run within an hour (metrics alert @ 2 hours)
 - `eventually` — run within a few hours (metrics alert @ 6 hours)

# Cron Declarations
I happened to catch this [in the sidekiq-scheduler readme](https://github.com/Moove-it/sidekiq-scheduler#notes-about-running-on-multiple-hosts):

>`cron` and `at` jobs are pushed once regardless of the number of sidekiq-scheduler running instances, assumming that time deltas between hosts is less than 24 hours.
>
>`every`, `interval` and `in` jobs will be pushed once per host.

We've been using `every` for our declarations, but we have two sidekiq hosts currently and may need to add more soon.  We've probably been running these tasks twice without even realizing it, so I switched to `cron` to fix that.

It's not pretty, but I put some docs for how to read a cron declaration, and comments on each declaration to explain it.

# Separate Average Rating Jobs
I split the average rating updates into two separate scheduled tasks so that they can run In Parallel™

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -71,7 +71,6 @@ module Kitsu
     config.action_mailer.delivery_method = :smtp
     config.action_mailer.perform_deliveries = true
     config.action_mailer.raise_delivery_errors = true
-    config.action_mailer.deliver_later_queue_name = :soon
     config.action_mailer.smtp_settings = {
       address: ENV['SMTP_ADDRESS'],
       port: ENV['SMTP_PORT']&.to_i,
--- a/config/application.rb
+++ b/config/application.rb
@@ -71,6 +71,7 @@ module Kitsu
     config.action_mailer.delivery_method = :smtp
     config.action_mailer.perform_deliveries = true
     config.action_mailer.raise_delivery_errors = true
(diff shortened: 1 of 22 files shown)


config/application.rb
```diff
@@ -71,6 +71,7 @@ class Application < Rails::Application
     config.action_mailer.delivery_method = :smtp
     config.action_mailer.perform_deliveries = true
     config.action_mailer.raise_delivery_errors = true
+    config.action_mailer.deliver_later_queue_name = :soon
```
NuckChorris: 🤔 Apparently Rails only added this in 5.x, so I'll need to figure out an alternative before merge

toyhammered (approved):

PR #232, opened 2017-10-04
Split webhooks controller

NuckChorris:
`WebhooksController` was getting fat, so I broke it into separate controllers in the `Webhooks::` namespace, and then added some auth to the Stream webhook endpoints in anticipation of enabling push notifications


app/controllers/webhooks/getstream_controller.rb
```diff
@@ -0,0 +1,24 @@
...
+      render text: StreamRails.client.api_key
+    end
+
+    def notify
+      notifications = JSON.parse(request.body.read)
+
+      # Since it may be up to 100 per request, send the notifications in background to prevent
+      # timeouts and handle errors better.
+      notifications.each do |notification|
+        OneSignalNotificationWorker.perform_async(notification) unless notification['new'].empty?
```
toyhammered, line 18: might want to do `if notification['new'].present?` for clarification?

app/controllers/webhooks/youtube_controller.rb
```diff
@@ -0,0 +1,46 @@
...
+    include CustomControllerHelpers
+
+    before_action :check_linked_account_param
+
+    def verify
+      mode = params['hub.mode']
+      topic = params['hub.topic']
+      challenge = params['hub.challenge']
+
+      if !linked_account && mode == 'unsubscribe' # deleted linked-account
```
toyhammered, line 12: This whole if statement is kinda messy. Not sure if there is a way to make it nicer though.

toyhammered (commented):
looks fine to me, just some ideas of places you can clarify/make better maybe.

app/controllers/webhooks/getstream_controller.rb
NuckChorris: I just moved the existing code, I'm gonna have a different PR soon which makes it suck less

app/controllers/webhooks/youtube_controller.rb
NuckChorris: Yeah pubsubhubbub doesn't lend itself to clean code tbh

PR #233, opened 2017-10-05
Add byebug gem

toyhammered:
🙏


NuckChorris:
Could you add `pry-byebug`?  I fucking LOVE byebug, but I'd be lost without pry integration

PR #235, opened 2017-10-10
Enhancement/embeds

cinquemb:


app/models/post.rb
```diff
@@ -79,6 +79,7 @@ class Post < ApplicationRecord
   validates :media, polymorphism: { type: Media }, allow_blank: true
   # posting to a group, posting to a profile, and posting to an interest are mutually exclusive.
   validates_with ExclusivityValidator, over: %i[target_user target_group target_interest]
+  validates_with ExclusivityValidator, over: %i[uploads embed], if: :uploads || :embed
```
NuckChorris: I don't think you can do `if: :uploads || :embed`, the `||` will run when the class is defined and just return `:uploads` so it literally just becomes `if: :uploads`

app/services/embed_service.rb
```diff
@@ -2,7 +2,11 @@ class EmbedService
   EXPIRY = 12.hours
   EMBEDDERS ||= [
     NicoVideoEmbedder,
-    OpenGraphEmbedder
+    TwitterEmbedder,
+    OembedEmbedder,
+    OpenGraphEmbedder,
+    MetaContentEmbedder,
+    GeneralUrlEmbedder
```
NuckChorris: I think the order probably should be
```
OpenGraphEmbedder,
OembedEmbedder,
TwitterEmbedder,
MetaContentEmbedder,
GeneralUrlEmbedder
```

app/embedders/twitter_embedder.rb
```diff
@@ -0,0 +1,43 @@
...
+    twitter :creator
+  end
+
+  def twitter_info
+    {
+      card: card,
+      site: site,
+      creator: creator,
+      twitter: twitter,
+      player_info: player_info
```
NuckChorris: We should try to keep this structure uniform across all embed systems so that the client doesn't need to know where the data came from.  When I wrote the OpenGraphEmbedder, I didn't change much because it was already pretty much a reasonable structure, but we should try to build an abstracted JSON structure for embeds

NuckChorris (changes requested):

NuckChorris (approved):

PR #241, opened 2017-10-11
Update sentry-raven to 2.7.0

NuckChorris:


greysteil:
FYI, this upgrade broke the tests on 60% of the codebases that use [Dependabot](https://dependabot.com) (we monitor it and publish compatibility scores in our update PRs). You should be able to upgrade to 2.6.3 just fine though - that version was fully backwards compatible with the version you're currently on.

PR #252, opened 2017-10-16
Maintenance - fix rubocop issues with concerns

toyhammered:
Also just refactored bits of code.

### DIFF
--- a/app/models/concerns/retrieve_media.rb
+++ b/app/models/concerns/retrieve_media.rb
@@ -8,6 +8,6 @@ module RetrieveMedia
   end
 
   def retrieve_media
-    anime.presence || manga.presence || drama.presence || nil
+    anime.presence || manga.presence || drama.presence
   end
 end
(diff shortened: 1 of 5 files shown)


app/models/concerns/retrieve_media.rb
```diff
@@ -8,12 +8,6 @@ module RetrieveMedia
...
 
   def retrieve_media
-    if anime.present?
-      anime
-    elsif manga.present?
-      manga
-    elsif drama.present?
-      drama
-    end
+    anime.presence || manga.presence || drama.presence || nil
```
NuckChorris: `|| nil` isn't needed tbh

NuckChorris (changes requested):

app/models/concerns/retrieve_media.rb
toyhammered: wasn't sure so just put it there to be safe. Will remove though.

NuckChorris (approved):

PR #258, opened 2017-10-17
Fix more rubocop errors

toyhammered:


.pryrc
```diff
@@ -6,6 +6,6 @@ if defined?(PryByebug)
 end
 
 # Hit Enter to repeat last command
-Pry::Commands.command /^$/, 'repeat last command' do
+Pry::Commands.command (/^$/), 'repeat last command' do
```
NuckChorris: I... don't think this does what you think it does????

app/controllers/sessions_controller.rb
```diff
@@ -6,7 +6,7 @@ def new
   def create
     return render status: 403, text: 'Token missing' if params[:token].blank?
     token = Doorkeeper::AccessToken.by_token(params[:token])
-    is_admin = token.resource_owner.roles.where(name: 'admin').count > 0
+    is_admin = token.resource_owner.roles.where(name: 'admin').positive?
```
NuckChorris: It didn't tell you to change `.count > 0`, it told you to change `> 0`

Anyways, the better option is `token.resource_owner.roles.where(name: 'admin').exists?`

app/models/block.rb
```diff
@@ -29,15 +29,15 @@ class Block < ApplicationRecord
...
+    errors.add(:blocked, 'cannot be an admin') if blocked.has_role?(:admin)
+    errors.add(:blocked, 'cannot be a moderator') if blocked.title == 'Mod'
   end
 
   def self.hidden_for(user)
     return [] if user.nil?
     user = user.id if user.respond_to?(:id)
-    Block.where('user_id = ? or blocked_id = ?', *[user]*2).
-          pluck(:blocked_id, :user_id).flatten.uniq - [user]
+    Block.where('user_id = ? or blocked_id = ?', *[user] * 2)
```
NuckChorris: prolly actually best to just do `.where('user_id = ? or blocked_id = ?', user, user)` instead of my stupid cleverness

NuckChorris (changes requested):

app/controllers/sessions_controller.rb
toyhammered: oh jeez good catch, yeah i should have just removed the `> 0` but will change to what you wrote above.

app/models/block.rb
toyhammered: (I was trying to figure out the meaning behind this, thanks for explaining)

NuckChorris: Yeah it builds an array of two `user` (multiplies `[user]` times `2`) and then uses the splat operator (`*`) on it to make those into separate parameters.  It's clever, but it's also really really hard to read and I should never have done it

.pryrc
NuckChorris: I think you mean `Pry::Comments.command(/^$/, 'repeat last command') do`, to be clear

toyhammered: yeah.. I think you are right.

app/models/block.rb
toyhammered: I mean I'll give you credit the idea is A+++ but other than that...

.pryrc
toyhammered: well technically what I did is "right" (I think my stupidity beat rubocop). But anyways, changed it to be more correct.

PR #310, opened 2018-01-06
Clean up Stats system

NuckChorris:
Took a hell of a lot longer to do this than I'd hoped, since I got sick and it's taken a while to regain my momentum, but it's almost done, just gotta clean up the ActivityHistory and FavoriteYear stats, but we only need one of those for now (ActivityHistory) so I'll just leave FavoriteYear unhooked for now


app/models/concerns/stat/activity_history.rb
```diff
@@ -2,59 +2,212 @@ class Stat < ApplicationRecord
...
+
+      if last_update <= library_event.created_at.to_date
+        stats_data['days'].deep_merge!(record.generate_missing_days(last_update))
+      end
+
+      self
+    end
+
+    def preset_hash
+      Hash.new { |h, y| h[y] = Hash.new { |h2, m| h2[m] = Hash.new { |h3, d| h3[d] = {} } } }
```
toyhammered: I think this can be used instead: `Hash.new { |hash, key| hash[key] = Hash.new(&hash.default_proc) }`.

NuckChorris: don't worry about the activity history stuff, I'm gonna be refactoring it

app/models/concerns/stat/activity_history.rb
```diff
@@ -2,59 +2,49 @@ class Stat < ApplicationRecord
...
-    def recalculate!
-      activity = user.library_events.by_kind(media_column)
-                     .eager_load(media_column)
-
-      # clear stats_data
-      self.stats_data = {}
-      self.stats_data = {
-        total: activity.count,
-        activity: activity # collection of events
+    def default_data
```
toyhammered, line 5: how come you removed `week_high_score` and `total_progress`?

Also, because I don't feel like actually testing your code does it return days in this format?
```
    # Example of days stats:
    # 'days' => {
    #   '2017' => {
    #     '08' => {
    #       '20' => {
    #
    #       }
    #     }
    #   }
    # }
```

toyhammered (commented):
Honestly, looks fine to me for the most part.

app/models/concerns/stat/activity_history.rb
NuckChorris: Nope, changed the format to `{ '2017-08-20' => {} }` since it's waaaaaay easier for d3 to process

toyhammered: oh okay, and the first question?

NuckChorris: Neither one belongs in this Stat.  For now I've completely removed "best week" functionality since it's not a launch feature and the logic was really big, but I'll bring that back out when we're ready for it.  The total_progress is now in AmountConsumed, where it belongs

PR #342, opened 2018-01-31
Update because.moe url

kazzarin:
I guess it changed the day after my last PR was merged.


NuckChorris:
Prolly cause I called them out on Twitter 💃

PR #354, opened 2018-02-14
Stop sending progress events to feeds

NuckChorris:
Progress events clog up feeds without adding significant value.  Most users seem to stay to "User Activity" filter and honestly, the only good usage (see what your friends are watching) is better served by a trending system.

### DIFF
--- a/app/models/library_entry.rb
+++ b/app/models/library_entry.rb
@@ -217,8 +217,6 @@ class LibraryEntry < ApplicationRecord
     unless imported || private?
       activity.rating(rating)&.create if rating_changed? && rating.present?
       activity.status(status)&.create if status_changed?
-      # If the progress has changed, make an activity unless status is changing to completed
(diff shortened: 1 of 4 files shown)


matthewdias:
Considered having a separate feed sort of like fb used to for activity? Or are progress events causing deeper problems so that they need to be removed?

NuckChorris:
@matthewdias It's heavy in technical debt, light in usage, it has a negative effect on feed performance, and it's hard to design around.  In general, it's just a great candidate for removal.

It may come back in a different form someday, but our current approach massively sucks and I think it's better to just raze it now, throw out the technical debt, and start fresh once we have a more clear vision.  We'll be tracking LibraryEvent data in the meantime (just fixed it with this!) so if/when that happens we'll still have all the historical data!

matthewdias:
fair enough, kill it with fire!

PR #356, opened 2018-02-16
Block error text changed - for mods/admins

Uricorn:
We will not be blocked!

### DIFF
--- a/app/models/block.rb
+++ b/app/models/block.rb
@@ -29,8 +29,8 @@ class Block < ApplicationRecord
 
   validate :not_blocking_admin
   def not_blocking_admin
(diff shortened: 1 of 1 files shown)


NuckChorris (commented):
Rails prefixes the text with the field name so this would produce `blocked You cannot block admins.` which makes no sense.

Uricorn:
It does `blocked - You cannot block admins` or currently `blocked - cannot be a moderator` which doesn't make much sense.

Is it possible to remove the field name from showing without breaking anything? Because if not I think this is still an improvement.

NuckChorris:
Ah, you’re right. Okay then this isn’t too bad

PR #375, opened 2018-03-12
Rails Admin: Add more mappings

kazzarin:
Adds the new mappings from Aozora.

On another note, I noticed the TVDB mappings that came from Aozora don't appear to have the season number which might cause issues.


NuckChorris:
It's not a problem if it doesn't have a season number, we have a thing to handle that iirc

PR #380, opened 2018-03-27
Rails Admin: Exclude release schedule field

kazzarin:
I don't think this field was meant to be editable.


NuckChorris:
You're right, thanks! Hopefully we can make it editable when we leave rails admin

PR #515, opened 2019-01-03
Make abbreviated_titles required

matthewdias:
Was causing a graphql error


matthewdias:
Will this set defaults for the existing nils?

NuckChorris:
Afraid not. The migration will crash with the `change_column_null`, because it'll find null data in the table. You'll need to fix the existing nulls before adding the constraint; if this is more than a few rows to update you might need to use UpdateInBatches

NuckChorris (approved):

PR #516, opened 2019-01-06
Add MangaType to graphql

toyhammered:
I feel like MediaUnitInterface sounds weird...

### DIFF
--- a/app/graphql/types/chapter.rb
+++ b/app/graphql/types/chapter.rb
@@ -1,45 +1,14 @@
 class Types::Chapter < Types::BaseObject
+  implements Types::MediaUnitInterface
   description 'A chapter of a Manga'
 
   field :id, ID, null: false
-
-  field :titles, Types::TitlesList,
-    null: false,
-    description: 'the Titles for this chapter in various locales'
-
-  def titles
-    {
-      localized: object.titles,
-      canonical: object.canonical_title
-    }
-  end
 
   field :volume_number, Integer,
     null: false,
     description: 'The volume this chapter is related to'
 
-  field :number, Integer,
-    null: false,
-    description: 'The sequence number of this chapter'
-
-  field :synopsis, [Types::LocalizedString],
-    null: true,
-    description: 'A brief summary or description of the chapter'
-
-  def synopsis
-    # TODO: actually store localized synopsis data
-    [{ locale: 'en', text: object.synopsis }] if object.synopsis
-  end
-
-  field :length, Integer,
-    null: true,
-    description: 'The length of the chapter by pages'
-
   field :published, GraphQL::Types::ISO8601DateTime,
     null: true,
     description: 'The time when the chapter was released'
-
-  field :thumbnail, Types::Image,
-    null: true,
-    description: 'A thumbnail image for the chapter'
 end
--- a/app/graphql/types/chapter.rb
+++ b/app/graphql/types/chapter.rb
@@ -1,5 +1,5 @@
 class Types::Chapter < Types::BaseObject
-  implements Types::MediaUnitInterface
+  implements Types::MediaUnit
   description 'A chapter of a Manga'
 
   field :id, ID, null: false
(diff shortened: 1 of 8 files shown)


app/graphql/types/chapter.rb
```diff
@@ -0,0 +1,14 @@
+class Types::Chapter < Types::BaseObject
+  implements Types::MediaUnitInterface
```
NuckChorris: Despite my stupid ass using `EpisodicInterface`, GraphQL apparently doesn't recommend the suffix of `Interface` so this should probably be `Types::MediaUnit`

app/graphql/types/chapter.rb
```diff
@@ -0,0 +1,14 @@
...
+  description 'A chapter of a Manga'
+
+  field :id, ID, null: false
+
+  field :volume_number, Integer,
+    null: false,
+    description: 'The volume this chapter is related to'
```
NuckChorris, line 9: `volume_number` is unused now that we have a Volume model.  Better to expose `field :volume, Types::Volume` instead, and let API users just access `volume { number }`

app/graphql/types/chapter.rb
```diff
@@ -0,0 +1,14 @@
...
+
+  field :published, GraphQL::Types::ISO8601DateTime,
```
NuckChorris: Could you change this to `released_at`?  You can pass `method: :published` to map it to the published field in the db

app/graphql/types/manga.rb
```diff
@@ -1,3 +1,27 @@
...
+    description 'Chapters for this manga'
+    argument :number, [Integer], required: false
+  end
+
+  def chapters(number: nil)
+    if number
+      object.chapters.where(number: number)
+    else
+      AssociationLoader.for(Manga, :chapters).load(object).then(&:to_a)
+    end
```
NuckChorris, line 23: :ok_hand: This whole implementation looks right to me so far

app/graphql/types/chapter.rb
toyhammered: sure, will add this.

app/graphql/types/volume.rb
```diff
@@ -0,0 +1,37 @@
...
+    }
+  end
+
+  field :number, Integer,
+    null: false,
+    description: 'The sequence number of this unit'
+
+  field :thumbnail, Types::Image,
+    null: true,
+    description: 'A thumbnail image for the unit'
```
NuckChorris: You've got some references to `unit` throughout here but it's volume not unit

app/graphql/types/volume.rb
```diff
@@ -0,0 +1,37 @@
...
+
+  field :chapters_count, Integer,
+    null: false,
+    description: 'Total chaters per Volume'
```
NuckChorris: *chapters

app/graphql/types/volume.rb
```diff
@@ -0,0 +1,37 @@
...
+
+  field :isbn, String,
+    null: false,
+    description: 'Unique Identification for this Volume'
```
NuckChorris: 1. This is an array of ISBN numbers (so `field :isbn, [String]`)
2. The description doesn't help much (they're ISBN numbers so they actually identify a specific release of the volume, but the description doesn't explain that)

NuckChorris (changes requested):

app/graphql/types/volume.rb
toyhammered: well thank you for clarifying that I really wasn't sure how to explain it tbh.

NuckChorris: Yeah ISBN numbers are a bit weird.  We don't use them internally to identify things, but they're useful for things like finding physical copies for sale

app/graphql/types/chapter.rb
NuckChorris: Could you remove the `volume_number` field entirely?  I want to remove some of the older stuff in the switch to GraphQL

app/graphql/types/volume.rb
```diff
@@ -0,0 +1,37 @@
...
+    null: false,
+    description: 'The sequence number of this volume'
+
+  field :thumbnail, Types::Image,
+    null: true,
+    description: 'A thumbnail image for this volume'
+
+  field :chapters_count, Integer,
+    null: false,
+    description: 'Total chapters per volume'
```
NuckChorris, line 27: This makes it sound like a Manga-level attribute vs a volume-level thing.  How about this?

```suggestion
    description: 'Total chapters in this volume'
```

NuckChorris (changes requested):

toyhammered:
Already done by @matthewdias

PR #550, opened 2019-08-10
base setup for mangadex import; handling base cases

toyhammered:
# What

This feature will import all the data we scraped from mangadex and update our database.
This has the potential to update or create:
1. Manga
2. Chapters
3. Genres
4. Mappings (either for MyAnimeList if it doesn't exist, and for MangaUpdates
5. MediaStaff

# Why

Still in the process, but the general idea is to copy what we did for the tvdb scheduled job import.

# Checklist

- [ ] All files pass Rubocop

- [ ] All the tests pass


lib/mangadex_import/row.rb
```diff
@@ -0,0 +1,230 @@
...
+      MangaStaff.find_or_create_by(person_id: staff.id, manga_id: kitsu_data.id)
+    end
+
+    # I am thinking of moving everything chapte related to another class.
+    def mangadex_chapters
+      return if mangadex_data['chapters'].blank?
+
+      mangadex_data['chapters'].each do |mangadex_chapter|
+        kitsu_chapter = kitsu_data.chapters.where(number: mangadex_chapter['chapter']).first_or_initialize
+        kitsu_chapter.volume_number ||= mangadex_chapter['volume']
```
toyhammered: @NuckChorris Are we getting rid of volume_number and just going to be using the Volume association? Should I just find or create a new volume based on the chapter?

lib/mangadex_import/chapter1.rb
```diff
@@ -0,0 +1,34 @@
+class MangadexImport
+  class Chapter1
```
toyhammered, line 2: @NuckChorris I am having naming conflicts with ActiveRecord `Chapter`. Unsure how to fix

lib/mangadex_import/row.rb
```diff
@@ -0,0 +1,190 @@
...
+        category_slug = mapped_mangadex_category(category_name)
+
+        # skip if this category already exists in the association array.
+        next if kitsu_data.categories.find { |cat| cat.slug == category_slug }
+
+        category = Category.find_by(slug: category_slug)
+        # NOTE: I don't think this works the way I intended.
+        # I was thinking that we can just do save at the end and it will save
+        # the associations.
+        @kitsu_data.categories << category
```
toyhammered, line 144: This saves right away is what I realized

lib/mangadex_import.rb
```diff
@@ -0,0 +1,105 @@
...
+    'chinese (trad)' => 'zh_Hant',
+    'chinese (simp)' => 'zh_Hans',
+    'english' => 'en',
+    'filipino' => 'fil',
+    'french' => 'fr',
+    'german' => 'de',
+    'hungarian' => 'hu',
+    'indonesian' => 'id_in',
+    'italian' => 'it',
+    'japanese' => 'en_jp',
```
toyhammered, line 18: Is  this correct?

lib/mangadex_import.rb
```diff
@@ -0,0 +1,105 @@
...
+  attr_reader :file_location
+
+  LANGUAGES = {
+    'arabic' => 'ar',
+    'bulgarian' => 'bg',
+    'burmese' => 'my',
+    'catalan' => 'ca',
+    'chinese (trad)' => 'zh_Hant',
+    'chinese (simp)' => 'zh_Hans',
+    'english' => 'en',
```
toyhammered, line 11: should this also be `en_jp` because when saving with Chapter it will fail validation...

PR #631, opened 2020-03-13
Anime Mutations

matthewdias:
✏️


app/graphql/mutations/anime/create.rb
```diff
@@ -0,0 +1,13 @@
+class Mutations::Anime::Create < Mutations::BaseMutation
+  argument :input, Inputs::AnimeCreateInput,
+    required: true,
```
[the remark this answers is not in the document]
NuckChorris: I think Hound ignores branch rubocop.yml

.rubocop.yml
```diff
@@ -7,8 +7,8 @@ AllCops:
   DisplayStyleGuide: true
   StyleGuideCopsOnly: false
   TargetRubyVersion: 2.5
-Layout/AlignArguments:
-  EnforcedStyle: with_fixed_indentation
+Layout/ArgumentAlignment:
+  Enabled: false
```
NuckChorris: Is there a reason for this?  Don't we like to indent with 2 spaces for argument alignment?

app/graphql/types/anime.rb
```diff
@@ -1,4 +1,8 @@
 class Types::Anime < Types::BaseObject
   implements Types::Media
   implements Types::EpisodicInterface
+
+  field :youtube_video_id, String,
```
NuckChorris: I've been trying to generously rename fields to saner things in the GraphQL API.

What would you say to `youtube_trailer_video_id`?

(eventually I'd love to replace this with an Embed object for any video, since we have a concept of Embeds now, but that seems like a bigger, dreamier effort)

app/graphql/types/mutation_type.rb
```diff
@@ -1,4 +1,16 @@
 class Types::MutationType < Types::BaseObject
+  # rubocop:disable Layout/LineLength
+
   # Kitsu Pro
-  field :pro, Types::ProMutation, null: false
+  field :subscribe_with_stripe, mutation: Mutations::Pro::SubscribeWithStripe, description: 'Subscribe to Pro using Stripe'
+  field :unsubscribe_pro, mutation: Mutations::Pro::Unsubscribe, description: "End the user's pro subscription"
+  field :set_pro_message, mutation: Mutations::Pro::SetMessage, description: "Set the user's Hall-of-Fame message"
+  field :set_pro_discord, mutation: Mutations::Pro::SetDiscord, description: "Set the user's discord tag"
+
```
NuckChorris, line 9: Is there a reason for flattening this?  I know it's unorthodox to have a tree of mutations but I really cannot fathom having a list for the literal thousands of mutations in our app

AFAIK the only downside is unspecified execution order for multiple mutations in the same query, which I'm not convinced is a major issue?

app/graphql/types/query_type.rb
```diff
@@ -1,20 +1,48 @@
...
   end
 
+  field :manga, Types::Manga.connection_type, null: false do
+    description 'Manga in the Kitsu database'
+  end
+
+  def manga
+    ::Manga.all
+  end
+
```
NuckChorris, line 45: I can agree with splitting these (1 vs N item queries) but we might wanna outright remove `anime` and `manga` fields for now, until we have actual queries on them

app/models/concerns/titleable.rb
```diff
@@ -12,6 +12,10 @@ def canonical_title
     titles[self[:canonical_title]]
   end
 
+  def canonical_title_key
```
NuckChorris, line 15: hmmmmm maybe `canonical_title_locale` or something might be better? I'm not a huge fan of just saying `_key` and then not contextualizing what the key _is_

app/graphql/mutations/anime/create.rb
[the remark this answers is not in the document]
matthewdias: Ah. of course it does

.rubocop.yml
matthewdias: Yeah, I was trying to get it to let me have 2 args on the first line. But i can change it back and just have 1 there.

app/graphql/types/anime.rb
matthewdias: Sounds good to me 👍

app/graphql/types/mutation_type.rb
matthewdias: Yeah I think we can work around the execution order thing. But I'm not a big fan of how it looks to write the graphql for this.

eg.
```
mutation UpdateAnime($id: ID!, $input: UpdateAnimeInput!) {
  anime {
	updateAnime(id: $id, input: $input) {
      <fields>
    }
	<whatever other anime mutations>
  }
}
```

It's just a couple extra lines but it feels bad to add more boilerplate where we don't need to. And then you have to dive another level to get at the result when it comes back. And it's a slippery slope towards multiple levels of nesting. I'm of the mind we shouldnt sacrifice the api to make our file structure better. But could we maybe like just reopen the root mutation type in multiple files to keep it organized that way so we have the best of both worlds?

app/graphql/types/query_type.rb
matthewdias: I'm querying these for the list view in Wiki that db mods would have access to.

app/models/concerns/titleable.rb
matthewdias: 👍

.rubocop.yml
NuckChorris: Hmm, I think that's gotta be a different rule, we do wrapping like this everywhere in kitsu don't we?

app/graphql/types/mutation_type.rb
NuckChorris: My bigger worry is how this shows up in api docs, being a giant flat list of RPC with no namespacing

.rubocop.yml
matthewdias: Yeah I'm wondering if maybe the fact that we're using the wrong rule name on `the-future` is causing hound to complain about these. When I run rubocop locally it's all good.

app/graphql/types/mutation_type.rb
matthewdias: Ah. Yeah i mean we could standardize on having the "namespace" be at the start of the mutation name, since most things sort them alphabetically i believe. But i don't feel too strongly about it either way. With the non-flat structure we could also do `anime { update() }` instead of `anime { updateAnime() }` or `anime { animeUpdate() }`.

.rubocop.yml
NuckChorris: Ah yeah probably, I think hound only uses the master yml so you can't hide stuff in a PR by changing the rules

matthewdias: Reverted this. And also fixed the rails rules by adding the rubocop-rails gem. Forgot I had left these commented there whoops. Is it okay if we merge this once approved, even with the outstanding hound messages, given that it passes local rubocop?

matthewdias:
Oh so you are looking at the config on my branch? 🤔

PR #633, opened 2020-03-21
Check reaction_skipped as string

matthewdias:
In Rails 5 enums switch from returning as ints to strings. This was causing all entries to get excluded from the media reactions issues list.


NuckChorris (approved):
Thanks for figuring this out! I'm glad to see they've polished enums tbh

matthewdias:
Yeah 👍

PR #640, opened 2020-04-19
Anilist V2 Import!

toyhammered:
# What
Update Anilist import to integrate with V2 (graphql)

# Why
Anilist has deprecated their V1 api, thus the previous import I wrote doesn't work anymore. I will be following a very similar structure to all our other imports in utilizing the ListImport Model.

# Checklist

- [ ] All files pass Rubocop

- [x] All the tests pass


Gemfile.lock
```diff
@@ -749,4 +753,4 @@ DEPENDENCIES
   webmock
 
 BUNDLED WITH
-   2.0.2
+   2.1.4
```
toyhammered, line 756: @NuckChorris is it bad to just update the bundle version?

app/models/list_import/anilist.rb
```diff
@@ -36,77 +12,122 @@ class Anilist < ListImport
     validate :ensure_user_exists, on: :create
 
     def ensure_user_exists
-      return if input_text.blank?
-
-      @auth_token ||= get_auth_token
```
matthewdias, line 41: Don't we need to authenticate to get private entries?

app/models/list_import/anilist.rb
```diff
@@ -36,77 +12,122 @@ class Anilist < ListImport
...
+      )
     end
+
+    MEDIA_LIST_QUERY = AnilistApiWrapper::Client.parse <<-'GRAPHQL'
+      query($user_name: String) {
+        anime: MediaListCollection(userName: $user_name, type: ANIME) {
+          lists {
+            name
+            entries {
+              score
```
matthewdias: I think you need to specify `score(format: POINT_100)` to do calcs against it assuming a 100 point rating.

app/models/list_import/anilist/row.rb
```diff
@@ -1,98 +1,126 @@
...
-      rescue
-        nil
+      def formatted_date(date_node)
+        date_node.to_h.symbolize_keys.values_at(:year, :month, :day).join('-').to_date.to_s
       end
 
-      def finished_at
-        return nil if node['finished_on'].nil?
+      def anilist_key
+        "AniList/#{type}"
```
matthewdias: Can we do lower case `anilist/#{type}` to be consistent?

app/models/list_import/anilist/row.rb
```diff
@@ -1,98 +1,126 @@
...
+      end
 
-        date = node['finished_on'].split('/')
+      def mal_key
+        "MyAnimeList #{type.capitalize}"
```
matthewdias: Shouldn't this be `myanimelist/#{type}` ?

app/models/list_import/anilist/row.rb
```diff
@@ -1,98 +1,126 @@
...
       end
 
-      def progress
-        type == 'anime' ? node['episodes_watched'] : node['chapters_read']
+      # reconsume_count -> repeat
+      def reconsume_count
+        node.repeat
       end
 
-      def volumes
```
matthewdias, line 42: Shouldn't we still import this value? should be able to use `progress_volumes` right?

matthewdias (changes requested):
Looking good! just pointed out a few things.

app/models/list_import/anilist.rb
toyhammered: We might. I think we require they make their list public though for us to import (we don't want to do any auth ideally)

app/models/list_import/anilist/row.rb
toyhammered: good catch! I totally forgot to readd this.

app/models/list_import/anilist/row.rb
toyhammered: Screen Shot 2020-04-24 at 8 07 43 AM
Rails admin is what I went by.

app/models/list_import/anilist/row.rb
toyhammered: sure

app/models/list_import/anilist.rb
matthewdias: If we're willing to tradeoff losing private individual entries then 👍 . We'll want to make sure we communicate that in the UI.

app/models/list_import/anilist/row.rb
matthewdias: Yeah I think that's just a name specified to make it human-readable in the admin panel. It doesn't actually tie back to anything. https://github.com/hummingbird-me/kitsu-server/blob/the-future/config/initializers/rails_admin.rb#L195

toyhammered: ahhh....... fml... I hate you rails admin....

app/models/list_import/anilist/row.rb
toyhammered: I am not seeing a `progress_volume` anywhere. All I can see in the schema is `volumes_owned` which is updated from library entry (Volume)..... Well that makes no sense but I can shove the amount in there? 
Screen Shot 2020-04-24 at 5 32 15 PM

app/models/list_import/anilist.rb
toyhammered: @NuckChorris what is your take on this?

app/models/list_import/anilist/row.rb
matthewdias: `progress_volumes` is from anilist, and it should map to `volumes_owned` (misnomer) on our end

matthewdias (approved):
Approved pending an answer on the auth question from @NuckChorris

app/models/list_import/anilist.rb
toyhammered: Totally forgot to mention on the original one it used my personal keys so no way in hell did private ever work 🤷

NuckChorris: Yeah all our imports work this way, it's something I'd like to solve in future but it's not a huge priority imo

app/models/list_import/anilist/row.rb
```diff
@@ -1,98 +1,139 @@
...
+        other_mapping = Mapping.lookup(mal_key, mal_id) ||
+                        Mapping.guess(type.classify.safe_constantize, media_info)
+
+        if other_mapping.present?
+          Mapping.create(
+            item: other_mapping,
+            external_site: anilist_key,
+            external_id: media.id
+          )
+        end
```
NuckChorris, line 22: Seems iffy to record the result of Mapping.guess imo

Could you make this only created for idMal mappings?

app/models/list_import/anilist.rb
```diff
@@ -36,77 +12,122 @@ class Anilist < ListImport
...
-          # status[0] is always going to be either
-          # completed, on-hold, etc... just skip over that part
-          status[1].each do |media|
-            row = Row.new(media, type)
+        send("#{type}_list").each do |media|
+          row = Row.new(media, type)
 
-            yield row.media, row.data
-          end
+          yield row.media_mapping, row.data
```
NuckChorris: Shouldn't this be row.media?

NuckChorris (changes requested):
Few minor notes but otherwise looks great

app/models/list_import/anilist.rb
toyhammered: I actually changed to media_mapping thinking it was more understandable (and they have a media field that I reference by #media).

toyhammered: I have changed back to be row.media though.

app/models/list_import/anilist/row.rb
toyhammered: done

NuckChorris (approved):

PR #645, opened 2020-05-01
Update FB graph api

NuckChorris:
Swap Graph API v2.11 for the latest (v6.0)

Turns out no changes are necessary to the code, since we use so little of their API lol


toyhammered (approved):
holy cow we were outdated....

PR #647, opened 2020-05-08
GraphQL - query - categories through media

toyhammered:
# What

Add Categories to media queries.

# Checklist

- [x] All files pass Rubocop

- [ ] All the tests pass


app/graphql/types/category.rb
```diff
@@ -0,0 +1,50 @@
...
+    description: 'The amount of media that include this category.'
+
+  field :nsfw, Boolean,
+    null: false,
+    description: 'Whether the category is Not-Safe-for-Work.',
+    method: :nsfw?
+
+  field :image, Types::Image,
+    null: true,
+    description: 'An image of the category.'
```
matthewdias: Lets take this out since its not used

app/graphql/types/category.rb
```diff
@@ -0,0 +1,50 @@
...
+
+  field :description, String,
+    null: true,
+    description: 'A description of the category.'
+
+  field :slug, String,
+    null: false,
+    description: 'The URL-friendly identifier of this Category.'
+
+  field :anidb_id, Integer,
```
matthewdias: Can we make this type `ID` instead? Just so you know it doesnt indicate a primary key concept so totally kosher to have multiple `ID` types.

matthewdias (changes requested):

app/graphql/types/category.rb
toyhammered: oh, totally didn't know that! Thank you!

app/graphql/types/category.rb
toyhammered: should we take it out or deprecate it?

wopian: Mark it as deprecated in the JSON:API docs and don't include it here.

app/graphql/types/category.rb
```diff
@@ -0,0 +1,46 @@
...
+
+  field :description, String,
+    null: true,
+    description: 'A description of the category.'
+
+  field :slug, String,
+    null: false,
+    description: 'The URL-friendly identifier of this Category.'
+
+  field :anidb_id, ID,
```
NuckChorris: Could you remove this for now? I'd like to switch it to Mappings

app/graphql/types/category.rb
```diff
@@ -0,0 +1,46 @@
...
+    null: true,
+    description: 'Kitsu category mapping to Anidb category.'
+
+  field :total_media_count, Integer,
```
NuckChorris: I think it would be better if we have a media connection on this and expose the count in there. For now, maybe leave it off?

app/graphql/types/category.rb
```diff
@@ -0,0 +1,46 @@
...
+  field :nsfw, Boolean,
+    null: false,
+    description: 'Whether the category is Not-Safe-for-Work.',
+    method: :nsfw?
+
+  field :parent, Types::Category,
+    null: true,
+    description: 'The parent category. Each category can have one parent.'
+
+  field :child_count, Integer,
```
NuckChorris: As above, we should push these counts onto connection objects. For this one you might be able to customize the category connection type already and ditch the counter cache!

NuckChorris (changes requested):
Solid start, just a few tweaks mostly related to counter fields

app/graphql/types/category.rb
toyhammered: for child count we are only going 1 level down correct? We wouldn't want to count the children of children.

app/graphql/types/category.rb
```diff
@@ -0,0 +1,34 @@
...
+
+  field :id, ID, null: false
+
+  field :title, String,
+    null: false,
+    description: 'The name of the category.'
+
+  field :description, String,
+    null: true,
+    description: 'A description of the category.'
```
NuckChorris: Can you change these to localized strings?  For now you can hardcode `{ en: title }` like we do in [Types::Media#synopsis](https://github.com/hummingbird-me/kitsu-server/blob/nuck/nolt-sso/app/graphql/types/media.rb#L33-L36) but it's likely that we'll want to localize them on the server instead of on the client

NuckChorris (changes requested):

NuckChorris (approved):
LGTM

PR #650, opened 2020-05-12
Update Profile (user) and Account (user) to include additional fields.

toyhammered:
# What
Update the Profile and Account to have additional Fields. For the Profile, add additional relations including followers, followings, library_entries, and pinned post.

# Why
This is basic functionality that needs to exist. Currently have no filtering based on status which we most likely want to add. Also, this is quite slow to load all library_entries for a user based on a media_type....

# Checklist

- [ ] All files pass Rubocop

- [ ] All the tests pass


app/graphql/types/library_entry/status.rb
```diff
@@ -1,7 +1,7 @@
 class Types::LibraryEntry::Status < Types::BaseEnum
-  value 'current', ''
-  value 'planned', ''
-  value 'completed', ''
-  value 'on_hold', ''
-  value 'dropped', ''
+  value 'current', 'Media that you currently read/watch.'
```
wopian, line 2: `that you are currently reading/watching.`

wopian: Although can probably strip `that` from all of these since it's a bit redundant.

toyhammered: yeah, you are right, thank you!

toyhammered:
Going to be making a bunch of smaller PRs

PR #651, opened 2020-05-13
Custom Field - batch_load_field

toyhammered:
# What
Create a new field type called `batch_load_field` to handle automatically creating the `AssociationLoader` method as to not pollute our Types with tons of `defs`

# Why

Cut down on duplicate code. This will automatically create the method that the `field` will call to grab all the association records. It will also check if the method exists (it shouldn't because then you wouldn't be using this method).

# Checklist

- [ ] All files pass Rubocop

- [ ] All the tests pass

### DIFF
--- a/app/graphql/types/base_object.rb
+++ b/app/graphql/types/base_object.rb
@@ -1,2 +1,11 @@
 class Types::BaseObject < GraphQL::Schema::Object
+  def self.batch_load_field(name, *args, association_name: nil, **kwargs, &block)
+    if instance_methods(false).exclude?(name)
+      define_method name do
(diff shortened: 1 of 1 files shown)


app/graphql/types/base_object.rb
```diff
@@ -1,2 +1,11 @@
 class Types::BaseObject < GraphQL::Schema::Object
+  def self.batch_load_field(name, *args, association_name: nil, **kwargs, &block)
+    if instance_methods(false).exclude?(name)
```
toyhammered, line 3: Just a note, there is a good chance we don't even need this check.

NuckChorris:
I don't think this is the right call for now, since there's multiple ways to batch a load

PR #660, opened 2020-05-23
GraphQL - Add Posts and Comments to User

toyhammered:
# What

Add ability to query post/comments and pinned post from a user.

# Why

# Checklist

- [ ] All files pass Rubocop

- [ ] All the tests pass

### DIFF
--- a/app/graphql/types/comment.rb
+++ b/app/graphql/types/comment.rb
@@ -1,0 +1,43 @@
+class Types::Comment < Types::BaseObject
+  description 'A comment on a post'
+
+  field :id, ID, null: false
+
+  field :post, Types::Post,
+    null: false,
+    description: 'The post that this comment is attached to.'
+
+  field :profile, Types::Profile,
+    null: false,
+    description: 'The owner of this post',
+    method: :user
+
+  field :content, String,
+    null: false,
+    description: 'Unmodified content.'
+
+  field :content_formatted, String,
+    null: false,
+    description: 'Html formatted content.'
+
+  field :parent, Types::Comment,
+    null: true,
+    description: 'The parent comment if this comment was a reply to another.'
+  #
+  # field :likes, Types::CommentLike.connection_type,
+  #   null: false,
+  #   description: ''
+
+  # NOTE: bit trickier because of different foreign_key
+  # field :replies, Types::Comment.connection_type,
+  #   null: false,
+  #   description: ''
+
+  def likes
+    AssociationLoader.for(object.class, :likes).load(object)
+  end
+
+  def replies
+    AssociationLoader.for(object.class, :replies).load(:object)
+  end
+end
--- a/app/graphql/types/comment.rb
+++ b/app/graphql/types/comment.rb
@@ -24,16 +24,18 @@ class Types::Comment < Types::BaseObject
     null: true,
     description: 'The parent comment if this comment was a reply to another.'
 
-  # field :likes, Types::CommentLike.connection_type,
-  #   null: false,
-  #   description: ''
+  field :likes, Types::Profile.connection_type,
+    null: false,
+    description: 'Users who liked this comment.'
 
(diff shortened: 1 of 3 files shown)


app/graphql/types/comment.rb
```diff
@@ -0,0 +1,44 @@
...
+  field :id, ID, null: false
+
+  field :post, Types::Post,
+    null: false,
+    description: 'The post that this comment is attached to.'
+
+  field :profile, Types::Profile,
+    null: false,
+    description: 'The owner of this post',
+    method: :user
```
NuckChorris: You don't necessarily want the field to be named just `profile`

You might instead consider something descriptive like `creator`

app/graphql/types/comment.rb
```diff
@@ -0,0 +1,44 @@
...
+    null: false,
+    description: 'Html formatted content.'
+
+  field :parent, Types::Comment,
+    null: true,
+    description: 'The parent comment if this comment was a reply to another.'
+
+  field :likes, Types::Profile.connection_type,
+    null: false,
+    description: 'Users who liked this comment.'
```
NuckChorris, line 29: I think this is reasonable to hide the Like object, if we need more advanced shit like multiple types of "likes" we can add it later.  Easier to add than remove!

app/graphql/types/comment.rb
```diff
@@ -0,0 +1,44 @@
...
+
+  field :replies, Types::Comment.connection_type,
+    null: false,
+    description: 'All replies to a specific comment.'
+
+  def likes
+    AssociationLoader.for(object.class, :likes).load(object).then do |likes|
+      likes.map(&:user)
```
NuckChorris: This is an N+1 problem, you need to use a loader to get the User

app/graphql/types/post.rb
```diff
@@ -0,0 +1,63 @@
+class Types::Post < Types::BaseObject
+  description 'A post that is visible to your followers and globally in the news-feed.'
+
+  field :id, ID, null: false
+
+  field :profile, Types::Profile,
```
NuckChorris: As above, I think you should be more descriptive here

app/graphql/types/post.rb
```diff
@@ -0,0 +1,63 @@
...
+
+  field :media, Types::Media,
+    null: true,
+    description: 'The media tagged in this post.'
+
+  field :spoiler, Boolean,
+    null: false,
+    description: 'If this post spoils the tagged media.'
+
+  field :nsfw, Boolean,
```
NuckChorris: I think graphQL generally puts `is` prefixes on these fields like `isSpoiler` or `isNsfw`

app/graphql/types/post.rb
```diff
@@ -0,0 +1,63 @@
...
+
+  def likes
+    AssociationLoader.for(object.class, :post_likes).load(object).then do |likes|
+      likes.map(&:user)
+    end
+  end
+
+  def follows
+    AssociationLoader.for(object.class, :post_follows).load(object).then do |follows|
+      follows.map(&:user)
```
NuckChorris: Here and a few lines up where you do `likes.map` are both N+1 too!  AssociationLoader only loads `post->likes/post_follows`, you need to layer another AssociationLoader to get `post->likes->user`

NuckChorris (changes requested):

app/graphql/types/comment.rb
toyhammered: Ok, I'll go through all the places I use a profile and change them to make more sense per each context.

toyhammered: Debating between `creator` and `owner` 🤔

wopian: `author`?

NuckChorris (approved):

wopian (approved):

PR #662, opened 2020-05-23
GraphQL - Base Connection and Base Edge

toyhammered:
# What

Create a base_connection class to automatically add total_count. Also create a base_edge class.

- [ ] All files pass Rubocop

- [ ] All the tests pass

### DIFF
--- a/app/graphql/types/base_object.rb
+++ b/app/graphql/types/base_object.rb
@@ -1,2 +1,3 @@
 class Types::BaseObject < GraphQL::Schema::Object
+  connection_type_class(Types::BaseConnection)
 end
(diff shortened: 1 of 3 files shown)


app/graphql/types/base_object.rb
```diff
@@ -1,2 +1,3 @@
 class Types::BaseObject < GraphQL::Schema::Object
+  connection_type_class(Types::BaseConnection)
```
NuckChorris, line 2: Have you tested that this actually works?

NuckChorris (approved):
this seems good as long as it actually works, but I recommend testing it.  `connection_type_class` might not work the way you think it does

app/graphql/types/base_object.rb
toyhammered: yeah I just retested. It allows this type of query now which is what I was aiming for. Anything that is a `connection_type` has access to this.
```
query {
  anime{
    totalCount
    nodes {
    
    }
  }
}
```

PR #663, opened 2020-05-23
GraphQL - user media reactions + votes

toyhammered:
# What

Add media reactions to a user. This will also need to add media reactions to a library entry.

# Checklist
Must Merge https://github.com/hummingbird-me/kitsu-server/pull/662 first

- [ ] All files pass Rubocop

- [ ] All the tests pass


app/graphql/types/media_reaction_vote.rb
```diff
@@ -0,0 +1,14 @@
+class Types::MediaReactionVote < Types::BaseObject
```
NuckChorris: Should this match the approach on Likes, where you just have a list of users?

NuckChorris (changes requested):

app/graphql/types/media_reaction_vote.rb
toyhammered: yeah it should, I'll fix it up.

app/graphql/types/media_reaction_vote.rb
```diff
@@ -0,0 +1,14 @@
...
+
+  field :author, Types::Profile,
+    null: false,
+    description: 'Profile who upvoted this media reaction.',
+    method: :user
+
+  field :media_reaction, Types::MediaReaction,
+    null: false,
+    description: 'Reaction that was upvoted.'
+end
```
NuckChorris: Is this type used anywhere or can it be deleted?

app/graphql/types/library_entry.rb
```diff
@@ -15,9 +15,9 @@ class Types::LibraryEntry < Types::BaseObject
     null: false,
     description: ''
 
-  # field :media_reaction, Types::MediaReaction,
-  #   null: true,
-  #   description: 'Reaction for this specific media'
+  field :media_reaction, Types::MediaReaction,
```
NuckChorris: Thinking maybe this should just be `:reaction` to cut unnecessary context

NuckChorris (changes requested):

app/graphql/types/media_reaction_vote.rb
toyhammered: good catch totally forgot to delete.

app/graphql/types/library_entry.rb
toyhammered: yes.

NuckChorris (approved):

PR #667, opened 2020-05-27
GraphQL - Add stats for a user (profile)

toyhammered:
# What
Add the anime/manga_amount_consumed and anime/manga_category_breakdown to a user.

# Checklist

- [ ] All files pass Rubocop

- [ ] All the tests pass

### DIFF
--- a/app/graphql/types/profile_stats/amount_consumed.rb
+++ b/app/graphql/types/profile_stats/amount_consumed.rb
@@ -1,9 +1,9 @@
 class Types::ProfileStats::AmountConsumed < Types::BaseObject
-  description ''
+  description 'Generic Amount Consumed based on Media'
 
   field :time, Integer,
     null: false,
-    description: ''
+    description: 'Total time spent in minutes.'
 
   def time
     object['time']
@@ -11,7 +11,7 @@ class Types::ProfileStats::AmountConsumed < Types::BaseObject
 
   field :media, Integer,
     null: false,
-    description: ''
+    description: 'Total amount of media.'
 
   def media
     object['media']
@@ -19,7 +19,7 @@ class Types::ProfileStats::AmountConsumed < Types::BaseObject
 
   field :units, Integer,
     null: false,
-    description: ''
+    description: 'Total progress per media including reconsuming.'
 
   def units
     object['units']
@@ -27,7 +27,7 @@ class Types::ProfileStats::AmountConsumed < Types::BaseObject
 
   field :completed, Integer,
     null: false,
-    description: ''
+    description: 'Total media completed atleast once.'
 
   def completed
     object['completed']
--- a/app/graphql/types/profile_stats/amount_consumed.rb
+++ b/app/graphql/types/profile_stats/amount_consumed.rb
@@ -1,0 +1,35 @@
+class Types::ProfileStats::AmountConsumed < Types::BaseObject
+  description ''
+
+  field :time, Integer,
+    null: false,
+    description: ''
+
+  def time
+    object['time']
+  end
+
+  field :media, Integer,
+    null: false,
+    description: ''
+
+  def media
+    object['media']
+  end
+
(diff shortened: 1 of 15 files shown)


app/graphql/types/stat.rb
```diff
@@ -0,0 +1,19 @@
...
+  description 'The basic stat information.'
+
+  field :id, ID, null: false
+
+  field :profile, Types::Profile,
+    null: false,
+    description: 'The profile related to the user for this stat.',
+    method: :user
+
+  field :data, GraphQL::Types::JSON,
```
NuckChorris: Big JSON blobs are bad GraphQL practice, so I think we should create separate types defining the schema for each stat type and implementing a Stat interface like we do for Media

app/graphql/types/stat_type.rb
```diff
@@ -0,0 +1,50 @@
+class Types::StatType < Types::BaseObject
```
NuckChorris: I don't think this is a good name, what if you renamed this to ProfileStats to be more clear? 🤔

NuckChorris (changes requested):

app/graphql/types/stat.rb
toyhammered: fineeeeeeeeeee. lol

app/graphql/types/profile_stats/anime_amount_consumed.rb
```diff
@@ -0,0 +1,8 @@
+class Types::ProfileStats::AnimeAmountConsumed < Types::BaseObject
+  implements Types::ProfileStats::ProfileStatInterface
+
+  field :data, Types::ProfileStats::AmountConsumed,
+    null: false,
+    description: 'The breakdown of this specific stat.',
+    method: :stats_data
```
NuckChorris: Is there a reason for this, versus using an interface like `implements Types::ProfileStats::AmountConsumed`?

app/graphql/types/profile_stats/amount_consumed.rb
```diff
@@ -0,0 +1,35 @@
+class Types::ProfileStats::AmountConsumed < Types::BaseObject
+  description 'Generic Amount Consumed based on Media'
+
+  field :time, Integer,
+    null: false,
+    description: 'Total time spent in minutes.'
+
+  def time
+    object['time']
+  end
```
NuckChorris: This should probably only be on the AnimeAmountConsumed

app/graphql/types/profile_stats/amount_consumed.rb
```diff
@@ -0,0 +1,35 @@
...
+    null: false,
+    description: 'Total amount of media.'
+
+  def media
+    object['media']
+  end
+
+  field :units, Integer,
+    null: false,
+    description: 'Total progress per media including reconsuming.'
```
NuckChorris: "per media" seems confusing.  How about "total progress of library including reconsuming"?

app/graphql/types/profile_stats/category_breakdown.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  def total
+    object['total']
+  end
+
+  field :categories, Types::Map,
+    null: false,
+    description:
+      <<~DESCRIPTION.squish
+        A Dictionary of category_id -> count for all categories
```
NuckChorris: Call it a Map since that's the type name, we should be consistent ;)

NuckChorris (changes requested):

app/graphql/types/profile_stats/anime_amount_consumed.rb
toyhammered: 🤔 no, interface is most def the right way to go.

NuckChorris (approved):

wopian (approved):

PR #674, opened 2020-06-06
GraphQL - add library_entries to user profile

toyhammered:
# What

Add ability to query library entries through user profile and filter based on anime/manga.

# Checklist

- [ ] All files pass Rubocop

- [ ] All the tests pass


app/graphql/types/library_entry.rb
```diff
@@ -0,0 +1,68 @@
...
+    null: false,
+    description: ''
+
+  # field :media_reaction, Types::MediaReaction,
+  #   null: true,
+  #   description: 'Reaction for this specific media'
+
+  field :progress, Integer,
+    null: false,
+    description: ''
```
NuckChorris:
```suggestion
    description: 'The number of episodes/chapters this user has watched/read'
```

app/graphql/types/library_entry.rb
```diff
@@ -0,0 +1,68 @@
...
+    null: true,
+    description: 'When the profile started this media.'
+
+  field :finished_at, GraphQL::Types::ISO8601DateTime,
+    null: true,
+    description: 'When the profile finished this media.'
+
+  field :progressed_at, GraphQL::Types::ISO8601DateTime,
+    null: true,
+    description: 'unsure'
```
NuckChorris:
```suggestion
    description: 'When the user last watched an episode or read a chapter of this media'
```

app/graphql/types/library_entry.rb
```diff
@@ -0,0 +1,68 @@
...
+    null: true,
+    description: 'How much you enjoyed this media (lower meaning not liking).'
+
+  field :started_at, GraphQL::Types::ISO8601DateTime,
+    null: true,
+    description: 'When the profile started this media.'
+
+  field :finished_at, GraphQL::Types::ISO8601DateTime,
+    null: true,
+    description: 'When the profile finished this media.'
```
NuckChorris: I really don't think we need to bend this lingo from "user" to "profile"

It sounds unnatural, and while it's true that it's linked to a Profile object in GraphQL, I think people will understand if we refer to them as a "user" in docs

app/graphql/types/library_entry_status.rb
```diff
@@ -0,0 +1,7 @@
+class Types::LibraryEntryStatus < Types::BaseEnum
+  value 'current', 'Media you currently read/watch.'
+  value 'planned', 'Media you plan to read/watch at some point.'
+  value 'completed', 'Media you have completed.'
+  value 'on_hold', 'Media you have started but decided to stop.'
+  value 'dropped', 'Media you will not complete.'
```
NuckChorris: We should pick either second or third person in docs and stick to it.

So like, in third person:
```suggestion
  value 'current', 'The user is currently reading or watching this media'
  value 'planned', 'The user plans to read or watch this media in future'
  value 'completed', 'The user completed this media'
  value 'on_hold', 'The user started but paused reading or watching this media'
  value 'dropped', 'The user started but chose not to finish this media'
```

app/graphql/types/media_association.rb
```diff
@@ -0,0 +1,14 @@
+class Types::MediaAssociation < Types::BaseObject
```
NuckChorris: Could you elaborate on why you feel MediaAssociation, MangaList, and AnimeList should exist, instead of going directly from Profile->Library->LibraryEntry?

This is kinda the GQL I was thinking would be good:

```gql
enum MediaType {
  ANIME,
  MANGA
}

type Library {
  all: LibraryEntriesConnection!,
  current: LibraryEntriesConnection!,
  planned: LibraryEntriesConnection!,
  onHold: LibraryEntriesConnection!,
  dropped: LibraryEntriesConnection!
}

type Profile {
  library(type: MediaType): Library!
}
```

NuckChorris (changes requested):
Also I think a shorter term for `LibraryEntryList` would be... `Library`

app/graphql/types/library_entry_status.rb
toyhammered: Ok, Also just realized I should upcase all values `CURRENT` just to be consistent for how other enums are.

app/graphql/types/media_association.rb
toyhammered: I had a discussion at some point with @matthewdias and that approach you did above was 1/2. I decided to go with having `anime` and `manga` types (sub types?) because I wasn't sure at the time how to pass parameters around, and it read nicer than having to alias all the queries.

wopian: I would prefer `Profile->Library->LibraryEntry` too.

NuckChorris (approved):

PR #713, opened 2020-06-19
Update connections to use AssociationLoader scope

toyhammered:
# What
AssociationLoader only has `scope` as a public method. `load` should not be used directly.

# Checklist

- [ ] All files pass Rubocop

- [ ] All the tests pass


wopian (commented):
Lgtm

PR #716, opened 2020-06-21
Override auto generation of return types in mutations

toyhammered:
# What

Override the return type for mutations to prevent conflicts if following standard CRUD pattern. This will also automatically supply an errors field.

# Why

If we had `Mutations::Anime::Create` it would automatically generate a payload named `CreatePayload`. If we had `Mutations::LibraryEntry::Create` it would have a name conflict and raise an error. This change will now use for any CRUD operation, it will generate a name like `AnimeCreatePayload` or `LibraryEntryPayload`.

One thing that I would like to do is be able to override the Error Type of field that is getting generated automatically (similar how you can supply a `graphql_name` but something like an `error_type`. This is possible, just not sure if ideal.

# Checklist

- [ ] All files pass Rubocop

- [ ] All the tests pass


app/graphql/concerns/custom_payload_type.rb
```diff
@@ -0,0 +1,35 @@
...
+    # NOTE: Hook Override - https://github.com/rmosolgo/graphql-ruby/issues/2737
+    def generate_payload_type
+      resolver_name = generate_resolver_name
+      resolver_fields = fields
+
+      Class.new(object_class) do
+        graphql_name("#{resolver_name}Payload")
+        description("Autogenerated return type of #{resolver_name}")
+
+        field :errors, [Errors::UserError],
```
toyhammered: I think I just need to check if the resolver_fields includes an errors field and only create if nothing there. This will allow for overrides per mutation level and they can use different error types.

toyhammered:
@NuckChorris I think I am going to remove the Error field. From what I am reading, it isn't good practice to have something like `{ data { createLibraryEntry { libraryEntry {} error{}` but rather have it like `{ data {} error{} }`

app/graphql/concerns/custom_payload_type.rb
```diff
@@ -0,0 +1,37 @@
...
+      resolver_fields = fields
+
+      Class.new(object_class) do
+        graphql_name("#{resolver_name}Payload")
+        description("Autogenerated return type of #{resolver_name}")
+
+        if resolver_fields.keys.exclude?('errors')
+          field :errors, [Errors::UserError],
+            null: true,
+            description: 'Graphql Errors'
```
NuckChorris: We can't be sure it'll only return a `UserError`.  It's entirely possible that they'll receive a server error or something else.

This should almost certainly return `[Errors::ErrorInterface]` instead.

NuckChorris (changes requested):
The CustomPayloadType stuff seems logical, but I'm not sure what's going on in the error handling code, I'd like to discuss that before I give it a ✅

app/graphql/concerns/custom_payload_type.rb
toyhammered: I didn't realize we can straight up return an `ErrorInterface` 😆.

toyhammered:
> The CustomPayloadType stuff seems logical, but I'm not sure what's going on in the error handling code, I'd like to discuss that before I give it a ✅

I'll add some comments, but essentially you can define a field inside your mutation `field :errors, Types::YourCustomError` and make it return whatever type you want. This will only create a default error type if it does not exist.

NuckChorris (approved):

PR #728, opened 2020-07-05
GraphQL - Add Mappings Query

toyhammered:
# What

Add the ability to query mappings.

# Checklist

- [ ] All files pass Rubocop

- [ ] All the tests pass


app/graphql/types/union/mapping_item.rb
```diff
@@ -0,0 +1,5 @@
+class Types::Union::MappingItem < Types::Union::Base
+  description 'Objects which are Mappable'
+
+  possible_types Types::Anime, Types::Manga
```
NuckChorris: Quick spitup from our database:

| Type | Count |
|----|---:|
| Anime | 66937 |
| Category | 132 |
| Character | 105669 |
| Episode | 792 |
| Manga | 68084 |
| Person | 35053 |
| Producer | 1326 |

app/graphql/types/query_type.rb
```diff
@@ -143,4 +143,12 @@ def find_category_by_id(id:)
...
     ::Category.find_by(slug: slug)
   end
+
+  field :mapping, Types::Mapping.connection_type, null: false do
+    description 'All Mappings in the Kitsu database'
+  end
+
+  def mapping
+    ::Mapping.all
+  end
```
NuckChorris: I don't think this is necessary or useful, how about we just expose `Mapping.lookup` like this

```gql
type Query {
  lookupMapping(externalSite: String!, externalId: ID!): MappingItem
}
```

NuckChorris (changes requested):

app/graphql/types/union/mapping_item.rb
toyhammered: Thank you, I will add this as a concern `has_many :mappings, as: 'item', dependent: :destroy` and start adding to the Types that are missing (I know for certain character doesn't respond to `character.mappings`

app/graphql/types/query_type.rb
toyhammered: Ok, I'll also like to add one for just gettin all based on an `externalSite` which I think can be useful. Will also make that `externalSite` an Enum so the user knows what options they have.

app/graphql/types/union/mapping_item.rb
toyhammered: @NuckChorris can you also run a query to get me all `external_sites` we have?

NuckChorris: | external_site | count |
|---|--:|
| anidb | 10215 |
| anilist | 9236 |
| anilist/anime | 10850 |
| anilist/manga | 14138 |
| animenewsnetwork | 119 |
| aozora | 12182 |
| hulu | 910 |
| imdb/episodes | 792 |
| kitsu/genres | 51 |
| mangaupdates | 2606 |
| myanimelist/anime | 15607 |
| myanimelist/character | 105669 |
| myanimelist/genres/anime | 40 |
| myanimelist/genres/manga | 41 |
| myanimelist/manga | 50726 |
| myanimelist/people | 35017 |
| myanimelist/person | 36 |
| myanimelist/producer | 1326 |
| mydramalist | 2 |
| thetvdb | 2965 |
| thetvdb/season | 621 |
| thetvdb/series | 2157 |
| trakt | 2687 |

app/graphql/types/query_type.rb
wopian: `findMapping` to remain consistent with the API terms

app/graphql/types/union/mapping_item.rb
wopian: @NuckChorris what resources are getting mapped by `anilist`, `myanimelist` and `thetvdb` without suffixes?

toyhammered: anilist was mainly from zorro I think. thetvdb is from our background job (will verify). Nothing for myanimelist?

app/graphql/types/query_type.rb
toyhammered: > `findMapping` to remain consistent with the API terms

I was thinking that but it isn't really finding a mapping, it is finding the mapping item.

app/graphql/types/enum/mapping_external_site.rb
```diff
@@ -0,0 +1,17 @@
...
+  value 'THE_TVDB_SERIES', value: 'thetvdb/series'
+  value 'THE_TVDB_SEASON', value: 'thetvdb/season'
+  value 'ANI_DB', value: 'anidb'
+  value 'ANIME_NEWS_NETWORK', value: 'animenewsnetwork'
+  value 'MANGA_UPDATES', value: 'mangaupdates'
+  value 'HULU', 'hulu'
+  value 'IMDB_EPISODES', value: 'imdb/episodes'
+  value 'AOZORA', value: 'aozora'
+  value 'TRAKT', value: 'trakt'
+  value 'MY_DRAMA_LIST', value: 'mydramalist'
```
wopian: Why are all these brand names being split apart? MyDramaList is _MyDramaList_ not _My Drama List_. Anime News Network is the only service that isn't one word. 

Although I'd prefer to keep Anime News Network as `ANIMENEWSNETWORK` and the enum format be `siteName_specificResource` (all caps)

app/graphql/types/query_type.rb
wopian: `findMapping` to remain consistent with the rest of the GraphQL API?

app/graphql/types/enum/mapping_external_site.rb
toyhammered: I can keep them as one word, it just seemed weird when in all caps.

NuckChorris: It is rather weird but I think wopian has the right idea to use an underscore as the separator

NuckChorris (approved):

NuckChorris:
I'm rerunning tests before you merge, since they didn't pass. Looks like probably a fluke

app/graphql/types/union/mapping_item.rb
NuckChorris: Anilist is garbage data, there's no bare MAL in the list (you might be seeing mydramalist)

Dunno about TVDB

wopian: I wrote `myanimelist` for `anilist` for some reason - then checked the list again and didn't see `anilist` in my message so added `anilist` "again". Oops

PR #730, opened 2020-07-06
banned users cannot follow other users anymore

toyhammered:
# What

If a user is banned, they are not allowed to follow other users anymore.

# Why

They shouldn't have the capabilities to interact with anyone anymore.

# Checklist

- [ ] All files pass Rubocop

- [ ] All the tests pass

### DIFF
--- a/app/policies/follow_policy.rb
+++ b/app/policies/follow_policy.rb
@@ -13,6 +13,7 @@ class FollowPolicy < ApplicationPolicy
   # Override to user follower instead of user
   def is_owner? # rubocop:disable Style/PredicateName
     return false unless user && record.follower_id == user.id
+    return false if user.has_role?(:banned)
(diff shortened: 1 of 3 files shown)


app/policies/follow_policy.rb
```diff
@@ -13,6 +13,7 @@ def visible_attributes(all)
   # Override to user follower instead of user
   def is_owner? # rubocop:disable Style/PredicateName
     return false unless user && record.follower_id == user.id
+    return false if user.has_role?(:banned)
```
NuckChorris: The user still owns their own follows even if they're banned.

This should probably be up in `def create?` as `is_owner? && !user.has_role?(:banned)`

NuckChorris (changes requested):

NuckChorris (approved):

PR #735, opened 2020-07-12
GraphQL - update to use new interpreter

toyhammered:
This will eventually become the default, so might as well change now while it is easy.
https://graphql-ruby.org/queries/interpreter

# Checklist

- [ ] All files pass Rubocop

- [ ] All the tests pass


NuckChorris (approved):
Smart!

PR #736, opened 2020-07-13
GraphQL - Library Entry Mutations with Error Handling

toyhammered:
# What

Add mutations for LibraryEntry.

# Why

This also includes architecting handling errors as data. This is the generic flow that most simple models can follow (one-to-one). 

# Checklist

- [ ] All files pass Rubocop

- [ ] All the tests pass

### DIFF
--- a/app/graphql/kitsu_schema.rb
+++ b/app/graphql/kitsu_schema.rb
@@ -1,4 +1,4 @@
-require 'custom_errors'
+require_dependency 'custom_errors'
 
 class KitsuSchema < GraphQL::Schema
   use GraphQL::Execution::Interpreter
--- a/app/graphql/kitsu_schema.rb
+++ b/app/graphql/kitsu_schema.rb
@@ -1,3 +1,5 @@
+require 'custom_errors'
+
 class KitsuSchema < GraphQL::Schema
   use GraphQL::Execution::Interpreter
   use GraphQL::Analysis::AST
--- a/app/graphql/mutations/library_entry/create.rb
+++ b/app/graphql/mutations/library_entry/create.rb
@@ -1,6 +1,6 @@
 class Mutations::LibraryEntry::Create < Mutations::Base
   argument :input,
-    Inputs::LibraryEntry::Create,
+    Types::Input::LibraryEntry::Create,
     required: true,
     description: 'Create a Library Entry',
     as: :library_entry
--- a/app/graphql/mutations/library_entry/create.rb
+++ b/app/graphql/mutations/library_entry/create.rb
@@ -1,0 +1,29 @@
+class Mutations::LibraryEntry::Create < Mutations::Base
+  argument :input,
+    Inputs::LibraryEntry::Create,
+    required: true,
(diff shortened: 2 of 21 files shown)


app/graphql/kitsu_schema.rb
```diff
@@ -1,3 +1,5 @@
+require 'custom_errors'
```
NuckChorris: Generally it’s better to use `require_dependency` so Rails can reload this

app/graphql/mutations/library_entry/create.rb
```diff
@@ -0,0 +1,29 @@
...
+    description: 'Create a Library Entry',
+    as: :library_entry
+
+  field :library_entry, Types::LibraryEntry, null: true
+
+  def load_library_entry(value)
+    LibraryEntry.new(value.to_h)
+  end
+
+  def authorized?(library_entry:)
```
NuckChorris, line 14: Where is this called? if it’s automatically called, how is library_entry: passed in?  In Ruby 3, they will be separating hash parameters from named parameters, so if you use a named parameter here you need to make sure that it’s a named parameter (not a hash!) wherever it’s called

NuckChorris (changes requested):
Mostly good, some small questions here and there

app/graphql/kitsu_schema.rb
toyhammered: Didn't know about that, thank you.

app/graphql/mutations/library_entry/create.rb
toyhammered: So the argument `input` is aliased using `as: :library_entry`. The `#load_library_entry` will grab the record if we override it, and pass that through as `library_entry: <loaded item>`. 
From there graphql has hooks for a `#ready?` (this one is before `#load_library_entry` funny enough) and `#authorized?` (after `#load_library_entry`). For `#authorized?` it injects the argument directly into as `library_entry:` due to us aliasing it. (if we didn't alias, it would be `input:` then). After that it will send the record directly the `#resolve` exactly the same way it did to authorized.

I did leave this comment in one of the mutations, unsure where it should actually be because all of them use this:
```
  # NOTE: https://github.com/rmosolgo/graphql-ruby/issues/1733
  # NOTE: https://www.rubydoc.info/github/rmosolgo/graphql-ruby/GraphQL/Schema/Resolver#argument-class_method
  # the argument loads: ... is not required, it is called every time.
```

Sidenote: I'll need to ready about what Ruby 3 is doing with named parameters vs hash parameters.

toyhammered: This shouldn't cause an issue when we go to Ruby 3

toyhammered:
Having input name conflicts that I need to fix.... 
```
Multiple definitions for `Create`. Previously found Inputs::Anime::Create (Class), then found Inputs::LibraryEntry::Create (Class) at Mutation.createLibraryEntry.input
```

NuckChorris (approved):

PR #749, opened 2020-07-27
GraphQL - add basic search for anime and manga titles

toyhammered:
# What

Add the basic ability to search for anime/manga through Algolia.

# Checklist

- [ ] All files pass Rubocop

- [ ] All the tests pass

### DIFF
--- a/app/graphql/types/query_type.rb
+++ b/app/graphql/types/query_type.rb
@@ -34,7 +34,8 @@ class Types::QueryType < Types::BaseObject
   end
 
   def search_anime_by_title(title:)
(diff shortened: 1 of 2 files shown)


NuckChorris (changes requested):
This needs to apply auth scopes similar to what `services/algolia_key_service.rb` does

wopian:
Could searching users be added to this? That would just leave groups left (which haven't been added yet anyway)

toyhammered:
I'll make a new PR for users I think?

wopian (approved):

PR #750, opened 2020-07-27
fix(algolia): isolate filters of different types for groups

wopian:
# Why

ID and Privacy are different types and can't be mixed in the same OR group. This isolates them into separate groups. This was causing the query for logged in users to error and return no results.

# Output

- `privacy:open OR privacy:restricted`
- `(privacy:open OR privacy:restricted) AND NOT nsfw:true`
- `(id = 1 OR id = 2 OR id = 3) OR (privacy:open OR privacy:restricted)`
- `((id = 1 OR id = 2 OR id = 3) OR (privacy:open OR privacy:restricted)) AND NOT nsfw:true`


wopian:
~~This is wrong. Users with no hidden groups would see `() OR (privacy...)`. I just overlooked that while editing the file and heaving off to sleep.~~

NuckChorris (approved):

PR #748, opened 2020-07-25
GraphQL - Anime Mutations

toyhammered:
# Checklist

- [ ] All files pass Rubocop
- [ ] Any complex logic is commented
- [ ] Any new systems have thorough documentation
- [ ] Any user-facing changes are behind a feature flag (or: explain why they can't be)

- [ ] All the tests pass
- [ ] Tests have been added to cover the new code

PR #751, opened 2020-07-28
GraphQL - Manga Mutations

toyhammered:
# Checklist

- [ ] All files pass Rubocop

- [ ] All the tests pass

### DIFF
--- a/app/graphql/mutations/manga/create.rb
+++ b/app/graphql/mutations/manga/create.rb
@@ -1,0 +1,29 @@
+class Mutations::Manga::Create < Mutations::Base
+  argument :input,
+    Types::Input::Manga::Create,
+    required: true,
+    description: 'Create Manga',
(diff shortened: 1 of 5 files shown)


toyhammered:
merge #748 first!

app/graphql/mutations/manga/create.rb
```diff
@@ -0,0 +1,29 @@
...
+    as: :manga
+
+  field :manga, Types::Manga, null: true
+
+  def load_manga(value)
+    ::Manga.new(value.to_h)
+  end
+
+  def authorized?(manga:)
+    super(manga, :create?)
```
NuckChorris, line 15: Actually does this work at all?

I don't think this would work

NuckChorris (approved):
After discussion, seems this will work through Magic™

PR #752, opened 2020-07-30
fix(concerns): allow en to pass English title validation

wopian:
# What

`en`, `en_us`, `en_jp`, `en_kr` pass validation

Removed `ends_with('en')` as that will never be true. ``EN` is not a **ISO 3166** country code and would not be an English title if a country existed.

# Why

`en` currently does not pass validation, preventing it being used as the canonical key for Korean and Chinese media without a Romanised title, where they instead need to add a region-specific English title (e.g `en_us`) as canonical.

# Checklist

- [x] All files pass Rubocop

- [x] All the tests pass

### DIFF
--- a/app/models/concerns/titleable.rb
+++ b/app/models/concerns/titleable.rb
@@ -19,7 +19,7 @@ module Titleable
   private
 
   def has_english_title?
-    titles.keys.any? { |k| k.end_with?('_en') || k.start_with?('en_') }
+    titles.keys.any? { |k| k.start_with?('en') }
   end
 
   def has_english_title


app/models/concerns/titleable.rb
```diff
@@ -19,7 +19,7 @@ def canonical_title_key
   private
 
   def has_english_title?
-    titles.keys.any? { |k| k.end_with?('_en') || k.start_with?('en_') }
+    titles.keys.any? { |k| k.start_with?('en') }
```
toyhammered, line 22: Can we confirm in our database that we don't have any `_en`?

toyhammered (changes requested):
👎

app/models/concerns/titleable.rb
wopian: Why would we?

wopian: ### Anime

Found 0 entries on page 837, 16740 checked
Total 0 entries

### Manga

Found 0 entries on page 2778, 55560 checked
Total 0 entries

NuckChorris (approved):

toyhammered (approved):
fine... 👎

PR #759, opened 2020-08-01
GraphQL - add account query

toyhammered:
# Checklist

- [ ] All files pass Rubocop

- [ ] All the tests pass

### DIFF
--- a/app/graphql/types/query_type.rb
+++ b/app/graphql/types/query_type.rb
@@ -1,9 +1,9 @@
 class Types::QueryType < Types::BaseObject
-  field :my_account, Types::Account, null: true do
-    description 'Your Kitsu account details. You must supply an Authorization token in header.'
+  field :current_account, Types::Account, null: true do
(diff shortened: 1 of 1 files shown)


app/graphql/types/query_type.rb
```diff
@@ -1,4 +1,12 @@
 class Types::QueryType < Types::BaseObject
+  field :my_account, Types::Account, null: true do
```
wopian: `current_account`? `my_account` feels kinda odd when you're doing auth for other users. Same with description.

toyhammered: `me` lololol

NuckChorris: I agree with wopian, `current_account` is probably a better phrasing

toyhammered: ok

NuckChorris (approved):

wopian (approved):

PR #768, opened 2020-08-01
GraphQL - add media streams

toyhammered:


toyhammered:
Need to add a new regions field and then also have a dubs/subs field (we will keep the video sub_language and dub_language, but deprecate) and then just add subs/dubs to them, and these will just be an array.

timw4mail (approved):

app/graphql/types/anime.rb
```diff
@@ -11,4 +11,12 @@ class Types::Anime < Types::BaseObject
     null: true,
     method: :youtube_video_id,
     description: 'Video id for a trailer on YouTube'
+
+  field :streaming_links, Types::StreamingLink.connection_type,
+    null: false,
+    description: ''
```
NuckChorris: Could you add a description for this?

app/graphql/types/video.rb
```diff
@@ -0,0 +1,34 @@
+class Types::Video < Types::BaseObject
+  implements Types::Interface::Streamable
+
+  description ''
```
NuckChorris: you're missing a bunch of descriptions

NuckChorris (changes requested):

app/graphql/types/video.rb
toyhammered: I shoulda converted to draft instead of WIP, sorry.

app/graphql/types/video.rb
```diff
@@ -0,0 +1,32 @@
...
+    null: false,
+    description: 'The language this is subbed in',
+    deprecation_reason: 'Please use subs',
+    method: :sub_lang
+
+  field :dub_language, String,
+    null: false,
+    description: 'The language this is dubbed in.',
+    deprecation_reason: 'Please use dubs',
+    method: :dub_lang
```
NuckChorris: Please don't add new fields that are deprecated from day one, just leave em off

NuckChorris (changes requested):

NuckChorris (approved):

PR #774, opened 2020-08-05
GraphQL - add media character to character

toyhammered:


app/graphql/types/character.rb
```diff
@@ -27,4 +27,12 @@ def names
   field :image, Types::Image,
     null: true,
     description: 'An image of the character'
+
+  field :media, Types::MediaCharacter.connection_type,
+    null: true,
+    description: 'Additional media this character showed up in'
```
wopian: Media this character appears in

wopian (changes requested):

PR #776, opened 2020-08-07
GraphQL - add library event to library entry

toyhammered:


app/graphql/types/library_entry.rb
```diff
@@ -62,4 +62,18 @@ class Types::LibraryEntry < Types::BaseObject
   field :progressed_at, GraphQL::Types::ISO8601DateTime,
     null: true,
     description: 'When the user last watched an episode or read a chapter of this media.'
+
+  field :library_events, Types::LibraryEvent.connection_type, null: true do
```
NuckChorris: Could you rename to just events?

NuckChorris (changes requested):

app/graphql/types/library_entry.rb
toyhammered: yup

PR #783, opened 2020-08-09
Rename Synopsis -> Description and update to jsonb column.

toyhammered:
# What

Rename Synopsis to Description. Update existing and new Description to jsonb

- What bugs did you fix?
- What features did you add?
- Did you add any dependencies? -->

# Why

We wanted to normalize all the data cross multiple models to use the same field.

# Checklist

- [ ] All files pass Rubocop

- [ ] All the tests pass


wopian (approved):

app/graphql/types/person.rb
```diff
@@ -24,19 +26,13 @@ def names
     }
   end
 
+  localized_field :description,
+    description: 'A brief summary or description of the character.'
```
NuckChorris:
```suggestion
    description: 'A brief biography or description of the person.'
```

app/chewy/media_index.rb
```diff
@@ -89,7 +91,9 @@ def get_streamers(type, ids)
       # Titles and freeform text
       translatable_field :titles
       field :abbreviated_titles, type: 'string'
-      field :synopsis, type: 'string', analyzer: 'english'
+      field :description do
+        field :en, type: 'string', analyzer: 'english'
+      end
```
NuckChorris: I think this will break the search functionality in the API 🤔

app/models/concerns/description_sanitation.rb
```diff
@@ -0,0 +1,11 @@
+module DescriptionSanitation
+  extend ActiveSupport::Concern
+
+  included do
+    before_save :sanitize_description
+  end
+
+  def sanitize_description
+    description['en'] = Sanitize.fragment(description['en'], Sanitize::Config::RESTRICTED)
```
NuckChorris: This should be set up to sanitize all the fields like so:

```suggestion
    description.transform_values { |desc| Sanitize.fragment(desc, Sanitize::Config::RESTRICTED) }
```

config/initializers/rails_admin.rb
```diff
@@ -67,7 +67,7 @@ def edit_user_link
     field :abbreviated_titles, :serialized do
       html_attributes rows: '6', cols: '70'
     end
-    fields :canonical_title, :synopsis, :slug, :subtype, :poster_image, :cover_image,
+    fields :canonical_title, :description, :slug, :subtype, :poster_image, :cover_image,
```
NuckChorris: This will probably break Rails Admin, it needs to be set up as `:serialized` like `titles` is

NuckChorris (changes requested):
Not bad! Definitely a few things need addressing but overall, it's mostly right!

One thing I want to mention is that `spec/fixtures/my_anime_list/**/*.json` files... just don't seem to be used anywhere? You wanna try deleting and see if any specs fail?

app/chewy/media_index.rb
toyhammered: how come?

config/initializers/rails_admin.rb
toyhammered: ahhh yeah, you are right.

NuckChorris (approved):

PR #785, opened 2020-08-10
GraphQL - add volume and chapters

toyhammered:
# What

# Why

# Checklist

- [ ] All files pass Rubocop

- [ ] All the tests pass

### DIFF
--- a/app/graphql/types/manga.rb
+++ b/app/graphql/types/manga.rb
@@ -16,4 +16,12 @@ class Types::Manga < Types::BaseObject
   field :volume_count, Integer,
     null: true,
     description: 'The number of volumes in this manga.'
+
+  field :chapters, Types::Chapter.connection_type,
+    null: true,
+    description: 'The chapters of the manga.'
+
+  def chapters
+    AssociationLoader.for(object.class, :chapters).scope(object)
+  end
 end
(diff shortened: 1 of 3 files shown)


app/graphql/types/manga.rb
```diff
@@ -16,4 +16,12 @@ class Types::Manga < Types::BaseObject
   field :volume_count, Integer,
     null: true,
     description: 'The number of volumes in this manga.'
+
+  field :chapters, Types::Chapter.connection_type,
+    null: true,
+    description: 'The chapters of the manga.'
```
wopian:
```suggestion
    description: 'The chapters in the manga.'
```

app/graphql/types/volume.rb
```diff
@@ -0,0 +1,41 @@
...
+  def titles
+    {
+      localized: object.titles,
+      canonical: object.canonical_title
+    }
+  end
+
+  field :number, Integer,
+    null: false,
+    description: 'The number of pages in this chapter.'
```
wopian: Is this the chapter count or page count?

wopian (changes requested):

app/graphql/types/volume.rb
toyhammered: actually that is just wrong, that is the volume number.

wopian (approved):

PR #791, opened 2020-08-15
Generate blurhashes for images

NuckChorris:
# What

Generate a blurhash for images! [Blurhash](https://blurha.sh) is a way of encoding a pretty blurred version of an image in a tiny string to display while you're loading the full thing!

I've hooked it up with MiniMagick (an ImageMagick wrapper) to load and prescale the image data, and linked it up with Paperclip-Meta to store it in the existing meta column.

# Why

Because it's pretty and we have a _lot_ of images that could benefit from this!

I chose MiniMagick over RMagick because it doesn't use a huge (and frankly super buggy) native extension, instead shelling out to ImageMagick or GraphicsMagick.

I'll also mention: I haven't added any tests for this because I'm intending for this to be a temporary solution until we have [Clerk](https://github.com/hummingbird-me/clerk) handling files.

### DIFF
--- a/app/graphql/types/image.rb
+++ b/app/graphql/types/image.rb
@@ -3,12 +3,6 @@ class Types::Image < Types::BaseObject
     null: false,
     description: 'The original image'
 
-  field :views, [Types::ImageView],
-    null: false,
-    description: 'The various generated views of this image' do
-      argument :names, [String], required: false
-    end
-
   def original
     {
       name: 'original',
@@ -17,6 +11,12 @@ class Types::Image < Types::BaseObject
       height: object.height(:original)
     }
   end
+
+  field :views, [Types::ImageView],
+    null: false,
+    description: 'The various generated views of this image' do
+      argument :names, [String], required: false
+    end
 
   def views(names: object.styles.keys)
     styles = object.styles.keys
@@ -29,4 +29,8 @@ class Types::Image < Types::BaseObject
       }
     end
   end
+
+  field :blurhash, String,
+    null: true,
+    description: 'A blurhash-encoded version of this image'
 end
--- a/app/graphql/types/image.rb
+++ b/app/graphql/types/image.rb
@@ -12,11 +12,10 @@ class Types::Image < Types::BaseObject
     }
   end
 
-  field :views, [Types::ImageView],
-    null: false,
-    description: 'The various generated views of this image' do
-      argument :names, [String], required: false
-    end
+  field :views, [Types::ImageView], null: false do
+    description 'The various generated views of this image'
+    argument :names, [String], required: false
+  end
 
   def views(names: object.styles.keys)
     styles = object.styles.keys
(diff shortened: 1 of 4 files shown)


app/graphql/types/image.rb
```diff
@@ -18,6 +12,12 @@ def original
     }
   end
 
+  field :views, [Types::ImageView],
+    null: false,
+    description: 'The various generated views of this image' do
+      argument :names, [String], required: false
+    end
```
toyhammered: Let's keep it consistent to what we did for other code.

```suggestion
  field :views, [Types::ImageView], null: false do
    description 'The various generated views of this image'
    argument :names, [String], required: false
  end
```

app/graphql/types/image.rb
```diff
@@ -29,4 +29,8 @@ def views(names: object.styles.keys)
       }
     end
   end
+
+  field :blurhash, String,
+    null: true,
+    description: 'A blurhash-encoded version of this image'
```
toyhammered:
```suggestion
    description: 'A blurhash-encoded version of this image.'
```

config/initializers/paperclip.rb
```diff
@@ -30,6 +30,43 @@ def nil?
   end
 end
 
+module PaperclipBlurhash
+  def blurhash
+    return unless instance.respond_to?(:"#{name}_meta") && instance_read(:meta)
+
+    if (meta = meta_decode(instance_read(:meta)))
```
toyhammered, line 37: This _kinda_ bugs me 😆 . I have always hated inline assignments inside of `if` statements.

lib/circuit_breaker.rb
```diff
@@ -0,0 +1,44 @@
...
+# reset.  You should always have code that retries and resets the breaker.
+class CircuitBreaker
+  REDIS_KEY = 'circuit-breakers:tripped'.freeze
+  class Tripped < StandardError; end
+
+  def initialize(name, rate:, per:)
+    @name = name
+    @rate = rate
+    @per = per
+    @strait = Strait.new(name) do
```
toyhammered: 😍

toyhammered (changes requested):
Just some minor changes, but this is awesome! Super stoked to see this in production!

app/graphql/types/image.rb
NuckChorris: Bitch no periods

config/initializers/paperclip.rb
NuckChorris: Me too, but it's what they use in the code I pasted so 🤷

toyhammered: Ok, at-least I ain't crazy for thinking that.

app/graphql/types/image.rb
toyhammered: Fine, we going no periods for any single line periods?

app/graphql/types/image.rb
toyhammered: Do I just put another comment here because this GLARINGLY LARGE SUGGEST TEXT seems to have been ignored :P

toyhammered (changes requested):

app/graphql/types/image.rb
NuckChorris: I s2g I committed this wtf

toyhammered (approved):
dancing_tohka

wopian (approved):

PR #792, opened 2020-08-16
Improve Rails Admin experience

NuckChorris:
# What

Create three new inputs for Rails Admin:

### LocalizedString
A set of single-line strings for locales

### LocalizedText
A set of multi-line strings for locales

### StringList
A list of single-line strings

# Why

Mods are currently forced to know YAML and there's no good error messages, which is stupid.  This simplifies the syntax and makes it so mods don't need to really know YAML


toyhammered:
This is a misleading title. You can't possible make rails admin better...

toyhammered (approved):

PR #800, opened 2020-08-22
GraphQL - add library events query_type

toyhammered:
# What

# Why

# Checklist

- [ ] All files pass Rubocop

- [ ] All the tests pass


NuckChorris (commented):
I'm not sure we want to have a global LibraryEvents query, honestly.

What would this be used for? It's not even sorted in any useful way 😅

PR #801, opened 2020-08-22
Add units to GraphQL

NuckChorris:
# What

We're missing an interface for Units in GraphQL and the corresponding associations on LibraryEntry which power the quick update


toyhammered (approved):
looks_good

PR #832, opened 2020-09-10
GraphQL - Fix Override for graphql_name.

toyhammered:
# What

This fixes and prepends the proper Type when overriding `graphql_name`. It also allows for the setter method to work properly.

# Why

namespacing in graphql is 🤮 and this fixes the issue.

# Checklist

- [ ] All files pass Rubocop

- [ ] All the tests pass

### DIFF
--- a/app/graphql/types/input/base.rb
+++ b/app/graphql/types/input/base.rb
@@ -4,14 +4,16 @@ class Types::Input::Base < GraphQL::Schema::InputObject
   # NOTE: https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/schema/input_object.rb#L136
   # These need to be unique, so I am forcible doing that.
   # I am thou, and thou art I, but you will always be an input!
-  def self.graphql_name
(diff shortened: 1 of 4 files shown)


app/graphql/types/input/base.rb
```diff
@@ -4,14 +4,16 @@ class Types::Input::Base < GraphQL::Schema::InputObject
...
+    return super(new_name) if new_name.present?
+
     split_name = name.split('::')[2..-1]
 
     if CRUD_OPERATION.exclude?(split_name.last.downcase)
       "#{super}Input"
     else
       # Types::Input::Anime::Create -> AnimeCreate
-      "#{split_name.join}Input"
+      "#{super(split_name.join)}Input"
```
toyhammered: @NuckChorris I was unsure if we needed to do an update for this because it technically overrides the initial name completely. I can change this back if this is not correct.

NuckChorris (approved):
Lgtm!

PR #847, opened 2020-09-19
GraphQL - add ability to get random media

toyhammered:
# What
Add ability to get random media, separated by sfw/nsfw

# Why
We obviously need this...

# Checklist

- [ ] All files pass Rubocop

- [ ] All the tests pass

### DIFF
--- a/app/graphql/types/library.rb
+++ b/app/graphql/types/library.rb
@@ -1,6 +1,19 @@
 class Types::Library < Types::BaseObject
   description 'The user library filterable by media_type and status'
 
+  field :random_media, Types::Interface::Media, null: true do
+    description 'Random anime or manga from your library'
+
+    argument :media_type, Types::Enum::MediaType, required: true
+    argument :status, [Types::Enum::LibraryEntryStatus], required: true
+  end
+
+  def random_media(media_type:, status:)
+    library_entries(media_type: media_type, status: status).then do |library_entries|
+      library_entries.order(Arel.sql('RANDOM()')).first&.media
+    end
+  end
+
   field :all, Types::LibraryEntry.connection_type, null: false do
     description 'All Library Entries for a specific Media'
     argument :media_type, Types::Enum::MediaType, required: true
--- a/app/graphql/types/library.rb
+++ b/app/graphql/types/library.rb
@@ -2,7 +2,7 @@ class Types::Library < Types::BaseObject
   description 'The user library filterable by media_type and status'
 
   field :random_media, Types::Interface::Media, null: true do
-    description 'Random anime or manga from your library'
+    description 'Random anime or manga from this library'
 
     argument :media_type, Types::Enum::MediaType, required: true
     argument :status, [Types::Enum::LibraryEntryStatus], required: true
--- a/app/graphql/types/query_type.rb
+++ b/app/graphql/types/query_type.rb
@@ -165,7 +165,7 @@ class Types::QueryType < GraphQL::Schema::Object
 
     User.current.library_entries
         .where(media_type: media_type, status: library_entry_statuses)
-        .order(Arel.sql('RANDOM()')).first.media
+        .order(Arel.sql('RANDOM()')).first&.media
   end
 
   field :global_trending, Types::Interface::Media.connection_type, null: false do
--- a/app/graphql/types/query_type.rb
+++ b/app/graphql/types/query_type.rb
@@ -119,8 +119,8 @@ class Types::QueryType < GraphQL::Schema::Object
     service.search(title)
   end
 
-  field :random_media, Types::Interface::Media, null: true do
-    description 'Random Safe-for-Work Anime or Manga'
+  field :random_media, Types::Interface::Media, null: false do
+    description 'Random anime or manga'
     argument :media_type, Types::Enum::MediaType, required: true
     argument :age_ratings, [Types::Enum::AgeRating],
       required: true,
@@ -141,6 +141,8 @@ class Types::QueryType < GraphQL::Schema::Object
   def random_media(media_type:, age_ratings:)
     media = nil
 
+    # Most cases should find a media first time, but there is a chance
+    # due to TABLESAMPLE one won't be found.
     while media.blank?
       media = media_type.safe_constantize
                         .from(Arel.sql("#{media_type.downcase} TABLESAMPLE BERNOULLI(1)"))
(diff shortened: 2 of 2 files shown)


wopian:
Can the (n)sfw option be an argument instead of a duplicated field?

E.g `randomMedia(content: sfw | nsfw | all)` (I can't think of a more descriptive argument name atm)

toyhammered:
> Can the (n)sfw option be an argument instead of a duplicated field?
> 
> E.g `randomMedia(content: sfw | nsfw | all)` (I can't think of a more descriptive argument name atm)

I feel like they should be separated. I don't know if there is any value in having an `all` option.

app/graphql/types/query_type.rb
```diff
@@ -111,6 +113,21 @@ def random_media(media_type:, age_ratings:)
     media
   end
 
+  field :random_user_media, Types::Interface::Media, null: true do
```
wopian: `:random_library_media`? Thoughts?

wopian (commented):
👍 on using an age ratings array. That's a much better solution than `nsfw` or `sfw` - allows API consumers to offer age-appropriate options if they desire.

app/graphql/types/query_type.rb
toyhammered: hmmm... I think user makes more sense??

NuckChorris: How about putting it under the Library type?

```graphql
query {
  currentAccount {
    profile {
      library {
        randomMedia(status: PLANNED)
      }
    }
  }
}
```

toyhammered: hmm hmm, I'm not sure if I like that or not. It is in an understandable place (but might be to hidden).

wopian: I think putting it under the library scope is a good idea.

wopian: Also can status be an array too

toyhammered: yeah, it is, nuck just made a mistake in the example.

wopian: I was just looking at the wrong code snippet (the Media type response instead of the argument 😭)

app/graphql/types/query_type.rb
```diff
@@ -119,6 +119,40 @@ def search_media_by_title(title:)
...
+  end
+
+  def random_media(media_type:, age_ratings:)
+    media = nil
+
+    # Most cases should find a media first time, but there is a chance
+    # due to TABLESAMPLE one won't be found.
+    while media.blank?
+      media = media_type.safe_constantize
+                        .from(Arel.sql("#{media_type.downcase} TABLESAMPLE BERNOULLI(1)"))
```
wopian: SYSTEM_ROWS(1) for significantly faster table processing for a small reduction in randomness.

Extension needs to be added to psql initialisation (Google psql system rows for extension name)

This can remove the comment above about sometimes not returning a result as it will always return 1 result (table row) as long as the media type (or library) has at least one resource.

wopian (changes requested):

app/graphql/types/query_type.rb
NuckChorris: A migration like this should work to enable the extension:

```ruby
def up
  enable_extension 'tsm_system_rows'
end
```

toyhammered: ok, thank you.

toyhammered: @wopian @NuckChorris  Just ran this `media_type.safe_constantize.from(Arel.sql("#{media_type.downcase} TABLESAMPLE SYSTEM_ROWS(1)"))` and this literally only returns one record. It still has to filter out the `age_rating` which could result in a null. Not sure what the best way to handle retrying (or if this modification is even worth it?)

wopian: Don't you want to be filtering before randomly picking anyway? It's still considerably faster than BERNOULLI and the other methods (300μs to 20ms to filter + a couple of ms to sort, which is probably edging into μs with a heavily reduced pool to pick from). 

Otherwise you will get a sampled 1% of rows that don't contain the requested age rating on quite a few occasions when its scaled to mass usage. 

There's only ~1300 rows with the `age_rating` enum of `3` (<8% of anime) and ~960 rows with the `age_rating` enum of `1` (<2% of manga). 

5% of anime and 48% of manga do not have an age rating.

wopian: In fact, with `SELECT * from anime WHERE (age_rating = 0 AND random() <= 0.01) LIMIT 1` (1% sample, similar to `BERNOULLI(1)`). Increasing to 10% (`0.1`) has no discernible impact on the query times. Faster than `SYSTEM_ROWS` on the entire table and at worst, about the same or a little slower (manga)

`age_rating` | anime | manga
:-: | :-: | :-:
0 | 80μs-400μs | 40μs-10ms
1 | 60μs-400μs | 40μs-900μs
2 | 50μs-280μs | 80μs-1.2ms
3 | 70μs-540μs | 230μs-4.2ms
(no filter) | 80μs-190μs | 30μs-140μs

(20x explain analyse each on 17k and 56k rows respectively, rounded to 2 significant figures)

toyhammered: Sooo... Use what you are saying above?

wopian: I guess. I didn't realise TABLESAMPLE could only be used on an unfiltered table (you'd need to create a temporary table to do it, which I presume would wipe all/most of the speed benefits it had over the painfully slow initial implementation of `ORDER BY random()`).

Had the requirements been "get _any_ media" then TABLESAMPLE would definitely have been the way to go.

wopian (approved):

app/graphql/types/library.rb
```diff
@@ -1,6 +1,19 @@
...
+  field :random_media, Types::Interface::Media, null: true do
+    description 'Random anime or manga from this library'
+
+    argument :media_type, Types::Enum::MediaType, required: true
+    argument :status, [Types::Enum::LibraryEntryStatus], required: true
+  end
+
+  def random_media(media_type:, status:)
+    library_entries(media_type: media_type, status: status).then do |library_entries|
+      library_entries.order(Arel.sql('RANDOM()')).first&.media
```
wopian, line 13: Same here

wopian (changes requested):

wopian (approved):

PR #848, opened 2020-09-19
GraphQL - trending change argument string to enum

toyhammered:
# What
Change the `medium` argument to be an Enum and not have to run validation on a string.

# Why
Enum is better than a String in this case.

# Checklist

- [ ] All files pass Rubocop

- [ ] All the tests pass

### DIFF
--- a/app/graphql/types/query_type.rb
+++ b/app/graphql/types/query_type.rb
@@ -81,22 +81,19 @@ class Types::QueryType < GraphQL::Schema::Object
 
   field :global_trending, Types::Interface::Media.connection_type, null: false do
     description 'List trending media on Kitsu'
-    argument :medium, String, required: true
+    argument :medium, Types::Enum::MediaTypes, required: true
   end
 
   def global_trending(medium:)
-    raise GraphQL::ExecutionError, 'Unknown medium' unless %w[Anime Manga].include?(medium)
-
     TrendingService.new(medium.safe_constantize, token: context[:token]).get(10)
   end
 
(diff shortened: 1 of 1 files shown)


app/graphql/types/query_type.rb
```diff
@@ -81,22 +81,19 @@ def find_manga_by_slug(slug:)
 
   field :global_trending, Types::Interface::Media.connection_type, null: false do
     description 'List trending media on Kitsu'
-    argument :medium, String, required: true
+    argument :medium, Types::Enum::MediaTypes, required: true
```
wopian:
```suggestion
    argument :media_type, Types::Enum::MediaTypes, required: true
```

app/graphql/types/query_type.rb
```diff
@@ -81,22 +81,19 @@ def find_manga_by_slug(slug:)
...
   def global_trending(medium:)
-    raise GraphQL::ExecutionError, 'Unknown medium' unless %w[Anime Manga].include?(medium)
-
     TrendingService.new(medium.safe_constantize, token: context[:token]).get(10)
   end
 
   field :local_trending, Types::Interface::Media.connection_type, null: false do
     description 'List trending media within your network'
-    argument :medium, String, required: true
+    argument :medium, Types::Enum::MediaType, required: true
```
wopian:
```suggestion
    argument :media_type, Types::Enum::MediaType, required: true
```

wopian (changes requested):
Argument name is inconsistent with the rest of usage in GraphQL

wopian (approved):

PR #856, opened 2020-10-04
GraphQL - search by user slug

toyhammered:
# What
Integrate with Algolia to allow search by user slug.

# Why

# Checklist

- [ ] All files pass Rubocop

- [ ] All the tests pass

### DIFF
--- a/app/graphql/types/query_type.rb
+++ b/app/graphql/types/query_type.rb
@@ -157,6 +157,19 @@ class Types::QueryType < GraphQL::Schema::Object
     ::User.find_by(slug: slug)
   end
 
+  field :search_profile_by_slug, Types::Profile.connection_type, null: true do
+    description <<~DESCRIPTION.squish
+      Search for User by slug using Algolia.
+      The most relevant results will be at the top.
+    DESCRIPTION
+    argument :slug, String, required: true
+  end
+
+  def search_profile_by_slug(slug:)
+    service = AlgoliaGraphqlSearchService.new(::User, context[:token])
+    service.search(slug)
+  end
+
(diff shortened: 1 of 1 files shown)


wopian:
The slight problem with this is mobile users and new web users won't be searchable as they don't have a slug. Of course, search by display name could be added in addition.

wopian:
The slight problem with this is mobile users and new web users won't be searchable as they don't have a slug (the choice to not show slug during on boarding absolutely sucks - I'll see if its OK to add it for web users, mobile would have to wait for @NuckChorris' migration work to Expo). Of course, search by display name could be added in addition.

NuckChorris:
Yeah this search should use display name.  IMO we should continue not requiring a slug, it's a silly legacy thing for the most part.

wopian:
> Yeah this search should use display name. IMO we should continue not requiring a slug, it's a silly legacy thing for the most part.

But mentioning people without a slug is impossible (who knows someone else's ID???)

toyhammered:
> > Yeah this search should use display name. IMO we should continue not requiring a slug, it's a silly legacy thing for the most part.
> 
> But mentioning people without a slug is impossible (who knows someone else's ID???)

#4195

NuckChorris (changes requested):
This is weirdly specific, we pretty much never need to search by slug, you probably want to search by username instead

toyhammered:
> This is weirdly specific, we pretty much never need to search by slug, you probably want to search by username instead

Updated to search by username (prioritize name, then slug)

NuckChorris (approved):

PR #869, opened 2020-10-22
Generate Nolt SSO tokens

matthewdias:


app/graphql/types/session.rb
```diff
@@ -16,4 +16,15 @@ def account
...
     object unless object.blank?
   end
+
+  field :nolt_token, String,
+    null: false,
+    description: 'Single sign-on token for Nolt'
+
+  def nolt_token
+    user = context[:user]
+    raise GraphQL::ExecutionError, 'You must be logged in to do that' if user.blank?
```
toyhammered, line 26: I feel like we should create an `auth_field` at some point that does this. Prob not this PR just wanted to bring the idea up.

matthewdias: Yeah seems like a good idea

NuckChorris (approved):

PR #874, opened 2020-10-30
Mod Permissions Revisited

NuckChorris:
# What

Switches permissions from rolify to active_flag, and integrates the new active_flag-based permissions into Pundit policies.

All policies state which role administers them (assuming it's a policy for something which is admin-managed)

# Why

Before, we used Rolify to set permissions by assigning users roles within scopes such as `:admin, Manga` and then Pundit would detect that and apply the permissions automatically based on the underlying model. This seemed magical as an idea, but never really aligned with how we think about permissions.

We have admins, community mods, and database mods. These... don't really map into Rolify well. Mostly when we added a new mod I would copy an existing mod's roles and pray that it worked. It often didn't, and this made all kinds of problems for the mod team.

This new approach creates a `permissions` bitfield with just three permissions for now: admin, community_mod, and database_mod. Each one is either true or false. Conveniently, active_flag fields are settable by factory_bot so we can just use `create(:user, permissions: %i[community_mod database_mod admin])` to get a user with all permissions enabled. It's really nice!

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -42,7 +42,7 @@ class UsersController < ApplicationController
   end
 
   def alts
-    return render_jsonapi_error(401, 'Not permitted') unless user&.has_role?(:admin, User)
+    return render_jsonapi_error(401, 'Not permitted') unless user&.permissions&.community_mod?
     target_user = User.find(params[:id])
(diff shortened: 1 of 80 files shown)


toyhammered (approved):
This is bloody gorgeous!

PR #875, opened 2020-10-30
GraphQL - Post Lock

toyhammered:
# What

Allow posts to be locked. When a post is locked only admin will be able to make comments.
Locking/Unlocking a post will be done through a GraphQL Mutation.

# Why

Certain people don't know how to be nice to others. 🖕 them.

# Checklist

- [ ] All files pass Rubocop

- [ ] All the tests pass


wopian:
Will need to be adjusted for https://github.com/hummingbird-me/kitsu-server/pull/874

db/migrate/20201030020815_add_lock_unlock_to_post.rb
```diff
@@ -0,0 +1,12 @@
+class AddLockUnlockToPost < ActiveRecord::Migration[5.1]
+  def change
+    add_column :posts, :locked_by, :integer
+    add_column :posts, :locked_at, :datetime
+    add_column :posts, :lock_reason, :integer
+    add_index :posts, :locked_by
+
+    add_column :posts, :unlocked_by, :integer
+    add_column :posts, :unlocked_at, :datetime
+    add_index :posts, :unlocked_by
```
NuckChorris: This seems like it confuses matters. I think we should just unset the locked columns when we unlock, and worry about a proper audit log later

spec/factories/posts.rb
```diff
@@ -46,5 +46,11 @@
   factory :post do
     user
     content { Faker::Lorem.sentence }
+
+    trait :locked do
+      locked_by { user.id }
+      locked_at { DateTime.now }
+      lock_reason { 0 }
```
NuckChorris: I think you can do `lock_reason { :SPAM }` to be a bit more legible

db/migrate/20201030020815_add_lock_unlock_to_post.rb
wopian: Does rails admin only produce edit histories if it was edited within rails admin?

toyhammered: Ok, wasn't sure how we wanted to track this really.

toyhammered:
> Will need to be adjusted for #874

Yup, will let @NuckChorris merge in first and then fix mine up.

app/models/post.rb
```diff
@@ -51,13 +51,15 @@ class Post < ApplicationRecord
   update_algolia 'AlgoliaPostsIndex'
   embed_links_in :content, to: :embed
 
+  enum locked_reason: { SPAM: 0, TOO_HEATED: 1 }
```
NuckChorris: These should be `{ spam: 0, too_heated: 1 }` and then mapped from lowercase symbol in GraphQL

Rails and GraphQL case their enums differently 😅 Rails also generates methods, so you could do `post.SPAM?` with this, which just looks silly

app/policies/post_policy.rb
```diff
@@ -36,6 +37,13 @@ def group
     record.target_group
   end
 
+  def update_lock?
+    return true if can_administrate?
+    return true if group && has_group_permission(:content)
```
NuckChorris:
```suggestion
    return true if group && has_group_permission?(:content)
```

app/graphql/mutations/post/unlock_post.rb
```diff
@@ -0,0 +1,31 @@
...
+    required: true,
+    description: 'Unlock a Post.',
+    as: :post
+
+  field :post, Types::Post, null: true
+
+  def load_post(value)
+    post = ::Post.find(value.id)
+    post.assign_attributes(value.to_model)
+    post
```
NuckChorris, line 13: I'm distinctly not a fan of this thing where you get the changes from some other file and then shove them in over here... It seems like indirection without purpose

app/graphql/mutations/post/unlock_post.rb
```diff
@@ -0,0 +1,31 @@
...
+  def resolve(post:)
+    post.save
+
+    if post.errors.any?
+      Errors::RailsModel.graphql_error(post)
+    else
+      {
+        post: post
+      }
+    end
```
NuckChorris, line 29: FWIW I added an `Errors::ActiveRecord::RecordInvalid` in #880 that should do this for us in future with `post.save!`

We can change it once both are merged

app/graphql/types/input/post/lock.rb
```diff
@@ -0,0 +1,8 @@
+class Types::Input::Post::Lock < Types::Input::Base
+  argument :id, ID, required: true
+  argument :locked_reason, Types::Enum::LockReason, required: true
+
+  def to_model
+    to_h.merge(locked_at: DateTime.current, locked_by: current_user)
+  end
```
NuckChorris, line 7: This feels way too magical for me

app/models/post.rb
toyhammered: `post.SPAMSPAMSPASMAPSAPSMAPMS`

app/graphql/types/input/post/lock.rb
toyhammered: kyubey

app/graphql/mutations/post/unlock_post.rb
toyhammered: Funny enough this is how it was originally done (by Matt). I liked it so I kinda just started using it also lol.

app/graphql/types/enum/locked_reason.rb
```diff
@@ -0,0 +1,4 @@
+class Types::Enum::LockedReason < Types::Enum::Base
+  value 'SPAM', value: 'spam'
```
wopian, line 2: Closure for community events (or ancient Kitsu group posts? ... those half year old posts are still 95% of my notifications...)

```suggestion
  value 'CLOSED', value: 'closed'
  value 'SPAM', value: 'spam'
```

wopian (approved):
LGTM from a feature perspective for frontend. Will need Emma's review for code quality

NuckChorris (approved):
My comments still stand, I'd like to see improvements to the way we do GraphQL — but for now, this works

PR #880, opened 2020-10-31
GraphQL API: Site Permissions

NuckChorris:
# What

```graphql
mutation {
  account {
    setSitePermissions(input: { accountId: "5554", permissions: [ADMIN] }) {
      account {
        id
        permissions
      }
    }
  }
}
```

# Why

Whenever we add and remove moderators, I have to pop into a rails console to manually do it. This is the first step to on-site permission management.

# Checklist

- [x] All files pass Rubocop

- [x] All the tests pass


app/graphql/mutations/account/grant_site_permission.rb
```diff
@@ -0,0 +1,29 @@
...
+  argument :account, ID,
+    required: true,
+    description: 'Who to grant permissions to'
+
+  argument :permission, Types::Enum::SitePermission,
+    required: true,
+    description: 'The permission to grant to this user'
+
+  field :id, ID, null: false
+  field :permissions, [Types::Enum::SitePermission], null: false
```
NuckChorris: The reason for the mutation response objects looking like this is to allow Apollo clients to automatically update their cache with the mutation's changes. That said, I'm not sure this even works right. I'm not super familiar with Apollo's caching.

Maybe we should instead have the mutation return the entire Account object?

toyhammered: If this actually has caching benefits that is fine, if not, I would prefer the actual object (Account)

app/graphql/mutations/account/grant_site_permission.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  def load_account(value)
+    ::User.find(value)
+  end
+
+  def authorized?
+    current_user.permissions.admin?
```
toyhammered: Should we be adding this as a pundit scope?

app/graphql/mutations/account/grant_site_permission.rb
```diff
@@ -0,0 +1,29 @@
+class Mutations::Account::GrantSitePermission < Mutations::Base
+  argument :account, ID,
+    required: true,
+    description: 'Who to grant permissions to'
+
+  argument :permission, Types::Enum::SitePermission,
+    required: true,
+    description: 'The permission to grant to this user'
+
```
toyhammered: If we want to follow the way other mutations are implemented, we should have this inside an input.

app/graphql/mutations/account/revoke_site_permission.rb
```diff
@@ -0,0 +1,29 @@
+class Mutations::Account::RevokeSitePermission < Mutations::Base
+  argument :account, ID,
+    required: true,
+    description: 'Who to grant permissions to'
+
+  argument :permission, Types::Enum::SitePermission,
+    required: true,
+    description: 'The permission to grant to this user'
```
toyhammered: same as above.

app/graphql/types/mutation_type.rb
```diff
@@ -1,5 +1,6 @@
 class Types::MutationType < Types::BaseObject
   field :pro, Types::Mutations::ProMutation, null: false
+  field :account, Types::Mutations::AccountMutation, null: true
   field :anime, Types::Mutations::AnimeMutation, null: true
   field :episode, Types::Mutations::EpisodeMutation, null: true
   field :library_entry, Types::Mutations::LibraryEntryMutation, null: true
```
toyhammered: just want to leave this here, should we maybe alphabetize these?

toyhammered (changes requested):
couple small questions, but love it!

app/graphql/mutations/account/grant_site_permission.rb
NuckChorris: What do you mean by that? If you mean wrapping it in an input type, that's really not helpful for a simple mutation?

app/graphql/types/mutation_type.rb
NuckChorris: I was thinking that too, honestly

app/graphql/mutations/account/grant_site_permission.rb
NuckChorris: I was thinking about it, but I'm still not sure how best to represent it in Pundit.  Pundit's model is very much model-centric vs action-centric :/

toyhammered: We discussed this on slack and concluded we should just add more pundit methods.

app/graphql/mutations/account/set_site_permissions.rb
```diff
@@ -0,0 +1,21 @@
...
+    super(input.account, :set_site_permissions?)
+  end
+
+  def resolve(input:)
+    permissions = input.permissions.map(&:to_sym)
+
+    res = Accounts::SetSitePermissions.call(
+      user: input.account,
+      permissions: permissions
+    )
```
toyhammered, line 17: What type of errors can happen here and should we account for them?

app/graphql/types/input/account/set_site_permissions.rb
```diff
@@ -0,0 +1,14 @@
...
+    description: 'Who to grant permissions to',
+    as: :account
+
+  argument :permissions, [Types::Enum::SitePermission],
+    required: true,
+    description: 'The permissions to set for this user'
+
+  def load_account(value)
+    ::User.find(value)
+  end
```
toyhammered, line 13: Do we want this here, or inside the mutation?

toyhammered (commented):
Some questions, but looks good for the most part!

app/graphql/types/input/account/set_site_permissions.rb
NuckChorris: Here, because we use it in `authorized?`

app/graphql/mutations/account/set_site_permissions.rb
NuckChorris: `UnknownPermissions` could happen due to a programming error but otherwise it won't happen, and the only other ones are `RecordNotFound` or `RecordInvalid` which I believe are accounted for automatically

app/graphql/types/input/account/set_site_permissions.rb
toyhammered: Ooooh, I see what you've done. If you want to go that route shouldn't we rename the `value` argument to `id` to be more specific?

Also unless I am thinking of this wrong from the `load_account` it takes what that returns and passes that along to the `authorized?` method (which would be a user) not an input? 

I'll test this locally tomorrow because either this won't work at all or you are about to teach me something new (and I'm hoping it is the latter because this really opens up the possibilities).

NuckChorris: I don't think I'm doing anything weird, I'm pretty sure I've just copied it from your stuff… my thinking is that this should run on the input type to load the record and then when `authorized?` runs it's available as input.account

toyhammered: So the `load_*` isn't actually related to an `input`. Its the graphql-ruby gem magic and it must be in the mutation file itself.

PR #893, opened 2020-11-20
GraphQL - downcase all slug lookups

toyhammered:
# What

When looking up using a slug, this will always downcase it before hand.

# Why

# Checklist

- [ ] All files pass Rubocop

- [ ] All the tests pass

### DIFF
--- a/app/graphql/types/query_type.rb
+++ b/app/graphql/types/query_type.rb
@@ -37,7 +37,7 @@ class Types::QueryType < GraphQL::Schema::Object
 
   field :find_anime_by_slug, Types::Anime, null: true do
     description 'Find a single Anime by Slug'
-    argument :slug, String, required: true
+    argument :slug, String, required: true, prepare: ->(slug, _) { slug.downcase }
   end
 
   def find_anime_by_slug(slug:)
@@ -89,7 +89,7 @@ class Types::QueryType < GraphQL::Schema::Object
 
   field :find_manga_by_slug, Types::Manga, null: true do
     description 'Find a single Manga by Slug'
-    argument :slug, String, required: true
+    argument :slug, String, required: true, prepare: ->(slug, _) { slug.downcase }
   end
 
   def find_manga_by_slug(slug:)
@@ -187,7 +187,7 @@ class Types::QueryType < GraphQL::Schema::Object
 
   field :find_profile_by_slug, Types::Profile, null: true do
     description 'Find a single User by Slug'
-    argument :slug, String, required: true
+    argument :slug, String, required: true, prepare: ->(slug, _) { slug.downcase }
   end
 
   def find_profile_by_slug(slug:)
@@ -218,7 +218,7 @@ class Types::QueryType < GraphQL::Schema::Object
 
   field :find_character_by_slug, Types::Character, null: true do
     description 'Find a single Character by Slug'
-    argument :slug, String, required: true
+    argument :slug, String, required: true, prepare: ->(slug, _) { slug.downcase }
   end
 
   def find_character_by_slug(slug:)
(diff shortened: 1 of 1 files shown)


wopian:
I don't think this will work. [Slugs are case-**in**sensitive](https://github.com/hummingbird-me/kitsu-server/blob/a816c093e6c3df2151198e23f94bdde2d19ad9df/app/models/user.rb#L188) but the query performed in GraphQL _is_ case-sensitive. I believe the Postgres query will need to be told to use something along the lines of `WHERE slug ILIKE :slug` or the [citext module](http://www.postgresql.org/docs/current/interactive/citext.html), unless there's another way in `graphql-ruby`.

### Before PR

| Actual Slug | findProfileBySlug | Found Profile? 
| :-: | :-: | :-:
| `wopian` (l) | `wopian` (l) | ✔️ 
| `wopian` (l) | `Wopian` (u) | ❌ 
| `Wopian` (u) | `wopian` (l) | ❌ 
| `Wopian` (u) | `Wopian` (u) | ✔️ 

### With PR

| Actual Slug | findProfileBySlug  | Found Profile? 
| :-: | :-: | :-:
| `wopian` (l) | `wopian` (l) | ✔️ 
| `wopian` (l) | `Wopian` (u) | ✔️ 
| `Wopian` (u) | `wopian` (l) | ❌ 
| `Wopian` (u) | `Wopian` (u) | ❌ 

### JSON:API

| Actual Slug | filter[slug] | Found Profile? 
| :-: | :-: | :-:
| `wopian` (l) | `wopian` (l) | ✔️ 
| `wopian` (l) | `Wopian` (u) | ✔️ 
| `Wopian` (u) | `wopian` (l) | ✔️ 
| `Wopian` (u) | `Wopian` (u) | ✔️

wopian (changes requested):
See above

wopian:
~~Alternatively we could lowercase slugs on create/change, migrate existing slugs to lowercase and downcase search parameters?~~

Thoughts @NuckChorris?

NuckChorris:
I think @wopian hit the nail on the head, that this won't work. In fact, my own slug is `Nuck`, so my profile wouldn't work with this.

Interestingly, we [already use the `citext` module for user slugs](https://github.com/hummingbird-me/kitsu-server/blob/db0da1ec37195cb441c6d74aceaefedfcb2f28d7/db/schema.rb#L1596), so this should already work. It's weird that it doesn't!

wopian:
> Interestingly, we [already use the `citext` module for user slugs](https://github.com/hummingbird-me/kitsu-server/blob/db0da1ec37195cb441c6d74aceaefedfcb2f28d7/db/schema.rb#L1596), so this should already work. It's weird that it doesn't!

That's really odd.

Edit: Actually kinda makes sense for the error @toyhammered - it's getting `wopian` back (my slug) but trying to case-sensitive match somewhere (on GraphQL side).

```json
"error": {
  "message": "key not found: \"wopian\"\nDid you mean?  \"Wopian\""
}
```

```gql
query {
  findProfileBySlug(slug: "Wopian") {
    name
  }
}
```

Edit 2: Slugs with capitals only get `"message": "key not found: \"Nuck\""` when `findProfileBySlug(slug: "nuck")`

Edit 3: GraphQL(-Ruby)'s `String` type might not be aware of `CITEXT`?

wopian:
@toyhammered, for specifically users, should `findProfileBySlug` also do an ID lookup if the input is all numbers? (All-number slugs are not allowed, so it _will_ be an ID)  

Since for most mobile users, their ID is their slug as far as the mobile and web clients are concerned.

I feel if it's left to API consumers to correctly perform either of the two options (slug for slugs and Id for Ids), users without slugs will unintentionally be left out of the developers' apps...I've already made this mistake once in another repo that supports multiple services, where I naively assumed it already did this 🙄

toyhammered:
Thanks for all the research and testing @wopian! I'll see what is causing this error.

NuckChorris:
Yknow what, it's the data loader. The batch code uses the key you pass to find the corresponding returned record in a hash. It'll need to be adjusted to allow for the case insensitive matching

NuckChorris:
I just opened #929 with a full solution so I'm closing this one!

PR #898, opened 2020-11-27
GraphQL - Embeds

toyhammered:
# What

# Why

# Checklist

- [ ] All files pass Rubocop

- [ ] All the tests pass


app/graphql/concerns/optional_embed_description.rb
```diff
@@ -0,0 +1,23 @@
...
+    The locale these tags are marked up in.
+    Of the format language_TERRITORY. Default is en_US.
+  ]
+  LOCALE_ALTERNATIVE_DESCRIPTION = 'An array of other locales this page is available in.'
+  SITE_NAME_DESCRIPTION = %[
+    If your object is part of a larger web site,
+    the name which should be displayed for the overall site.
+  ]
+  VIDEO_DESCRIPTION = 'A URL to a video file that complements this object.'
+end
```
NuckChorris: What is this file? I'm legit confused by it

app/graphql/types/interface/required_embed.rb
```diff
@@ -0,0 +1,28 @@
+module Types::Interface::RequiredEmbed
```
NuckChorris: I don't really love the name, can we just make it BaseEmbed?

app/graphql/types/union/embed_item.rb
```diff
@@ -0,0 +1,5 @@
+class Types::Union::EmbedItem < Types::Union::Base
```
NuckChorris: Is there a good argument to using a union vs just referencing the base embed interface that all embeds inherit from?

NuckChorris (changes requested):

app/graphql/types/union/embed_item.rb
toyhammered: I'm not actually sure. I've always preferred using a union vs referencing the base embed, I think it reads nicer. But both function the same way.

app/graphql/concerns/optional_embed_description.rb
toyhammered: When we add optional fields to specific embeds so that we don't need to copy paste all the descriptions. I have mixed feelings about this.

app/graphql/types/union/embed_item.rb
toyhammered: After reading this: https://artsy.github.io/blog/2019/01/14/graphql-union-vs-interface/ I think I prefer Unions.

app/graphql/concerns/optional_embed_description.rb
NuckChorris: You can just use interfaces for those fields like VideoEmbed might have a video field, etc

toyhammered: yeah this was entirely a mistake on how I was thinking about optional fields. Fixed it up.

app/graphql/types/interface/base_embed.rb
```diff
@@ -0,0 +1,69 @@
...
+    null: true,
+    description: <<~DESCRIPTION.squish
+      The locale these tags are marked up in.
+      Of the format language_territory.
+      Default is en_us.
+    DESCRIPTION
+
+  field :locale_alternative, [String],
+    null: true,
+    description: 'An array of other locales this page is available in.'
```
NuckChorris, line 53: I don't think we use any of these

app/graphql/types/embed/audio_tag_embed.rb
```diff
@@ -0,0 +1,5 @@
+class Types::Embed::AudioTagEmbed < Types::BaseObject
```
NuckChorris, line 1: Why the `Tag`?

app/graphql/types/interface/base_embed.rb
toyhammered: should we not start bringing them in? They are optional.

app/graphql/types/embed/audio_tag_embed.rb
toyhammered: seemed kind of strange just being `AudioEmbed` and on the docs it used tag in the description. i.e: `The og:video tag has the identical tags` which I kinda liked. I can rename though.

PR #929, opened 2021-01-26
SlugLoader

NuckChorris:
# What
Solves the same problem that #893 attempted to

- Fixed duplicate slugs in all models
- Converted all `slug` fields to `citext` type
- Created a `SlugLoader` to handle loading things by their slug

# Why

## Fixing duplicate slugs
It turns out we have a number of slugs which were reused with different case, which breaks when we change to a citext column. To fix this, I have a query which uses window functions to find duplicates and then we set `slug: nil` to have friendly_id regenerate a slug for them.

## Converting all `slug` fields to `citext`
Slugs are case-insensitive, but we didn't always use `citext` to handle that. Traditionally, we used a normal string column and just downcased it, but that's not super reliable and Postgres has a lovely `citext` type which is perfect for this. We started using `citext` when we added slugs to users, but didn't ever switch other models to it, until now.

The reason why I did this as part of this PR is to completely solve the slug querying problem listed in comments on #893 

## Creating a `SlugLoader`
This is a very small loader class, inherited from `RecordLoader`, which just sets the query column to `slug` and converts the `cache_key` to lowercase. The reason for this is laid out in [a comment on #893](https://github.com/hummingbird-me/kitsu-server/pull/893#issuecomment-735041185), but basically boils down to the fact that loaders need to match loaded records to the call that asked for them, and they can't do that if the record's slug differs from the one that was passed in... unless we override `cache_key` to lowercase them both!

# Checklist

- [x] All files pass Rubocop

- [x] All the tests pass

### DIFF
--- a/spec/graphql/loaders/slug_loader_spec.rb
+++ b/spec/graphql/loaders/slug_loader_spec.rb
@@ -1,0 +1,80 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Loaders::SlugLoader do
+  context 'with an uppercase slug' do
+    let!(:anime) { create(:anime, slug: 'ANIME-TEST') }
+
+    it 'should find with a lowercase slug' do
+      slug = GraphQL::Batch.batch do
+        Loaders::SlugLoader.for(Anime).load('anime-test').then(&:slug)
+      end
+      expect(slug).to eq('ANIME-TEST')
+    end
+
+    it 'should find with an uppercase slug' do
+      slug = GraphQL::Batch.batch do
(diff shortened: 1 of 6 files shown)


spec/graphql/loaders/slug_loader_spec.rb
```diff
@@ -0,0 +1,80 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Loaders::SlugLoader do
```
toyhammered, line 5: Love it! Great example for how to do batch loading testing.

toyhammered (approved):
LGTM!

PR #959, opened 2021-02-15
Simplify GraphQL Mutation Error Handling

NuckChorris:
# What

- Added simpler error handling system based on standard rescue and Module#prepend
- Delete CustomErrors and CustomPayloadType
- Move mutation namespaces into the graphql/mutations folder

# Why

The existing mutation error handling was complex and rife with metaprogramming. It wasn't easily extensible and depended on matching classnames in different namespaces. The new approach? It's a reusable rescue, that's it.

This enabled me to delete the CustomErrors code, and made the CustomPayloadType much shorter — in fact, it became so simple it was just adding a single field to the payloads — that I just deleted that too, and added the error field to the types manually.

Then, as one final bit of cleanup, I moved the mutation namespace objects into the mutations folder. Not because it's related, because it had been nagging at the back of my brain.


app/graphql/mutations/namespace.rb
```diff
@@ -0,0 +1,5 @@
+class Mutations::Namespace < Types::BaseObject
+  def self.default_graphql_name
+    "#{name.split('::')[1..-1].join}Mutations"
```
toyhammered, line 3: Do we want this to be singular
```suggestion
    "#{name.split('::')[1..-1].join}Mutation"
```

toyhammered (changes requested):
Honestly, 1 small question/ potential change, but this looks awesome!

app/graphql/mutations/namespace.rb
NuckChorris: Nah, it's a collection of mutations so plural makes sense!

toyhammered (approved):

PR #961, opened 2021-02-20
GraphQL - update #total_count to use items instead of nodes

toyhammered:
# What

Fixes total_count not working on connections.

# Why

In graphql 1.11.6 we used object.nodes but in graphql 1.12.0 they added the `first` to limit the amount of nodes it gets. We now grab the `object.items` because that uses ActiveRecord directly instead of the `nodes` method in graphql (which wraps AR). This will not trigger a limit based on the `first` method and grab all records now.

# Checklist

- [ ] All files pass Rubocop

- [ ] All the tests pass

### DIFF
--- a/app/graphql/types/base_connection.rb
+++ b/app/graphql/types/base_connection.rb
@@ -3,7 +3,12 @@ class Types::BaseConnection < GraphQL::Types::Relay::BaseConnection
     null: false,
     description: 'The total amount of nodes.'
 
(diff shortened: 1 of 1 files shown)


NuckChorris (approved):
I don't think this needs a comment spanning 5 lines, I'd like if you could cut that down a bit and maybe fit it in one line… but overall 👍

PR #966, opened 2021-02-28
FancyLoader

NuckChorris:
# What
- A really fancy loader system
- Handles sorting and batching and pagination and applying policies
- Hooked up with a FancyConnection wrapper
- Doesn't even require a custom field extension or connection type

# Why
We needed solutions for sorting and batch-loading with pagination. These are all individually complex, but also need to integrate with each other to work effectively.

I opted to create a lightweight DSL which allows you to define your sorts and base model (and gives you access to modify the Arel query) by subclassing FancyLoader, and then it mostly handles the complexity of batching for you.  It builds a giant SQL query for the batch, using window functions to handle pagination.


app/graphql/loaders/fancy_loader/query_generator.rb
```diff
@@ -0,0 +1,113 @@
...
+  #   ROW_NUMBER() OVER (#{partition})
+  def row_number
+    Arel::Nodes::NamedFunction.new('ROW_NUMBER', []).over(partition).as('row_number')
+  end
+
+  # A count window function. Omits sort from the partition to get the total count.
+  #
+  #   COUNT(*) OVER (#{partition})
+  def count
+    count_partition = Arel::Nodes::Window.new.partition(table[@find_by])
```
toyhammered, line 80: can move that to a new method.

app/graphql/loaders/fancy_loader/query_generator.rb
```diff
@@ -0,0 +1,113 @@
...
+    Arel::Nodes::NamedFunction.new('COUNT', [Arel.star]).over(count_partition).as('total_count')
+  end
+
+  def pagination_filter(query)
```
toyhammered, line 84: I think this whole method can be moved to another class.

app/graphql/loaders/fancy_loader/type_generator.rb
```diff
@@ -0,0 +1,41 @@
+##
+# Generate parameter types for sorting
+class Loaders::FancyLoader::TypeGenerator
+  class SortDirection < GraphQL::Schema::Enum
+    graphql_name 'SortDirectionEnum'
+
+    value :ASCENDING, value: :asc
+    value :DESCENDING, value: :desc
+  end
```
toyhammered: can we move to a separate file, this is gross :sob:

app/graphql/loaders/fancy_loader/type_generator.rb
```diff
@@ -0,0 +1,41 @@
...
+
+  def initialize(loader, name: loader.model.name)
+    @loader = loader
+    @name = name
+  end
+
+  def sorts_enum
+    sorts = @loader.sorts
+    name = "#{@name}SortEnum"
```
toyhammered: can we not just memoize these or does that not play nicely when creating dynamic classes?

app/graphql/types/comment.rb
```diff
@@ -26,21 +26,29 @@ class Types::Comment < Types::BaseObject
...
-    null: false,
-    description: 'All replies to a specific comment.'
+  field :replies, Types::Comment.connection_type, null: false do
+    description 'Replies to this comment'
+    argument :sort, Loaders::CommentsLoader.sort_argument, required: false
+  end
 
-  def replies
-    AssociationLoader.for(object.class, :replies, policy: :comment).scope(object)
+  def replies(sort: [{ on: :created_at, direction: :asc }])
```
toyhammered, line 48: I wonder if we can have some generic default, I think that can be useful elsewhere.

toyhammered (changes requested):
Holy moly this is amazing!

My one question is, does the query analyzer still work properly with this. It doesn't seem like you did anything that would cause it to mess up just want to double check.

app/graphql/loaders/fancy_loader/query_generator.rb
NuckChorris: Nah, the partition setup is super brief, I don't think it needs to be a separation method

app/graphql/loaders/fancy_loader/query_generator.rb
NuckChorris: Hmmmm... I'm not sure that would make things any more clear than this?

app/graphql/loaders/fancy_loader/type_generator.rb
NuckChorris: Okay that's fair lol idk why I crammed it in here

app/graphql/loaders/fancy_loader/query_generator.rb
toyhammered: It felt very bloated when reading it, almost out of place.

app/graphql/types/comment.rb
NuckChorris: I thought about that, but I think I'll kick the can down the road on it. I couldn't find anything which was more intuitive than this :(

app/graphql/loaders/fancy_loader/query_generator.rb
NuckChorris: It's bulky yes, but I think it would feel a lot weirder extracted — it is the _core_ of generating a query!

app/graphql/types/comment.rb
toyhammered: :(

app/graphql/loaders/fancy_loader/query_generator.rb
toyhammered: 🤔

app/graphql/loaders/fancy_loader/pagination_filter.rb
```diff
@@ -0,0 +1,56 @@
...
+  def row
+    @row ||= @query[:row_number]
+  end
+
+  def count
+    @count ||= @query[:total_count]
+  end
+
+  def after_filter
+    row.gt(@after) if @after
```
toyhammered: `return unless @after`

app/graphql/loaders/fancy_loader/pagination_filter.rb
```diff
@@ -0,0 +1,56 @@
...
+  end
+
+  def before_filter
+    row.lt(@before) if @before
```
toyhammered: `return unless @before`

toyhammered (changes requested):
Someone trained me to be like this, not sure who though 😆

toyhammered (approved):
LGTM!

PR #972, opened 2021-03-06
Global Stats - refactor global stats to only add stats with more than 5 media

toyhammered:
# What

Updated Global Stats to only add stats from users who have added more than 5 media.
https://kitsu.io/feedback/feature-requests/p/dont-include-people-with-5-completed-titles-or-below-in-the-more-than

# Why

The onboarding process forces you to add 5 media, and there are many new users who are coming to see if we stream. Also fixed deprecation warning to use `Arel.sql` to clarify safe "user" injection. 

# Checklist

- [ ] All files pass Rubocop

- [ ] All the tests pass


NuckChorris (approved):
:+1:

PR #992, opened 2021-04-02
GraphQL - Franchise + Installment implementation

toyhammered:
# What

# Why

# Checklist

- [ ] All files pass Rubocop

- [ ] All the tests pass

### DIFF
--- a/app/graphql/types/enum/installment_tag.rb
+++ b/app/graphql/types/enum/installment_tag.rb
@@ -1,7 +1,17 @@
+# frozen_string_literal: true
+
 class Types::Enum::InstallmentTag < Types::Enum::Base
-  value 'MAIN_STORY', 'The main story', value: :main_story
-  value 'SIDE_STORY', '', value: :side_story
-  value 'SPINOFF', 'Related to main story with a new narrative', value: :spinoff
-  value 'CROSSOVER', 'When characters from different media come together', value: :crossover
-  value 'ALTERNATE_SETTING', 'Same universe, with different characters or visa-versa', value: :alternate_setting
+  MAIN_STORY = 'The main story.'
+  SIDE_STORY = 'Takes place sometime during the main storyline.'
+  SPINOFF = 'Uses characters of a different series, but is not an alternate setting or story.'
+  CROSSOVER = 'Characters from different media meet in the same story.'
+  ALTERNATE_SETTING = 'Same universe/world/reality/timeline, completely different characters.'
+  ALTERNATE_VERSION = 'Same setting, same characters, story is told differently.'
+
+  value 'MAIN_STORY', MAIN_STORY, value: :main_story
+  value 'SIDE_STORY', SIDE_STORY, value: :side_story
+  value 'SPINOFF', SPINOFF, value: :spinoff
+  value 'CROSSOVER', CROSSOVER, value: :crossover
+  value 'ALTERNATE_SETTING', ALTERNATE_SETTING, value: :alternate_setting
+  value 'ALTERNATE_VERSION', ALTERNATE_VERSION, value: :alternate_version
 end
--- a/app/graphql/types/enum/installment_tag.rb
+++ b/app/graphql/types/enum/installment_tag.rb
@@ -1,0 +1,4 @@
+class Types::Enum::InstallmentTag < Types::Enum::Base
+  value 'SIDE_STORY', '', value: :side_story
+  value 'SPINOFF', '', value: :spinoff
+end
--- a/app/graphql/types/enum/installment_tag.rb
+++ b/app/graphql/types/enum/installment_tag.rb
@@ -1,4 +1,8 @@
 class Types::Enum::InstallmentTag < Types::Enum::Base
+  value 'MAIN_STORY', '', value: :main_story
   value 'SIDE_STORY', '', value: :side_story
   value 'SPINOFF', '', value: :spinoff
+  value 'MAIN_ENTRY', '', value: :main_entry
+  value 'CROSSOVER', '', value: :crossover
+  value 'ALTERNATE_SETTING', '', value: :alternate_setting
 end
--- a/app/graphql/types/enum/installment_tag.rb
+++ b/app/graphql/types/enum/installment_tag.rb
@@ -1,8 +1,7 @@
 class Types::Enum::InstallmentTag < Types::Enum::Base
-  value 'MAIN_STORY', '', value: :main_story
+  value 'MAIN_STORY', 'The main story', value: :main_story
   value 'SIDE_STORY', '', value: :side_story
-  value 'SPINOFF', '', value: :spinoff
-  value 'MAIN_ENTRY', '', value: :main_entry
-  value 'CROSSOVER', '', value: :crossover
-  value 'ALTERNATE_SETTING', '', value: :alternate_setting
+  value 'SPINOFF', 'Related to main story with a new narrative', value: :spinoff
+  value 'CROSSOVER', 'When characters from different media come together', value: :crossover
+  value 'ALTERNATE_SETTING', 'Same universe, with different characters or visa-versa', value: :alternate_setting
 end
--- a/app/graphql/types/franchise.rb
+++ b/app/graphql/types/franchise.rb
@@ -9,9 +9,14 @@ class Types::Franchise < Types::BaseObject
     null: false,
     description: 'The name of this franchise in various languages'
 
-  field :progession_order, Types::Enum::FranchiseProgressionOrder,
-    null: false,
-    description: ''
+  def titles
+    {
+      localized: object.titles,
+      alternatives: object.abbreviated_titles.presence || [],
+      canonical: object.canonical_title,
+      canonical_locale: object.canonical_title_key
+    }
+  end
 
   field :installments, Types::Installment.connection_type, null: true do
     description ''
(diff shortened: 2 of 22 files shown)


app/graphql/types/franchise.rb
```diff
@@ -0,0 +1,27 @@
...
+
+  field :id, ID, null: false
+
+  field :titles, Types::TitlesList,
+    null: false,
+    description: 'The name of this franchise in various languages'
+
+  field :progession_order, Types::Enum::FranchiseProgressionOrder,
+    null: false,
+    description: ''
```
NuckChorris: What is this field doing? Is it a default? If so, you need to apply it within the `def installments` below by setting `sort: [{ on: object.progression_order, direction: :desc }]` ⬇️

I don't think it's necessary for a customizable default though, so could we just remove it?

db/schema.rb
```diff
@@ -493,7 +493,7 @@
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
     t.hstore "titles", default: {}, null: false
-    t.string "canonical_title", default: "en_jp", null: false
```
NuckChorris: Why is canonical_title removed? 🤔

db/schema.rb
```diff
@@ -714,10 +714,11 @@
     t.integer "media_id"
     t.integer "franchise_id"
     t.string "media_type", null: false
-    t.integer "position"
+    t.integer "release_position"
```
NuckChorris: We should probably not name these `_position` because ranked_model creates a virtual column with the name `#{column}_position` which would mean `release_position_position`

How about `release_order`?

db/schema.rb
```diff
@@ -714,10 +714,11 @@
...
     t.string "tag"
     t.datetime "created_at"
     t.datetime "updated_at"
+    t.integer "chronological_position"
```
NuckChorris: I'm kinda thinking maybe `alternative_order` would be a better name for this, since it might not always be chronological???

app/graphql/types/installment.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  field :id, ID, null: false
+
+  field :release_position, Integer,
+    null: true,
+    description: ''
+
+  field :chronological_position, Integer,
+    null: true,
+    description: ''
```
NuckChorris: We probably want to hide the actual order values and make it look contiguous to clients, so they don't have to think about our database implementation details. ranked_model provides a `release_position_rank` method which does this, but does it by `where(x < y).count` which means a separate query each time.

Since that would have absolute garbage-ass performance, we'll probably want to add a window function to our query to get (effectively) `ROW_NUMBER() OVER (PARTITION BY #{table[@find_by]} ORDER BY release_position ASC)`

So we probably want to add an escape hatch to FancyLoader to allow modifying the `base_query` before it generates the rest of the query. I imagine it could look something like this:

```ruby
class Loaders::InstallmentsLoader < Loaders::FancyLoader
  from Installment
  modify_query ->(query) {
    release_rank = RankQueryGenerator.new(:release_position, partition).arel
    query.project(release_rank)
  }

  sort :release_position
  sort :chronological_position
end
```

Then RankQueryGenerator would just do a bunch of Arel stuff to build up the partition statement like the existing `Loaders::FancyLoader::QueryGenerator`

db/schema.rb
toyhammered: There shouldn't be a need for this? Shouldn't it just use based on what the user has set? Maybe I'm mixing something up.... What would be the purpose of this?

db/schema.rb
toyhammered: hmmm, I don't like alternative because there is normally a set way to watch something (give or take a few series which might get some opinions).

db/schema.rb
toyhammered: sure, I just kept position because it was already there.

app/graphql/types/installment.rb
toyhammered: interesting. I'll see if I can manage that.

app/graphql/types/franchise.rb
toyhammered: Ok, I'm fine removing it, we can always add later on if we realize we want it.

db/schema.rb
```diff
@@ -83,7 +83,7 @@
     t.jsonb "description", default: {}, null: false
     t.index ["age_rating"], name: "index_anime_on_age_rating"
     t.index ["average_rating"], name: "anime_average_rating_idx"
-    t.index ["average_rating"], name: "index_anime_on_wilson_ci", order: { average_rating: :desc }
+    t.index ["average_rating"], name: "index_anime_on_wilson_ci", order: :desc
```
toyhammered, line 86: @NuckChorris why would this change??

NuckChorris: New Rails version dumps slightly different, this is fine

app/graphql/types/enum/installment_tag.rb
```diff
@@ -0,0 +1,4 @@
+class Types::Enum::InstallmentTag < Types::Enum::Base
+  value 'SIDE_STORY', '', value: :side_story
+  value 'SPINOFF', '', value: :spinoff
```
Reinachan: May need more installment tags. Some suggested ones (just following the same syntax here)
`value 'MAIN_ENTRY', '', value: :main_entry`
`value 'CROSSOVER', '', value: :crossover`
`value 'ALTERNATE_SETTING', '', value: :alternate_setting`

app/graphql/types/query_type.rb
```diff
@@ -354,4 +354,24 @@ def find_library_event_by_id(id:)
...
+  field :search_franchise_by_title, Types::Franchise.connection_type, null: true do
+    description <<~DESCRIPTION.squish
+      Search for Franchise by title using Algolia.
+      The most relevant results will be at the top.
+    DESCRIPTION
+    argument :title, String, required: true
+  end
+
+  def search_franchise_by_title(title:)
+    # NOTE: add Algolia search to allow for this.
```
Reinachan: Needs to be completed

app/models/installment.rb
```diff
@@ -21,11 +21,16 @@
 
 class Installment < ApplicationRecord
   include RankedModel
-  ranks :position
+  ranks :release_order
+  ranks :alternative_order
+
+  enum tag: {
+    side_story: 0,
+    spinoff: 1
```
Reinachan: Assuming this is related to the abovementioned franchise tags, this will also need to be updated alongside that

Reinachan (commented):
Looks good to my untrained eye. Missing some tags. We may need more than the ones I suggested too. We'll also need algolia hooked up to this before we can do all that much with it.

app/graphql/types/query_type.rb
toyhammered: @NuckChorris

app/models/installment.rb
toyhammered: yes it is.

app/graphql/types/enum/installment_tag.rb
Reinachan: You can probably base it off of these. Not all of them are relevant but some might be

db/migrate/20210405030335_change_tag_type_in_installments.rb
```diff
@@ -0,0 +1,12 @@
+class ChangeTagTypeInInstallments < ActiveRecord::Migration[5.2]
+  disable_ddl_transaction!
+
+  def up
+    Installment.update_all(tag: nil)
```
toyhammered, line 5: @NuckChorris wasn't sure in sql how to convert empty strings to null so I just did it this way.

app/graphql/types/enum/installment_tag.rb
```diff
@@ -0,0 +1,8 @@
+class Types::Enum::InstallmentTag < Types::Enum::Base
+  value 'MAIN_STORY', '', value: :main_story
+  value 'SIDE_STORY', '', value: :side_story
+  value 'SPINOFF', '', value: :spinoff
+  value 'MAIN_ENTRY', '', value: :main_entry
```
NuckChorris: Is there a difference between MAIN_STORY and MAIN_ENTRY?

app/graphql/types/query_type.rb
NuckChorris: Let's skip this for now, I'll add it in a later PR

app/graphql/types/installment.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  field :id, ID, null: false
+
+  field :release_order, Integer,
+    null: true,
+    description: 'Order based by date released'
+
+  field :alternative_order, Integer,
+    null: true,
+    description: 'Order based chronologically'
```
NuckChorris: These should be the rank fields from the other PR you made 😅

NuckChorris (commented):
Looks pretty great! Could you add descriptions to the InstallmentTags?

app/graphql/types/installment.rb
toyhammered: The other PR goes into this one. I just wanted you to check this before I merged that one in here.

app/graphql/types/enum/installment_tag.rb
toyhammered: hmm, good catch, me do bad copy pasta.

NuckChorris (approved):
LGTM!

PR #1002, opened 2021-04-15
GraphQL - Update Input Type Generation

toyhammered:
# What

Updating the way input's get generated to not only change CRUD methods, but anything.

# Why

Shit was bad. You can always just update the `graphql_name` directly if you don't want this auto default enabled. This does not modify any of our existing Input's, all of them are the exact same.

# Checklist

- [ ] All files pass Rubocop

- [ ] All the tests pass

### DIFF
--- a/app/graphql/types/input/titles_list.rb
+++ b/app/graphql/types/input/titles_list.rb
@@ -1,6 +1,4 @@
 class Types::Input::TitlesList < Types::Input::Base
-  graphql_name 'TitlesListInput'
-
   argument :canonical, String, required: false
   argument :localized, Types::Map, required: false
   argument :alternatives, [String], required: false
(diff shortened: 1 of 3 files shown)


app/graphql/types/input/titles_list.rb
```diff
@@ -1,4 +1,6 @@
 class Types::Input::TitlesList < Types::Input::Base
+  graphql_name 'TitlesListInput'
```
NuckChorris: Wouldn't these both autogenerate properly now, since this PR fixes the default_graphql_name?

toyhammered: You are entirely right! I was thinking of it differently (needing to be nested).

NuckChorris (approved):

PR #1007, opened 2021-04-28
Clean indices

NuckChorris:
# What

A bunch of migrations cleaning up indices!

The substory indexes are worthless since we no longer use substories (they're only there for archival reasons) and take up almost 600MB each!

image

The index on LibraryEntry (user_id, media_type) is covered by the other index on (user_id, media_type, media_id) so it's worthless, and it takes up 2GB to do absolutely fucking nothing.

image

The indexes on LibraryEvent anime_id/drama_id/manga_id are unused, but may become used in future. But they also waste a lot of space on null values, so we could save a bunch of space by using partial indexes. I expect a roughly 50% reduction in index size for the anime_id and manga_id indexes, and a 100% reduction on the drama_id (since we never finished launching drama lists)

image

Following up on the partial index stuff, I added new partial indexes to LibraryEntries to reduce size on another big table. Since I'm scared to break anything on this core table, though, I've only added them and not removed any existing indexes. I'll follow up later with removal!

# Why
Unused indices have a cost during writes, and take up space on disk

### DIFF
--- a/db/migrate/20210427225015_remove_unused_indices.rb
+++ b/db/migrate/20210427225015_remove_unused_indices.rb
@@ -1,0 +1,21 @@
+class RemoveUnusedIndices < ActiveRecord::Migration[5.2]
+  disable_ddl_transaction!
+  def change
+    # We no longer use substories
+    remove_index :substories, :created_at
+    remove_index :substories, :deleted_at
+    remove_index :substories, :story_id
+    remove_index :substories, :target_id
+    remove_index :substories, :user_id
+    # We already have LibraryEntries (user_id, media_type, media_id) so PG just uses prefix
+    remove_index :library_entries, %i[user_id media_type]
+    # These indexes are extremely space-inefficient and aren't used anyways
+    remove_index :library_events, :manga_id
+    remove_index :library_events, :anime_id
+    remove_index :library_events, :drama_id
+    # Replace them with partial indices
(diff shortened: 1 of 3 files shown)


db/migrate/20210427230857_create_partial_indices_on_library_entries.rb
```diff
@@ -0,0 +1,13 @@
+class CreatePartialIndicesOnLibraryEntries < ActiveRecord::Migration[5.2]
+  disable_ddl_transaction!
+  def change
+    add_index :library_entries, :manga_id,
+      where: 'manga_id IS NOT NULL',
```
toyhammered, line 5: I didn't know you could do partial indexes like this.

db/migrate/20210427225015_remove_unused_indices.rb
```diff
@@ -0,0 +1,21 @@
+class RemoveUnusedIndices < ActiveRecord::Migration[5.2]
+  disable_ddl_transaction!
+  def change
+    # We no longer use substories
+    remove_index :substories, :created_at
+    remove_index :substories, :deleted_at
+    remove_index :substories, :story_id
+    remove_index :substories, :target_id
+    remove_index :substories, :user_id
+    # We already have LibraryEntries (user_id, media_type, media_id) so PG just uses prefix
```
toyhammered, line 10: Didn't know this either.

toyhammered (approved):
Lookin Good!

PR #1010, opened 2021-05-07
Moving to Shrine: Step 1

NuckChorris:
# What

Start the process of moving from Paperclip to Shrine.

Specifically this begins the dual-write phase for ListImports. Once this is merged, I can migrate all the existing data and switch code to use the Shrine fields. Then we can remove Paperclip on this model and repeat for all the other models 😬 

# Why

Paperclip is unmaintained, and the top spot is now ActiveStorage. Sadly, ActiveStorage kinda sucks for public files, and isn't flexible enough for our needs. Shrine is the new #2, and it's way more flexible and can handle public files reliably.

### DIFF
--- a/vendor/paperclip_shrine_synchronization.rb
+++ b/vendor/paperclip_shrine_synchronization.rb
@@ -1,0 +1,80 @@
+# Borrowed from https://shrinerb.com/docs/paperclip#2-dual-write
+# Covered by the MIT license at https://github.com/shrinerb/shrine/blob/master/LICENSE.txt
+# Slightly modified to support paperclip-meta and our own blurhash integration
+
+module PaperclipShrineSynchronization
+  def self.included(model)
+    model.before_save do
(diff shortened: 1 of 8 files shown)


vendor/paperclip_shrine_synchronization.rb
```diff
@@ -0,0 +1,80 @@
...
+      shrine_attachment_file(object)
+    else
+      shrine_style_file(object)
+    end
+  end
+
+  def shrine_attachment_file(attachment)
+    location = attachment.path
+    # if you're storing files on disk, make sure to subtract the absolute path
+    location = location.sub(%r{^#{storage.prefix}/}, "") if storage.prefix
```
toyhammered:
```suggestion
    location = location.sub(%r{^#{storage.prefix}/}, '') if storage.prefix
```

toyhammered (approved):

PR #1030, opened 2021-06-05
GraphQL - reports

toyhammered:
# What

Add report functionality to GraphQL.

# Why

# Checklist

- [ ] All files pass Rubocop

- [ ] All the tests pass

### DIFF
--- a/app/graphql/types/query_type.rb
+++ b/app/graphql/types/query_type.rb
@@ -401,7 +401,7 @@ class Types::QueryType < GraphQL::Schema::Object
   end
 
   def reports
-    ReportPolicy::Scope.new(context[:token], ::Report).resolve
+    ReportPolicy::Scope.new(context[:token], ::Report).resolve.order(created_at: :desc)
   end
 
   field :find_report_by_id, Types::Report, null: true do
--- a/app/graphql/types/query_type.rb
+++ b/app/graphql/types/query_type.rb
@@ -395,4 +395,36 @@ class Types::QueryType < GraphQL::Schema::Object
   #   DESCRIPTION
   #   argument :title, String, required: true
   # end
+
+  field :reports, Types::Report.connection_type, null: true do
+    description 'All Reports in the Kitsu database'
+  end
+
+  def reports
+    ReportPolicy::Scope.new(context[:token], ::Report).resolve
+  end
+
+  field :find_report_by_id, Types::Report, null: true do
+    description 'Find a single Report by ID'
+    argument :id, ID, required: true
+  end
+
+  def find_report_by_id(id:)
+    RecordLoader.for(::Report, token: context[:token]).load(id)
+  end
+
+  field :reports_by_status, Types::Report.connection_type, null: true do
+    description 'Select all Reports that match with a supplied status.'
+    argument :statuses, [Types::Enum::ReportStatus],
+      required: false,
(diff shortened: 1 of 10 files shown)


app/graphql/types/report.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  field :status, Types::Enum::ReportStatus,
+    null: false,
+    description: 'The resolution status for this report'
+
+  field :explanation, String,
+    null: true,
+    description: <<~DESCRIPTION.squish
+      Additional information related to why the report was made.
+      The most relevant results will be at the top.
```
Reinachan: I'm not sure what this line is for? How is relevance related to a report description? Is this line here by accident?

app/graphql/types/union/report_item.rb
```diff
@@ -0,0 +1,5 @@
+class Types::Union::ReportItem < Types::Union::Base
+  description 'Objects which are Reportable'
+
+  possible_types Types::Post, Types::Comment, Types::MediaReaction, Types::Review
```
Reinachan, line 4: Does this mean that you can fetch details about the post/comment/reaction/review through your query? Like doing something like this:

```gql
query reports {
  // stuff
  naughty {
    ... on Post {
      content
      user {
        username
      }
      // more stuff
    }
  }
}
```

Reinachan (approved):
Other than these two questions, LGTM. If the answer to the second question is no, then that would need to be addressed :)

app/graphql/types/report.rb
toyhammered: oops, will remove.

app/graphql/types/union/report_item.rb
toyhammered: yes

app/graphql/types/query_type.rb
```diff
@@ -395,4 +395,36 @@ def franchises
   #   DESCRIPTION
   #   argument :title, String, required: true
   # end
+
+  field :reports, Types::Report.connection_type, null: true do
+    description 'All Reports in the Kitsu database'
+  end
+
+  def reports
+    ReportPolicy::Scope.new(context[:token], ::Report).resolve
```
NuckChorris: You probably want to add a sort to this, even if it's only `.order(created_at: :desc)` 😅

app/graphql/types/review.rb
```diff
@@ -0,0 +1,60 @@
...
+    description: 'Potentially migrated over from hummingbird.'
+
+  field :is_spoiler, Boolean,
+    null: false,
+    description: 'Does this review contain spoilers from the media',
+    method: :spoiler
+
+  field :deleted_at, GraphQL::Types::ISO8601DateTime,
+    null: true,
+    description: 'When this review was deleted'
```
NuckChorris: This doesn't need to be exposed, deleted reviews aren't accessible via the API

app/graphql/types/review.rb
```diff
@@ -0,0 +1,60 @@
...
+
+  field :likes, Types::Profile.connection_type,
+    null: false,
+    description: 'Users who liked this reaction'
```
NuckChorris:
```suggestion
    description: 'Users who liked this review'
```

NuckChorris (approved):
LGTM!

PR #1033, opened 2021-06-27
Wordfiltering

NuckChorris:
# What

- Adds an `InstancedCallbacks` superclass for encapsulating lifecycle callbacks, which handles options way better than the old `Callbacks` superclass and allows internal state across multiple callbacks (gasp!)
- Adds a System user to do stuff as automated systems
- Adds a `hidden_at` field to Posts, Comments, and Reactions which allows shadowbanning them
- Adds a new Rails Admin field type for `flags`, encapsulating an ActiveFlag field
- Adds a wordfilter system

# Why

We want a wordfilter system to help with moderation. So we need two things:
1. A way to hook up to content submission and apply wordfiltering
2. A way for mods to to add/remove/edit wordfilters

For (1) we want to encapsulate a reusable set of callbacks, but we also want to maintain state across before_validation and after_save so that we're not requerying the database. To that end, I built the `InstancedCallbacks` superclass.

For (2) we want to have the ability to specify which scopes the action should apply in, so I built a new `flags` field type.

I also wanted to allow shadowbanning content, since it's helpful for conflict-avoidance, so I added `hidden_at`.

# Checklist

- [x] All files pass Rubocop

- [x] All the tests pass

### DIFF
--- a/app/models/wordfilter.rb
+++ b/app/models/wordfilter.rb
@@ -1,9 +1,10 @@
 class Wordfilter < ApplicationRecord
   flag :locations, %i[post comment reaction]
   enum action: {
-    report: 10,
-    hide: 20,
-    reject: 30
+    censor: 10,
+    report: 20,
+    hide: 30,
+    reject: 40
   }, _prefix: 'action_'
 
   validates :pattern, presence: true
--- a/app/models/wordfilter.rb
+++ b/app/models/wordfilter.rb
@@ -1,0 +1,21 @@
+class Wordfilter < ApplicationRecord
+  flag :locations, %i[post comment reaction]
+  enum action: {
+    report: 10,
+    hide: 20,
+    reject: 30
+  }, _prefix: 'action_'
+
+  validates :pattern, presence: true
+
+  scope :matching, ->(text) {
+    # Match regexes with ~* and create an exact match via ILIKE
+    where("? ~* ('.*' || pattern || '.*')", text).where(regex_enabled: true).or(
+      where("? ILIKE ('%' || pattern || '%')", text).where(regex_enabled: false)
+    )
+  }
+
+  def self.action_for(location, text)
+    matching(text).where_locations(location).order(action: :desc).first&.action&.to_sym
+  end
+end
--- a/app/services/wordfilter_service.rb
+++ b/app/services/wordfilter_service.rb
@@ -1,0 +1,40 @@
+class WordfilterService
+  def initialize(location, text)
+    @location = location
+    @text = text
+  end
+
+  def reject?
+    wordfilters[:reject].present?
+  end
+
+  def hide?
+    wordfilters[:hide].present?
+  end
+
+  def report?
+    wordfilters[:report].present?
+  end
+
+  def censor?
+    wordfilters[:censor].present?
+  end
+
+  def censored_text
+    wordfilters[:censor].inject(@text) do |text, wordfilter|
+      # Convert SQL LIKE roughly into Regex
+      pattern = if wordfilter.regex_enabled? then wordfilter.pattern
+                else wordfilter.pattern.gsub('.', '\.').gsub('%', '.*').tr('_', '.')
+                end
+      pattern = Regexp.new(pattern, Regexp::IGNORECASE)
+
+      text.gsub(pattern, 'CENSORED')
+    end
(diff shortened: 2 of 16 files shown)


spec/callbacks/instanced_callbacks_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+    it 'attaches a callback with the given name' do
+      klass = class_spy(ActiveRecord::Base)
+      described_class.attach_callback(klass, :before_save)
+      expect(klass).to have_received(:before_save)
+    end
+  end
+
+  describe '.wrap_callback' do
+    it 'wraps a method call to be bound to the instance' do
+      callbacks = spy(described_class)
```
NuckChorris: Can't use them for either of these

app/views/rails_admin/main/_form_flags.html.haml
```diff
@@ -0,0 +1,5 @@
+= form.collection_check_boxes(field.method_name, field.pairs, :last, :first) do |box|
+  %div{ class: "checkbox" }
```
NuckChorris: I don't give a shit it's HAML, it's a trash fire regardless

app/views/rails_admin/main/_form_flags.html.haml
```diff
@@ -0,0 +1,5 @@
+= form.collection_check_boxes(field.method_name, field.pairs, :last, :first) do |box|
```
NuckChorris: Fuck off it's HAML

config/initializers/rails_admin.rb
```diff
@@ -240,25 +244,49 @@ def edit_user_link
...
+  config.model('Post') { navigation_label 'Social' }
   config.model('Comment') { parent Post }
   config.model('Repost') { parent Post }
   config.model('Upload') { parent Post }
+  config.model('Wordfilter') do
+    weight(-5)
+    navigation_label 'Social'
+    object_label_method { :pattern }
+    field :pattern, :string do
+      help <<~HELP.squish.html_safe
```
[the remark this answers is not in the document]
NuckChorris: Not a risk here, also the best option

NuckChorris:
Pulled some of this directly into the main branch, and then split a bunch out into #1038 and #1039

Once those are reviewed and merged, I'll rebase this and it'll be in a reviewable state

app/models/wordfilter.rb
```diff
@@ -0,0 +1,22 @@
...
+
+  scope :matching, ->(text) {
+    # Match regexes with ~* and create an exact match via ILIKE
+    where("? ~* ('.*' || pattern || '.*')", text).where(regex_enabled: true).or(
+      where("? ILIKE ('%' || pattern || '%')", text).where(regex_enabled: false)
+    )
+  }
+
+  def self.action_for(location, text)
+    matching(text).where_locations(location).order(action: :desc).first&.action&.to_sym
```
toyhammered, line 20: why are you using `where_locations` instead of `where(locations: location)`?

app/services/wordfilter_service.rb
```diff
@@ -0,0 +1,40 @@
...
+  def censor?
+    wordfilters[:censor].present?
+  end
+
+  def censored_text
+    wordfilters[:censor].inject(@text) do |text, wordfilter|
+      # Convert SQL LIKE roughly into Regex
+      pattern = if wordfilter.regex_enabled? then wordfilter.pattern
+                else wordfilter.pattern.gsub('.', '\.').gsub('%', '.*').tr('_', '.')
+                end
```
toyhammered, line 28: 🤮

toyhammered (approved):
LGTM! Do we want to expose any of this in the graphql?

app/models/wordfilter.rb
NuckChorris: Because active flag doesn't override where() it instead exposes this helper method

app/services/wordfilter_service.rb
NuckChorris: Lmao right? It's nasty but I want to make shit work the same across db and rails 😭

NuckChorris:
I don't think that's necessary right now. In future when we are fully phasing out Rails Admin in favor of our own custom admin panels we'll want it though

PR #1038, opened 2021-06-27
Add flags field type to rails admin

NuckChorris:
Adds a flags field type to rails admin to allow editing ActiveFlag bitfields

The code is kinda shite but that's normal for Rails Admin stuff


NuckChorris:
Ignore hound, this is garbage code, I don't care about style

toyhammered (approved):

PR #1065, opened 2021-08-29
Add posts field to media in GraphQL

Reinachan:
# What

Add a posts field to the GraphQL API allowing for fetching just the posts directly.

# Why

Expand the API for more usecases.

# Checklist

- [x] All files pass Rubocop

- [ ] All the tests pass

### DIFF
--- a/app/graphql/types/interface/media.rb
+++ b/app/graphql/types/interface/media.rb
@@ -166,7 +166,8 @@ module Types::Interface::Media
   def posts(sort: [{ on: :created_at, direction: :asc }])
     Loaders::PostsLoader.connection_for({
       find_by: :media_id,
-      sort: sort
+      sort: sort,
+      where: { media_type: type },
     }, object.id)
   end
 
--- a/app/graphql/types/interface/media.rb
+++ b/app/graphql/types/interface/media.rb
@@ -167,7 +167,7 @@ module Types::Interface::Media
     Loaders::PostsLoader.connection_for({
       find_by: :media_id,
       sort: sort,
-      where: { media_type: type },
+      where: { media_type: type }
     }, object.id)
   end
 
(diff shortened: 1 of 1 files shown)


Reinachan:
I am unable to test this properly myself. It doesn't error when fetching but because of the current environment I can't create posts to test it. It *should* work I think at least.

app/graphql/types/interface/media.rb
```diff
@@ -158,6 +158,18 @@ def reactions
...
 
+  field :posts, Types::Post.connection_type, null: false do
+    description 'All posts that tag this media.'
+    argument :sort, Loaders::PostsLoader.sort_argument, required: false
+  end
+
+  def posts(sort: [{ on: :created_at, direction: :asc }])
+    Loaders::PostsLoader.connection_for({
+      find_by: :media_id,
+      sort: sort
```
toyhammered: you'll need a `where: { media_type: type }` also to scope it correctly. (`type` is defined as a method somewhere above)

toyhammered (changes requested):

app/graphql/types/interface/media.rb
Reinachan: Sorry, been out of it for a couple of days. Will look into this once I got time which might be later today or it might be on Sunday

Reinachan: Fixed

toyhammered:
> I am unable to test this properly myself. It doesn't error when fetching but because of the current environment I can't create posts to test it. It _should_ work I think at least.

have you been able to test this yet?

Reinachan:
> have you been able to test this yet?

No. Emma told me that feeds don't work locally anyway, so I really have no idea how I'd be able to test it. Stream doesn't work locally. I can mention you in a thread where I talked about it over on Slack

NuckChorris (approved):

PR #1066, opened 2021-09-05
GraphQL - add media characters fancy loader

toyhammered:
# What

# Why

# Checklist

- [ ] All files pass Rubocop

- [ ] All the tests pass

### DIFF
--- a/app/graphql/loaders/media_characters_loader.rb
+++ b/app/graphql/loaders/media_characters_loader.rb
@@ -1,0 +1,6 @@
+class Loaders::MediaCharactersLoader < Loaders::FancyLoader
+  from MediaCharacter
+
+  sort :created_at
+  sort :updated_at
+end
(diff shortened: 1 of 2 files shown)


app/graphql/loaders/media_characters_loader.rb
```diff
@@ -0,0 +1,6 @@
+class Loaders::MediaCharactersLoader < Loaders::FancyLoader
+  from MediaCharacter
+
+  sort :created_at
+  sort :updated_at
```
NuckChorris, line 5: can you add a sort for role? We want to display main characters at the top

NuckChorris (changes requested):
Looks pretty good just one tweak

app/graphql/loaders/media_characters_loader.rb
toyhammered: hah, was thinking of adding that just wasn't sure.

PR #1067, opened 2021-09-06
GraphQL - Add myLibraryEntry

toyhammered:
# What

# Why

# Checklist

- [ ] All files pass Rubocop

- [ ] All the tests pass

### DIFF
--- a/app/graphql/types/interface/media.rb
+++ b/app/graphql/types/interface/media.rb
@@ -110,7 +110,12 @@ module Types::Interface::Media
       raise GraphQL::ExecutionError, 'You must be authorized to view your library entry'
     end
 
-    RecordLoader.for(LibraryEntry, column: :user_id, where: { media_id: object.id, media_type: type }, token: context[:token]).load(User.current.id)
+    RecordLoader.for(
+      LibraryEntry,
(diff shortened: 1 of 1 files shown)


app/graphql/types/interface/media.rb
```diff
@@ -101,6 +101,23 @@ def titles
...
+
+  def my_library_entry
+    if context[:token].blank?
+      raise GraphQL::ExecutionError, 'You must be authorized to view your library entry'
+    end
+
+    RecordLoader.for(
+      LibraryEntry,
+      column: :user_id,
+      where: { media_id: object.id, media_type: type },
```
NuckChorris, line 116: Should this be `object.type` or does `type` work too?

toyhammered: so `object.type` doesn't actually work because Anime/Manga doesn't have a `type`. There is custom code above in the `Media` interface
```
  # Types::Anime -> Anime
  def type
    self.class.name.split('::').last
  end
 ```
 which allows us to use our interface instead of copy/pasta in all the main types.

NuckChorris (approved):

PR #1086, opened 2021-10-10
GraphQL - update libary_entry create mutation to assume user_id

toyhammered:
# What

The libary_entry create mutation should not accept a user_id as a parameter. We should be able to assume this via the session.

# Why

# Checklist

- [ ] All files pass Rubocop

- [ ] All the tests pass


NuckChorris (approved):
Yayyyyy LGTM thanks

NuckChorris:
Wait, looks like one of the specs for graphql broke from this

PR #1123, opened 2021-10-31
Integrate with graphql-fancy_loader gem

toyhammered:
Modify all loaders to properly inherit from the correct class

fix transform to accept a required context (even if unused)

# What

This is a full integration with our new graphql-fancy_loader gem. I have done some basic testing but will do some additional testing once in dev.

# Why

# Checklist

- [ ] All files pass Rubocop

- [ ] All the tests pass


app/graphql/loaders/follows_loader.rb
```diff
@@ -1,13 +1,13 @@
...
   # Sort by whether you follow the followed person
   sort :following_followed,
-    transform: ->(ast) {
+    transform: ->(ast, context) {
       their_follows = Follow.arel_table
       your_follows = Follow.arel_table.alias('yours')
 
       condition = your_follows[:followed_id].eq(their_follows[:followed_id]).and(
-        your_follows[:follower_id].eq(User.current&.id)
+        your_follows[:follower_id].eq(context[:token]&.id)
```
NuckChorris: Tokens != Users

I recommend setting `context[:user] = token&.resource_owner` in the controller so that we don't have to look this up repeatedly

NuckChorris (approved):

PR #1140, opened 2021-11-30
Server-side Title Logic

NuckChorris:
# What

Implement title logic on server-side in GraphQL API

# Why

Putting this logic on the client side has shown some issues. We thought it would be good for caching, but well... we never really cached it anyways, so we never reaped that benefit, and it's marginal at best!

### Pros:

 - Third-party clients don't have to handle this, they get it automatically when the user is logged in
 - First-party clients don't have to redo this either, they just display a single string
 - We only have to load a single title in most places!

### Cons:

- Needs to be cleared from the cache when the setting is changed
- Can't cache server-side (not that we do this anyways)


app/graphql/types/titles_list.rb
```diff
@@ -3,16 +3,54 @@ class Types::TitlesList < Types::BaseObject
...
+    null: true,
+    description: 'The locale code that identifies which title is used as the original title'
+
+  field :preferred,
+    String,
+    null: false,
+    description: "The title that best matches the user's preferred settings"
+
+  def preferred
+    object.first_title_for(context[:user]&.title_preference_list || %i[canonical])
```
toyhammered: no `current_user` defined in the base somewhere?

app/models/titles_list.rb
```diff
@@ -0,0 +1,94 @@
...
+    if @titles.key?(:en_cn) then :en_cn
+    elsif @titles.key?(:en_kr) then :en_kr
+    elsif @titles.key?(:en_jp) then :en_jp
+    end
+  end
+
+  # @return [Symbol, nil] the key of the original title of the media in the localized titles hash
+  def original_locale
+    # TODO: we should aim to resolve all cases where this is nil or uses the fallback logic
+    preferred_key = "#{@original_languages.first}_#{@original_countries.first}".to_sym
```
toyhammered, line 47: most likely worth just extracting this out into a private method.

app/models/user.rb
```diff
@@ -330,6 +330,14 @@ def one_signal_player_ids
     one_signal_players.pluck(:player_id).compact
   end
 
+  def title_preference_list
+    case title_language_preference
```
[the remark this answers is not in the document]
toyhammered: This isn't a terrible suggestion for rubocop lol.

toyhammered (approved):
LGTM just a few small nitpicks.

app/graphql/types/titles_list.rb
NuckChorris: Y'know... that would probably be a good idea...

app/models/titles_list.rb
NuckChorris: It's used once and it's the only piece of logic here which *isn't* about dealing with missing data (and thus will be resolved once we fill the data in)

I don't think it's necessary to extract

app/models/user.rb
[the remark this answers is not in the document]
NuckChorris: I definitely considered it, but I think this is easier to read than a referece to a constant 300 lines up there :point_up: so I opted to keep it this way. I could also envision a future where the user configures this array themselves by drag-and-drop

PR #1242, opened 2022-05-13
Fancy Mutations, Mutators, and Account Creation

NuckChorris:
# What
Despite the small diff, there's actually three new things in this single PR. It's a work-in-progress, but I want to demonstrate the new concepts early and get feedback.

## FancyMutation
- A standardized way of structuring mutations
- A set of class methods to make that structure easy to adhere to
- **TODO:** provide an easy way to bind GraphQL error types to Ruby error types

## Mutators
- Plain-ass ruby classes
- Holds business logic like a controller
- Leaves any serialization/deserialization to the GraphQL layer
  - In future, I might make a shorthand for hooking this up to a FancyMutation with no de/serialization required

## `Account.create` mutation
- Allows creating an account
- Rate limited to 20 attempts in an hour or 20 attempts in 2 days.
  - **TODO:** limit *successful* mutations, so we can drop the limit significantly
- **TODO:** set up error handling

# Why

## FancyMutation
- Encourage consistency and adherence to best practices
  - Errors in GraphQL via an interface+union approach
  - Inputs as an object instead of separate parameters
- Make it easy to do input objects instead of putting your parameters in a separate file in a completely weird separate folder

## Mutators
- My previous Actions system was frankly just way overcomplicated and stupid
- I wanted something which could encapsulate and document business logic separate from the API layer
- Allow testing shit or performing actions from the console without parsing GraphQL
- Fuck it, class methods will do the trick

## `Account.create` mutation
- Demonstrate FancyMutation and Mutators
- Allow creating an account from V4

### DIFF
--- a/app/graphql/concerns/fancy_mutation.rb
+++ b/app/graphql/concerns/fancy_mutation.rb
@@ -30,6 +30,8 @@
 module FancyMutation
   extend ActiveSupport::Concern
 
+  class WarningsPresent < StandardError; end
+
   module PrependedMethods
     # The return-driven approach of #ready? and #authorized? is garbage, so we override it and allow
     # using the same error system as in #resolve. To achieve this, we prepend a module wrapping the
@@ -111,6 +113,10 @@ module FancyMutation
         end
         field :warnings, [warnings_union], null: true
 
+        # Add the ignore_warnings argument and error type
+        input { argument :ignore_warnings, GraphQL::Types::Boolean, required: false }
+        errors Types::Errors::WarningsPresent
+
         warnings_union
       end
     end
@@ -142,15 +148,24 @@ module FancyMutation
   # as the result type.
   # @param ignore_warnings [Boolean] Whether to ignore warnings or not
   def resolve_with_support(ignore_warnings: false, **args)
-    result = super(**args)
+    # Wrap the mutation in a transaction to allow for rollback if there are warnings
+    ApplicationRecord.transaction(requires_new: true) do
+      result = super(**args)
 
-    raise ActiveRecord::Rollback if warnings.present? && !ignore_warnings
+      # Trigger a rollback but allow us to catch it afterwards and control our response format.
+      raise WarningsPresent if warnings.present? && !ignore_warnings
 
+      {
+        # If the mutation returns the errors list, ignore it (it's not the actual result)
+        result: (result unless result == errors),
+        warnings: warnings,
+        errors: errors
+      }
+    end
+  rescue WarningsPresent
     {
-      # If the mutation returns the errors list, ignore it (it's not the actual result)
-      result: (result unless result == errors),
       warnings: warnings,
-      errors: errors
+      errors: [*errors, Types::Errors::WarningsPresent.build]
     }
   end
 
--- a/app/graphql/concerns/fancy_mutation.rb
+++ b/app/graphql/concerns/fancy_mutation.rb
@@ -48,31 +48,59 @@ module FancyMutation
       argument :input, input, required: true
     end
 
-    # Sets up the union of error types for the mutation
+    # Adds types to the union of error types for the mutation
     # @param types [Array<Class>] The error types this can return
     def errors(*types)
-      union_name = "#{graphql_name}ErrorsUnion"
-      union = Class.new(Types::Union::Base) do
-        graphql_name union_name
-        possible_types(*types)
-
-        def self.resolve_type(object, _context)
-          object.key?(:__type) ? object[:__type] : super
-        end
-      end
-      field :errors, [union], null: true
+      errors_union.possible_types(*types)
     end
 
-    # Sets up the union of warning types for the mutation
+    # Adds types to the union of warning types for the mutation
     # @param types [Array<Class>] The warning types this can return
     def warnings(*types)
-      union_name = "#{graphql_name}WarningsUnion"
-      union = Class.new(Types::Union::Base) do
-        graphql_name union_name
-        possible_types types.flatten
-      end
-      field :errors, [union], null: true
+      warnings_union.possible_types(*types)
     end
+
+    # These two methods are kinda strange. The first time they're called, they create a new union
+    # type, hook it up to the mutation, and return it. When they get called again, they return that
+    # same type, allowing us to add more possible types to the union from different locations. This
+    # is important for utilities which might hook the `#ready?` method, since they can return
+    # errors.
+    def errors_union
+      @errors_union ||= begin
+        union_name = "#{graphql_name}ErrorsUnion"
+        errors_union = Class.new(Types::Union::Base) do
+          graphql_name union_name
+
+          def self.resolve_type(object, _context)
+            object.key?(:__type) ? object[:__type] : super
+          end
+        end
+        field :errors, [errors_union], null: true
+
+        errors_union
+      end
+    end
+
+    def warnings_union
+      @warnings_union ||= begin
+        union_name = "#{graphql_name}WarningsUnion"
+        warnings_union = Class.new(Types::Union::Base) do
+          graphql_name union_name
+
+          def self.resolve_type(object, _context)
+            object.key?(:__type) ? object[:__type] : super
+          end
+        end
+        field :warnings, [warnings_union], null: true
+
+        warnings_union
+      end
+    end
+  end
+
+  included do
+    private_class_method :errors_union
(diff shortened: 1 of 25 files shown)


app/graphql/concerns/fancy_mutation.rb
```diff
@@ -0,0 +1,52 @@
...
+        possible_types types.flatten
+      end
+      field :errors, [union], null: true
+    end
+  end
+
+  def resolve_with_support(ignore_warnings: false, **args)
+    result = super(**args)
+
+    raise ActiveRecord::Rollback unless warnings.blank? || ignore_warnings
```
toyhammered: BAD BAD BAD! `unless ||` is cursed!

app/graphql/mutations/account/create.rb
```diff
@@ -0,0 +1,35 @@
...
+      description: 'The password for the user'
+
+    argument :external_identity, Types::Input::Account::ExternalIdentity,
+      required: false,
+      description: 'An external identity to associate with the account on creation'
+  end
+  result Types::Account
+
+  def resolve(input:)
+    AccountMutator.create(**input)
```
toyhammered, line 34: Just thinking a bit about this service object and are we gonna mix n match between doing class methods, and/or having these service object be something like `AccountCreator.new(**).call` or w/e.

app/graphql/concerns/fancy_mutation.rb
NuckChorris: The equivalent `if` would be `if warnings.present? && !ignore_warnings` which I guess works

app/graphql/mutations/account/create.rb
NuckChorris: Class methods only. If you need to instantiate, just use a regular Service object

app/graphql/mutations/account/create.rb
```diff
@@ -0,0 +1,42 @@
...
+  errors Types::Errors::Validation
+
+  def resolve(input:)
+    AccountMutator.create(**input)
+  rescue ActiveRecord::RecordInvalid => e
+    validation_errors = Types::Errors::Validation.for_record(e.record, transform_path: ->(path) {
+      path = path.map { |p| p == 'password_digest' ? 'password' : p }
+      ['input', *path]
+    })
+    errors.push(*validation_errors)
```
NuckChorris, line 40: I would be lying if I said I was happy about this part, but it solves the `password_digest` name leaking out as well as the `input` prefix. I might add a shortcut for the prefix in future, it seems like it'll be quite common (most models shouldn't have the weird `password_digest` bullshit)

app/graphql/concerns/fancy_mutation.rb
toyhammered: yeah one of the few times I am okay with `!method`.

app/graphql/mutations/account/create.rb
```diff
@@ -0,0 +1,42 @@
...
+    argument :password, String,
+      required: true,
+      description: 'The password for the user'
+
+    argument :external_identity, Types::Input::Account::ExternalIdentity,
+      required: false,
+      description: 'An external identity to associate with the account on creation'
+  end
+  result Types::Account
+  errors Types::Errors::Validation
```
toyhammered, line 31: i'm thinking this should be called `union_errors` just to be explicit about it. When I saw this the first thing I asked was "isn't this for unions" and had to check the class.

app/graphql/mutations/account/create.rb
toyhammered: couldn't we just override the `graphql_name` for that error so it can still be `password_digest` but just show as `password` instead of having to have the logic here?

toyhammered:
I am loving this, I think there should be some type of documentation or something explaining how to add a new mutation and what all the steps are to be done.

app/graphql/mutations/account/create.rb
NuckChorris: All of these are named after their fields, I think I would rather stay consistent

app/graphql/mutations/account/create.rb
NuckChorris: Afraid not. It's the `path` we're overriding, not the GraphQL name. And only applies to this one stupid context.

NuckChorris (commented):
Yeah I'm gonna add docs for this now that I like the feel of it!

app/graphql/concerns/fancy_mutation.rb
```diff
@@ -0,0 +1,106 @@
...
+#     end
+#     # Your result type should generally expose what changed so clients can update their cache
+#     result Types::Post
+#     # Build the union of errors which you may return
+#     errors Types::Errors::NotAuthorized,
+#       Types::Errors::NotAuthenticated,
+#       Types::Errors::Conflict
+#
+#     def resolve(post_id:)
+#       return errors << Types::Errors::NotAuthenticated.build unless context[:current_user]
```
toyhammered, line 21: would be nice if we tied this into the mutation class itself (or something via pundit) similar to how you did the `input do`.

NuckChorris: There's already a mixin for authenticated mutations I think, but you're right that it should get cleanly integrated into this. I also like the idea of an easy Pundit shorthand, like `authorize PostPolicy, :like?` in the class body, creating a `def ready?` hook 🤔

NuckChorris:
I should mention that I'm planning to add tests for this now that the interface is pretty much hammered out

app/graphql/concerns/fancy_mutation.rb
```diff
@@ -0,0 +1,172 @@
...
+
+        warnings_union
+      end
+    end
+  end
+
+  included do
+    prepend PrependedMethods
+    private_class_method :errors_union
+    private_class_method :warnings_union
```
toyhammered, line 168: not seen this approach in a while.

NuckChorris: It's hella janky that I can't put this in class_methods but at least it still achieves what I want

app/graphql/concerns/fancy_mutation.rb
```diff
@@ -0,0 +1,196 @@
...
+  class WarningsPresent < StandardError; end
+
+  module PrependedMethods
+    # Wraps the resolve method with some support. When ignore_warnings is false or not set, it will
+    # rollback when there are warnings. In all cases, the return value of the resolve method will be
+    # wrapped as the :result field, so that the mutation always returns the correct structure. When
+    # the return value is the errors list, we ignore it, so that we won't accidentally try to treat
+    # it as the result type.
+    # @param ignore_warnings [Boolean] Whether to ignore warnings or not
+    def resolve(ignore_warnings: false, **args)
```
NuckChorris, line 42: I switched this to a prepended `#resolve` instead of `#resolve_with_support` so that we're not touching any private interfaces. Additionally, this should ensure that code in `#ready?` and `#authorized?` will run outside this transaction, which is important for rate-limiting (or would be, if we used Postgres as the backing store for that)

app/graphql/concerns/fancy_mutation.rb
```diff
@@ -0,0 +1,196 @@
...
+
+        # Add the ignore_warnings argument and error type
+        input { argument :ignore_warnings, GraphQL::Types::Boolean, required: false }
+        errors Types::Errors::WarningsPresent
+
+        warnings_union
+      end
+    end
+
+    def input_type
```
NuckChorris, line 152: This makes the input type reopenable, so we can call it multiple times to add more arguments

app/graphql/concerns/fancy_mutation.rb
```diff
@@ -0,0 +1,196 @@
...
+  end
+
+  # Adds an error if the current user is not authorized to perform the action on the object.
+  # @param object [ActiveRecord::Base] The object to check authorization on
+  # @param action [Symbol] The action to check authorization for
+  def authorize!(object, action, policy: Pundit::PolicyFinder.new(object).policy)
+    authorized = policy.new(current_token, object).public_send(action)
+    errors << Types::Errors::NotAuthorized.build(object: object, action: action) unless authorized
+    true
+  end
```
NuckChorris, line 185: These two methods are helpers for use in `ready?` and `authorized?` methods, to make it easy

NuckChorris (commented):
I think it's pretty much done now, if you wanna give it a final look

PR #1297, opened 2022-12-15
Account Mutations for GraphQL

shomykohai:
# What

Added mutations related to Account & Profile of a User:
- Account mutations (name, slug, sfw filter)
- Profile mutations (birthday, gender, about..)
- Profile Links
- Favorites (Characters, Anime, Manga..)
- Blocks (Block or unblock someone)

# Why

To make this mutations avaiable on graphql for V4

# Checklist

- [x] All files pass Rubocop

- [x] All the tests pass


NuckChorris:
Do you think you could refactor these on top of the `FancyMutation` stuff like I did for [the MediaReaction mutations](https://github.com/hummingbird-me/kitsu-server/blob/the-future/app/graphql/mutations/media_reaction/create.rb)? Should make the code a lot nicer, it handles all the common structural stuff, pulls the input and output types into the mutation file, and provides a standard for how to deal with errors!

It should be pretty easy, you can splat the keyword params from the input object into the user like this:

```ruby
def resolve(**input)
  current_user.update!(**input)
end
```

or manually assign params through, like this:

```ruby
def resolve(blocked_id:)
  Block.create!(
    user: current_user,
    blocked_id:
  )
end
```

Also, it's now super easy to write tests for graphql stuff, using `with graphql helpers` as you can see [here](https://github.com/hummingbird-me/kitsu-server/blob/the-future/spec/graphql/mutations/media_reaction/create_spec.rb), so if you could add tests that would be ideal, so we know everything works as it should 🎉

NuckChorris:
Oh yeah also if you wouldn't mind, the profile mutations should probably be in a separate Profile namespace. The GraphQL API tries to maintain a solid separation between "Account" (private) and "Profile" (public) stuff 😅

app/graphql/mutations/account/update.rb
```diff
@@ -0,0 +1,58 @@
...
+
+  input do
+    argument :name,
+      String,
+      required: false,
+      description: 'The display name of the user'
+    argument :slug,
+      String,
+      required: false,
+      description: 'The slug (@username) of the user'
```
NuckChorris: These should probably be on the `Profile::Update` since they're not on the `Account` type!

app/graphql/mutations/block/create.rb
```diff
@@ -0,0 +1,32 @@
...
+    return errors << Types::Errors::NotFound.build(path: %w[input blocked_id]) if @block_user.nil?
+    true
+  end
+
+  def resolve(blocked_id:, **)
+    @block = Block.new(
+      user_id: current_user.id,
+      blocked_id:
+    )
+    authorize!(@block, :create?)
```
NuckChorris, line 27: You need to add `NotAuthorized` as a potential error, and probably want to move this up into the `ready?` method

app/graphql/mutations/favorite/create.rb
```diff
@@ -0,0 +1,46 @@
...
+    true
+  end
+
+  def resolve(item_id:, item_type:, **)
+    @favorite = Favorite.new(
+      item_type:,
+      item_id:,
+      user_id: current_user.id
+    )
+    authorize!(@favorite, :create?)
```
NuckChorris, line 41: Again, probably want to put this in `ready?` and add `NotAuthorized` as a possible error!

app/graphql/mutations/profile/update.rb
```diff
@@ -0,0 +1,47 @@
...
+
+class Mutations::Profile::Update < Mutations::Base
+  include FancyMutation
+
+  description 'Update profile'
+
+  input do
+    argument :id, ID,
+      required: true,
+      description: 'Your ID or the one of another user.'
```
NuckChorris: I think let's make this optional, it's only useful to admins, and the default can just be `current_user`

app/graphql/mutations/profile_link/create.rb
```diff
@@ -0,0 +1,46 @@
...
+    Types::Errors::Validation
+
+  def ready?(**)
+    authenticate!
+    return errors << Types::Errors::NotAuthenticated.build if current_user.nil?
+    profile_link = ProfileLink.find_by(
+      profile_link_site_id: profile_link,
+      user_id: current_user.id
+    )
+    raise GraphQL::ExecutionError, 'You already have this profile link.' if profile_link
```
NuckChorris: A better way to handle this is probably to add a `validates :profile_link_site, uniqueness: { scope: :user }` on the ProfileLink model, and rely on `Types::Errors::Validation` to handle this

app/graphql/mutations/profile_link/delete.rb
```diff
@@ -0,0 +1,30 @@
...
+
+class Mutations::ProfileLink::Delete < Mutations::Base
+  include FancyMutation
+
+  description 'Delete a profile link'
+
+  input do
+    argument :profile_link_id, ID,
+      required: true,
+      description: 'The profile link to delete'
```
NuckChorris: Using id here vs ProfileLinkSite in update makes me think we should just choose one or the other to "identify" the link. I'm kinda favoring just ProfileLinkSite, so that clients don't have to find the ID before deletion

app/graphql/types/block.rb
```diff
@@ -0,0 +1,20 @@
...
+    description: 'User who blocked.',
+    method: :user
+
+  field :blocked_user, Types::Profile,
+    null: false,
+    description: 'User who got blocked.'
+
+  def blocked_user
+    Loaders::RecordLoader.for(User).load(object.blocked_id)
+  end
```
NuckChorris, line 22: You probably also want this for `user` (not that it's likely to ever be navigated the other direction, really just in case)

app/graphql/types/query_type.rb
```diff
@@ -9,6 +9,14 @@ def current_account
     User.current
   end
 
+  field :current_profile, Types::Profile, null: true do
+    description 'Kitsu account details. You must supply an Authorization token in header.'
```
NuckChorris: Could you change the wording here?

```suggestion
    description 'Your Kitsu profile. You must supply an Authorization token in header.'
```

app/graphql/types/enum/favorite_item.rb
```diff
@@ -0,0 +1,6 @@
+class Types::Enum::FavoriteItem < Types::Enum::Base
```
NuckChorris: Hmmm, could you rename this to `FavoriteType`? I think it would better express what this is, and allow us to use it elsewhere (ie, in queries like `user { favorites(type: ANIME) }`

NuckChorris (changes requested):

app/graphql/mutations/account/update.rb
shomykohai: I was thinking the same, but then I saw how in the client these fields are in the account settings

image

Are these option in the new react client gonna be moved to the profile page or it does not matter (regarding on how the client ask the changes to the api)?

app/graphql/mutations/profile/update.rb
```diff
@@ -0,0 +1,59 @@
...
+      current_user
+    end
+    return errors << Types::Errors::NotFound.build if @profile.nil?
+    authorize!(@profile, :update?)
+    true
+  end
+
+  def resolve(**input)
+    @profile.update!(**input)
+    @profile.tap(&:save!)
```
NuckChorris: This saves twice, you probably want to use `assign_attributes` and then `save!` or just only use `update!`

Here's a cheat sheet on these methods:

https://scottbartell.com/2020/01/30/set-attributes-in-active-record-rails-6/

app/graphql/types/query_type.rb
```diff
@@ -442,4 +450,21 @@ def reports_by_status(statuses: nil, sort: [{ on: :created_at, direction: :asc }
...
+
+  field :find_block_by_id, Types::Block, null: true do
+    description 'Find a single Block by ID'
+    argument :id, ID, required: true
+  end
+
+  def find_block_by_id(id:)
+    Loaders::RecordLoader.for(::Block, token: context[:token]).load(id)
+  end
+
```
NuckChorris: Is this actually useful for any purpose? We should only add fields that we know will be useful, because it's a lot easier to add something than remove it 😉

NuckChorris (commented):
Some minor comments, but otherwise it looks good and I think we should be ready to merge after this!

I'm working on changing the `authenticate!` and `authorize!` methods to use a raised exception to end execution, but this will all continue to work so we can merge this first and I can change that later :+1:

app/graphql/types/query_type.rb
shomykohai: When I first added this I thought about some use (that unfortunately i don't remember atm).
But i agree that now it does not have purpose or use, I'll remove it!

shomykohai:
Just made the changes!
I also removed the double save from profile links

NuckChorris (approved):
Looks great, let's gooooo!

PR #1298, opened 2022-12-19
Fix library entry media field

shomykohai:
It was not giving back the media if it was NSFW and the sfw filter was turned off

# What
Fix media not being given back in library entry query if it was NSFW.

# Why
It throws an error whenever you try to query the entry

STEPS:
* Add a R18 media in your library on kitsu website as COMPLETED
* Go on [kitsu playground](https://kitsu.io/api/playground)
* Write the following query 
```graphql
query session{
  currentAccount {
    profile{
      library{
        completed(mediaType: ANIME, first: 100) {
          nodes {
            nsfw
            media{
              id
            }
          }          
        }
      }
    }
  }
}
```
and add the Authorization Header
* Query

It should return this error
image

# Checklist

- [ ] All files pass Rubocop

- [x] All the tests pass

### DIFF
--- a/app/graphql/types/library_entry.rb
+++ b/app/graphql/types/library_entry.rb
@@ -14,7 +14,10 @@ class Types::LibraryEntry < Types::BaseObject
     description: 'The media related to this library entry.'
 
   def media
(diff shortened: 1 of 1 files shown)


NuckChorris:
Thanks so much for fixing this!

PR #1459, opened 2023-10-29
GraphQL: Return all volume table data. Add description field

### LINKED ISSUE #1446: GraphQL: Add description to Volume type and expose thumbnails
Few, if any, chapters will have an official description and so relies on users to create them. Volumes on the other hand, will usually have descriptions created by the publisher.

Similarly to the above and it seems that the DB has an image field for volumes but isn't exposed via GQL.

mizaki:
First PR here. Not sure on the checklist, should I mark as done those that aren't required or remove them?

# What

Adds a description field to `volumes` table.
Returns additional fields: thumbnail and (new) description

Closes #1446 

# Why

"Few, if any, chapters will have an official description and so relies on users to create them. Volumes on the other hand, will usually have descriptions created by the publisher.

Similarly to the above and it seems that the DB has an image field for volumes but isn't exposed via GQL."

# Checklist

- [ ] All files pass Rubocop

- [ ] All the tests pass

### DIFF
--- a/app/graphql/types/volume.rb
+++ b/app/graphql/types/volume.rb
@@ -1,24 +1,8 @@
 class Types::Volume < Types::BaseObject
+  implements Types::Interface::Unit
   implements Types::Interface::WithTimestamps
 
   description 'A manga volume which can contain multiple chapters.'
-
-  field :id, ID, null: false
-
-  field :titles, Types::TitlesList,
-    null: false,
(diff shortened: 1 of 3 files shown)


shomykohai:
Seems good to me!
Sorry for the previous review, I haven't seen that you used the Unit interface at first 😅

PR #1465, opened 2023-11-13
Add new mutations and refactor old ones


shomykohai:
# What
Refactor old Mutations on top of FancyMutation.
Add new mutations to GraphQL:
- Posts
    - Delete
    - Edit
    - Follow
    - Unfollow
    - Like
    - Unlike
- Comments
    - Create
    - Delete
    - Edit
    - Like
    - Unlike

Add new Types:
* Embed
* Attachment (Uploads)
* PostFollow
* PostLike
* CommentLike

Add `AttachmentsLoader`

Removed old Manga & Anime mutations in favor of WikiSubmissions

# Why

# Checklist

- [ ] All files pass Rubocop

- [ ] All the tests pass


shomykohai:
There should also be a validation in app/models/post.rb to first check if the unit id belongs to the media so we avoid situations like this
Screenshot from 2023-09-30 19-21-15

Unfortunately I wasn't able to make it work, so if someone could help me with it, it would be really appreciated :D

NuckChorris:
I don't think there's any out-of-the-box Rails validators which can do that, but a [custom validation](https://guides.rubyonrails.org/active_record_validations.html#performing-custom-validations) could do it pretty easily, like:

```ruby
validate :unit_in_media

def unit_in_media
  return if unit.blank?

  unless unit.media_type == media_type && unit.media_id == media_id
    errors.add(:unit, "must be from the tagged media")
  end
end
```

